PR #43, opened 2015-06-15
Switching by events added to admin panel

rafalkowalski:
@leto @creativepsyco @mananwason @championswimmer


open_event/forms/admin/microlocation_form.py
```diff
@@ -0,0 +1,18 @@
+"""Written by - Rafal Kowalski"""
```
creativepsyco: Incorrect way of writing author and License info. See https://code.google.com/p/soc/wiki/PythonStyleGuide#Copyright_And_License_Notices

open_event/views/admin/models_views/event.py
```diff
@@ -1,8 +1,20 @@
 """Written by - Rafal Kowalski"""
+from flask import request, url_for, redirect
+
 from flask.ext.admin.contrib.sqla import ModelView
 from ....helpers.formatter import Formatter
 from ....helpers.update_version import VersionUpdater
-
+from flask.ext.admin import expose
+from ....models.track import Track
```
creativepsyco: I think you are extending the relative import model quite far.

tests/test_event_track.py
```diff
@@ -0,0 +1,29 @@
...
+        Setup.drop_db()
+
+    def test_api_tracks(self):
+        ev = Event(name="event1", start_time=datetime.datetime(2003, 8, 4, 12, 30, 45), end_time=datetime.datetime(2003, 8, 4, 12, 30, 45))
+        tr = Track(name="name", event_id=1, description="description")
+        with app.app_context():
+            db.session.add(ev)
+            db.session.add(tr)
+            db.session.commit()
+        self.assertEqual(self.app.get('/admin/event/1').status_code, 200)
```
creativepsyco: not just 200 status should also check for update create, boundary conditions, invalid inputs. etc.

creativepsyco:
I might not have much in-depth detail about flask and python, maybe @leto can do a more thorough review. However in general i think tests can be a bit more fool proof and not just check for HTTP status codes.

creativepsyco:
Also I think http://open-event.herokuapp.com/get/api/v1/event is not right? `event` is wrapped within another `events` data structure?

tests/test_event_track.py
leto: Agreed. 200 is not the only successful code, which will lead to some false negatives

tests/test_event_track.py
```diff
@@ -0,0 +1,31 @@
...
+    def setUp(self):
+        self.app = Setup.create_app()
+
+    def tearDown(self):
+        Setup.drop_db()
+
+    def test_api_tracks(self):
+        ev = Event(name="event1", start_time=datetime.datetime(2003, 8, 4, 12, 30, 45), end_time=datetime.datetime(2003, 8, 4, 12, 30, 45))
+        tr = Track(name="name", event_id=1, description="description")
+        with app.app_context():
```
leto: This should be wrapped in a try/except block and handle those errors if they arise. For instance, if adding the event fails, you can give a different test failure message than if a track fails

tests/test_session.py
```diff
@@ -0,0 +1,29 @@
...
+        self.app = Setup.create_app()
+
+    def tearDown(self):
+        Setup.drop_db()
+
+    def test_add_session_to_db(self):
+        session = Session(title='test', description='dsad', start_time=datetime.datetime(2003, 8, 4, 12, 30, 45), end_time=datetime.datetime(2003, 8, 4, 12, 30, 45))
+
+        with app.app_context():
+            db.session.add(session)
```
leto: try/except needed here

open_event/helpers/data.py
```diff
@@ -0,0 +1,179 @@
...
+            .update(dict(data))
+        track.session = form.session.data
+        db.session.add(track)
+        db.session.commit()
+        update_version(track.event_id, False,"tracks_ver")
+
+    @staticmethod
+    def remove_track(track_id):
+        track = Track.query.get(track_id)
+        db.session.delete(track)
```
leto: try/except or create a convenience function to use

leto: Here is a function you can use to reduce repeated code: https://gist.github.com/leto/76541d4b0d34ffeb906b

open_event/helpers/data.py
```diff
@@ -0,0 +1,179 @@
...
+                              subtitle=form.subtitle.data,
+                              description=form.description.data,
+                              start_time=form.start_time.data,
+                              end_time=form.end_time.data,
+                              event_id=event_id,
+                              abstract=form.abstract.data,
+                              type=form.type.data,
+                              level=form.level.data)
+        new_session.speakers = InstrumentedList(form.speakers.data if form.speakers.data else [])
+        db.session.add(new_session)
```
leto: try/except

open_event/helpers/data.py
```diff
@@ -0,0 +1,179 @@
...
+        db.session.commit()
+        update_version(event_id, True, "session_ver")
+
+    @staticmethod
+    def update_session(form, session):
+        data = form.data
+        speakers = data["speakers"]
+        del data["speakers"]
+        db.session.query(Session)\
```
leto: try/except

open_event/helpers/data.py
```diff
@@ -0,0 +1,179 @@
...
+            .update(dict(data))
+        session.speakers = InstrumentedList(speakers if speakers else [])
+        db.session.add(session)
+        db.session.commit()
+        update_version(session.event_id, False, "session_ver")
+
+    @staticmethod
+    def remove_session(session_id):
+        session = Session.query.get(session_id)
+        db.session.delete(session)
```
leto: try/except

open_event/helpers/data.py
```diff
@@ -0,0 +1,179 @@
...
+                              event_id=event_id,
+                              twitter=form.twitter.data,
+                              facebook=form.facebook.data,
+                              github=form.github.data,
+                              linkedin=form.linkedin.data,
+                              organisation=form.organisation.data,
+                              position=form.position.data,
+                              country=form.country.data)
+        new_speaker.sessions = InstrumentedList(form.sessions.data if form.sessions.data else [])
+        db.session.add(new_speaker)
```
leto: try/except

leto: All calls to db.session.commit() below need to be wrapped in a try/except

rafalkowalski:
@leto I fixed tests ;)

PR #47, opened 2015-06-19
add support for static file based API

championswimmer:
So basically why I need this support for /event/event_id/event endpoint is because if someone uses a set of static json files (instead of a hosted python server), like I am doing in my testapi for the webapp here, https://github.com/fossasia/open-event-webapp/tree/master/testapi then we cannot save a file called /event/1 if a folder called /event/1 exists. 

This makes it easy to switch from testapi/static api to the hoster python server without changing the endpoint suffix

@rafalkowalski @leto please review and merge :)

### DIFF
--- a/open_event/views/admin/models_views/event.py
+++ b/open_event/views/admin/models_views/event.py
@@ -64,6 +64,13 @@ class EventView(ModelView):
                            event_id=event_id,
                            events=events)
 
+    @expose('/<event_id>/event')
+    def event(self, event_id):
+        events = Event.query.all()
+        return self.render('admin/base1.html',
+                           event_id=event_id,
+                           events=events)
+
     @expose('/<event_id>/track')
     def event_tracks(self, event_id):
         tracks = Track.query.filter_by(event_id=event_id)
--- a/open_event/views/views.py
+++ b/open_event/views/views.py
@@ -34,6 +34,11 @@ def get_events():
 def get_event_by_id(event_id):
     return jsonify(Event.query.get(event_id).serialize)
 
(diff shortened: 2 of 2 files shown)


championswimmer:
@rafalkowalski @creativepsyco @leto 

Can you please check this ?

creativepsyco:
I don't know but `/get/api/v1/event/<event_id>/event` isn't really a RESTful route in my opinion. Also can you not simply change the client on your side to transparently give you the correct data instead of changing the server for a special case. That would be more viable, you could hit the URL with more funky data and help with testing as well.

Plus [Line 63][1] already has event rendering, then why duplicate?

[1]: https://github.com/fossasia/open-event-orga-server/blob/master/open_event/views/admin/models_views/event.py#L63

championswimmer:
Because... We can save a file /event/1/event along with /event/1/sessions

But /event/1 cannot be a file if its already a folder.

I am talking in the sense where we can  set up a static data endpoint (only
a directory of json files, as in the given testapi link above) instead of
the full python server

creativepsyco:
Don't really know this, AFAIK a RESTful API should follow CRUD semantics. I understand that you want this because of [ApiJsonFactory][1] being abstracted away from having to decide URL routes, but the truth is your implementation already differentiates between a `testMode` and normal mode, so I see no reason to add URL routes on the API backend to facilitate this, you can declare a REST Adapter that maps the API routes, while a Test Adapter which can change the routes, map them differently for static files. 

Then you are in control, your file structure does not have to follow REST rules while right now it does, which actually it shouldn't need to.

Does that makes sense?

[1]: https://github.com/fossasia/open-event-webapp/blob/master/app/components/common/ApiJsonFactory.js

PR #69, opened 2015-06-30
add flask-migrate dependency

championswimmer:
@leto @rafalkowalski @creativepsyco

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,3 +11,4 @@ pip install flask
 pip install flask-admin
 pip install flask-sqlalchemy
 pip install Flask-WTF
+pip install flask-migrate


creativepsyco:
I think instead of this we should simply do: 

```bash
pip install -r requirements.txt 
```

Of course after python and pip packages are installed. Let's not do duplicate stuff. Also please set assignee for the PR, mentioning usernames in comments sounds laziness to me :unamused: 

image

creativepsyco:
@rafalkowalski can we instead just add the above line `pip install -r requirements.txt ` in the `install.sh` file then we don't need to add dependencies explicitly and have double code just to install the dependencies? 

Did you understand my point?

rafalkowalski:
@creativepsyco I will add it ;)

leto:
Just saw this. We must absolutely use ```pip install -r requirements.txt ```. We also must freeze our dependency versions with ```pip freeze```

PR #73, opened 2015-07-01
Login view added, Flask deploy configuration changed

rafalkowalski:
@leto


open_event/forms/admin/auth/login_form.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def validate_login(self, field):
+        user = self.get_user()
+
+        if user is None:
+            raise validators.ValidationError('Invalid user')
+
+        # we're comparing the plaintext pw with the the hash from the db
+        if not check_password_hash(user.password, self.password.data):
+        # to compare plain text passwords use
```
leto: Remove these 2 commented-out lines

open_event/views/admin/models_views/event.py
```diff
@@ -146,7 +175,10 @@ def event_session_edit(self, event_id, session_id):
 
     @expose('/<event_id>/session/<session_id>/delete', methods=('GET', 'POST'))
     def event_session_delete(self, event_id, session_id):
-        DataManager.remove_session(session_id)
+        if is_event_owner(event_id):
```
leto: This repeated code is probably best refactored to be a decorator, such that you do

```
@requires_event_owner
def foo():
....
````

and then you only write this code once. We can refactor that in a future PR

File: requirements.txt [hunk 3]
```diff
@@ -12,4 +12,5 @@ colour
 flask-cors
 pep8
 nose
-Flask-Migrate
\ No newline at end of file
+Flask-Migrate
+Flask-Login
```
leto: We should use flask-scrypt, add that as a dep. It has almost the same interface as flask-bcrypt, which it looks like you are using

leto:
Let's use scrypt:

https://github.com/cryptojuice/flask-scrypt

It is MUCH more secure than other password hash solutions.

rafalkowalski:
@leto Could you check it?

leto:
Groove :steam_locomotive: :train: !

leto:
Looks like this does not apply cleanly! I suggest merging master into this branch, resolve the conflicts, make sure all the tests pass and then ask me to review this again.

If you are unsure how to resolve the merge conflicts, let me know

Issue #116, opened 2015-08-28
"title" field missing in sessions array in API /speakers


championswimmer:
```
    "sessions": [
        {
            "id": "rp13-session-5117",
            "title": "Comic Misunderstanding – A conversation with Graham Linehan (IT Crowd)"
        },
        {
            "id": "rp13-session-5866",
            "title": "YouTube macht die Stars von heute"
        } //...
```

championswimmer:
as available on this http://springboard.championswimmer.in/api/v1/event/1/speakers/page/1

leto:
@championswimmer the above URL no longer works

rafalkowalski:
@leto Please check http://openevent.championswimmer.in/api/v1/event/1/speakers/page/1 But I don't understand this issue.

leto:
@championswimmer I am with @rafalkowalski now, can you more clearly explain what you expect to see? Thanks

championswimmer:
Within speakers, each speaker should have a sessions array, with each
session object having a "title" column

leto:
@championswimmer I see that some elements have ``` "sessions": [ ]``` while others list session ID's such as ```"sessions": [
        35, 
        255
      ], ```. Do we really want to fully populate session with the actually data? I don't have a strong opinion at this point.

leto:
@creativepsyco do you have an opinion about fully populating ```sessions``` vs only giving ID's ?

championswimmer:
just titles and ids, not the whole  session object

doing this processing on client side (loading sessions based on id,) will
mean more queries and over bandwidth.

on contrary if we send the data from the server itself it will be more
economic

leto:
@championswimmer can you give an example of exactly how you would want the data structure such that it is very easy to parse on your side?

championswimmer:
In the very first comment, it is in the code block given

creativepsyco:
Returning titles and ids is completely fine however, I am assuming that in future you will probably also require other stuff like cover image, links etc which you may want to be rendered. From a display POV it makes sense but not so much from a logical standpoint. There will be data duplication and that is not desired. Further, `sessionStorage` should allow for retrieval of sessions by ids, then you can probably fetch the session details from the session storage using the id. Hence I think for the speaker API json schema we should return array of sessions ids, and that should be enough :)

leto:
[...]

PR #122, opened 2015-10-07
Add session titles for convenience in speaker API calls #116

championswimmer:


championswimmer:
this works, and the titles are coming. assigning it off to @rafalkowalski for pulling the trigger of the merge.

PR #129, opened 2015-12-01
Script update [Don't Merge Yet]

creativepsyco:
test

Fix some of the issues with create db & version of Flask-login see https://goo.gl/WdIcqf

Database url was incorrect

Improve Readme

Add DATABASE_URL config

Bring back updated version of Flask-Login


leto:
This is looking pretty good, thanks @creativepsyco . Shouldn't create_db and run_server be done as the last few steps of ```vagrant up``` ? That is what I would expect

creativepsyco:
`runserver` can't be done from vagrant script because it is interactive in nature and shell quits after executing the script. to run you would need to ssh into it and run it yourself. To make it a more robust production setup we would need to run `uwsgi` after the setup and hook up the site configuration through `nginx` scripts, which should be trivial.

`create_db` is being done at the end of the script, its pretty similar to your setup in branch `gh_118` except I have left out `virtualenv` management, I don't think its necessary if you are doing your dev work on a vagrant machine, if you don't want to use vagrant then I suppose `virtualenv` is necessary. What do you think? This way it makes it pretty flexible if we want to use docker later on because there is no shared dependency, everything is exported through Environment variables and picked up by individual components, something like a [12-factor app](https://www.packtpub.com/books/content/building-and-running-12-factor-microservices-docker-part-2). I have only exported `DATABASE_URL` but essentially for local development, using sqlite etc, we should have some too.

leto:
Good point about ```runserver```, I am used to using uwsgi+nginx.

I agree that virtualenv on a Vagrant machine is mostly unnecessary. The big reason to use virtualenv is to not mix system dependencies and deps for our code on our prod and dev instances. I have used vagrant to provision prod/dev servers in the past, but it is an added complication.

 Do we want to use vagrant just for local development? How do we make sure the environments stay in sync? If we don't have lots of non-Python deps it should be easy, since we can rely on requirements.txt.

I really like your idea about packing a box to make things faster and simpler for new people. And maybe we should think about a docker box/image/etc too for our prod/dev servers.

creativepsyco:
For deployment purposes yes having robust vagrant scripts makes sense and
having docker and all of those goodies etc.

But mostly for local development purposes virtualenv is enough, that gets
people get started pretty fast. The only added component is Postgres which
I think we can flip to sqlite for local dev purposes using a config switch.

In my opinion the dev setup should be as dumb as possible, while a
production one should be as complicated as possible :) with tons scripts
etc.

For now, for production purposes etc, I have setup a deploy to heroku
button so you can do 1-click install.

creativepsyco:
Merging this looks fixed.

PR #136, opened 2015-12-09
Update menu.html

aditya1702:
This will remove the whitespace. But it will be the same even when there are any events.


leto:
Looks good

PR #149, opened 2015-12-18
Fixes #143

### LINKED ISSUE #143: Add map to New/Edit Event View like Microlocation View
The objective of the task is to add map to New/Edit Event View like Microlocation View. Latitude and Longitude fields should be filled by map pointer position.

xennon:
It fixes #143 issue but I don't know is it a proper and the easiest way.


leto:
@xennon I just tested this locally and I see the map. The edit page works too, nice job!

PR #158, opened 2015-12-21
Updated views.py

aditya1702:
Added methods to return speaker and session data with a given event id X

### DIFF
--- a/open_event/views/views.py
+++ b/open_event/views/views.py
@@ -196,6 +196,18 @@ def get_event_version(event_id):
     return jsonify({"version": []})
 
 
+@app.route('/api/v1/event/<event_id>/sessions/<string:session_name>', methods=['GET'])
+@cross_origin()
+def get_sessions_at_event(event_id, session_name):
+    sessions=Session.query.filter(Session.event_id == event_id,Session.name == session_name)
+    return ObjectFormatter.get_json("sessions", sessions, request)
+    
+@app.route('/api/v1/event/<event_id>/speakers/<string:speaker_name>', methods=['GET'])
+@cross_origin()
+def get_speakers_at_event(event_id, speaker_name):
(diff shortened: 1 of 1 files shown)


open_event/views/views.py
```diff
@@ -196,6 +196,18 @@ def get_event_version(event_id):
     return jsonify({"version": []})
 
 
+@app.route('/api/v1/event/<event_id>/sessions/<string:session_name>', methods=['GET'])
+@cross_origin()
+def get_sessions_at_event(event_id, session_name):
+    sessions=Session.query.filter(Session.event_id == event_id,Session.name == session_name)
```
leto: Let's us .contains() on the session name, so people can search for all sessions containing a certain word/etc.

open_event/views/views.py
```diff
@@ -196,6 +196,18 @@ def get_event_version(event_id):
...
+    return ObjectFormatter.get_json("sessions", sessions, request)
+    
+@app.route('/api/v1/event/<event_id>/speakers/<string:speaker_name>', methods=['GET'])
+@cross_origin()
+def get_speakers_at_event(event_id, speaker_name):
+    speakers=Speaker.query.filter(Speaker.event_id == event_id,Speaker.name == speaker_name)
```
leto: Same here, let's use .contains() on Speaker.name so people can search for just a last name

leto:
One of these API works and the other generates a run-time error when tested with curl:
` curl http://0.0.0.0:8001/api/v1/event/1/speakers/Duke`

```
  File "/vagrant/open_event/views/views.py", line 202, in get_sessions_at_event
    sessions=Session.query.filter(Session.event_id == event_id,Session.name.contains(session_name))
AttributeError: type object 'Session' has no attribute 'name'
```

Also please change the URL structure to `api/v1/event/1/sessions/title/X` and `api/v1/event/1/speakers/name/Y', because we will add searching by other params, too.

leto:
Thanks @aditya1702 !

PR #160, opened 2015-12-23
Added code to handle 500

aditya1702:
Added the required code in sponsor creation to handle the 500 error generated when no photo file was being submitted.


leto:
Nice catch, @aditya1702 !

PR #161, opened 2015-12-23
Added Swagger support using Flask Autodoc

aditya1702:


leto:
Nice work! Note that Swagger is a different library for doing what Autodoc does :smiley_cat:

creativepsyco:
This PR just adds autodoc and not really swagger per say.

aditya1702:
@creativepsyco Yeah i realized that afterwards. But there were only 2 repos available to document Flask applications in a way which swagger does. So i implemented one of them

PR #164, opened 2015-12-27
Added API method for ical

aditya1702:
Added API method to export event in ical format

### DIFF
--- a/open_event/views/views.py
+++ b/open_event/views/views.py
@@ -13,8 +13,11 @@ from ..models.version import Version
 from ..helpers.object_formatter import ObjectFormatter
 from flask import Blueprint
 from flask.ext.autodoc import Autodoc
+from icalendar import Calendar,Event
 
 auto=Autodoc()
+cal=Calendar()
+event=Event()
 
 app = Blueprint('', __name__)
 @app.route('/', methods=['GET'])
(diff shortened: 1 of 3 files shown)


leto:
@aditya1702 doesn't this require a change to requirements.txt? Please add that with that exact version of the library that is known to work for you. Also, what happens when `event_id` does not exist? Does it fail gracefully?

aditya1702:
@leto  updated requirements.txt. If the event id does not exist it simply returns the following:

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request.  Either the server is overloaded or there is an error in the application.</p>
Will it be fine ?

leto:
@aditya1702 a 500 is not considered a "graceful failure". Please return something like `{ error: 1, description: "event does not exist" }`, that way, the clients that use that API can tell the difference between a server problem and a non-existent event.

aditya1702:
@leto I have added the code.Will this do?

leto:
@aditya1702 that should be ok. I am having trouble installing version 3.9.2 of icalendar though, is that a development version?

```
Downloading/unpacking icalendar==3.9.2 (from -r requirements.txt (line 6))
  Could not find a version that satisfies the requirement icalendar==3.9.2 (from -r requirements.txt (line 6)) (from versions: 1.1, 1.2, 2.0.1, 2.0, 2.1, 2.2, 3.0.1b1, 3.0.1b2, 3.1, 3.2, 3.3, 3.4, 3.5, 3.6.1, 3.6.2, 3.6, 3.7, 3.8.1, 3.8.2, 3.8.3, 3.8.4, 3.8, 3.9.0, 3.9.1)
```

aditya1702:
@leto It showed version 3.9.2 here. I dont know why the error occurred. However I tried it with 3.9.1 and it works fine. I have made the changes.Do let me know if there are any issues:)

open_event/views/views.py
```diff
@@ -13,8 +13,12 @@
 from ..helpers.object_formatter import ObjectFormatter
 from flask import Blueprint
 from flask.ext.autodoc import Autodoc
+from icalendar import Calendar,Event
 
 auto=Autodoc()
+cal=Calendar()
+event=Event()
+track=Event()
```
leto: Unnecessary global variables are not good practice. Why not declare them in the functions that use them? And why not use icalendar.Event() so you don't have to change already existing code (which there is more of) that references Event?

open_event/views/views.py
```diff
@@ -237,6 +240,44 @@ def get_speakers_at_event(event_id, speaker_name):
     speakers=Speaker.query.filter(Speaker.event_id == event_id,Speaker.name.contains(speaker_name))
     return ObjectFormatter.get_json("speakers", speakers, request)
 
+@app.route('/api/v1/event/<int:event_id>/export/iCal', methods=['GET'])
+@auto.doc()
+@cross_origin()
+def generate_icalender_event(event_id):
+	"""Takes an event id and returns the event in iCal format"""		
+	event=icalendar.Event()
+	matching_event=Events.query.get(event_id)
```
leto: `NameError: global name 'Events' is not defined` . It should be `event`

leto:
Please do a test with `curl http://0.0.0.0:8001/api/v1/event/1/export/iCal` to make sure your code works and give more descriptive commit messages than "Updated file". For example, "Fixed bug X caused by Y"

leto:
I did a test of the track exporting with `curl http://0.0.0.0:8001/api/v1/track/1/export/iCal` and it seems to work. Nice job @aditya1702 !

leto:
@aditya1702 this is almost ready! It needs to be squashed into a single commit though, since it has 2 merge commits in it and is hard to follow. Follow my instructions that I gave in your other open PR. Since this PR has 10 commits, you will be using `git rebase -i HEAD~10`

PR #173, opened 2016-01-17
Added Search functionality

aditya1702:
This version of open street map adds a search functionality which allows us to search a location and get redirected to that location on map.Also a new map2.js file had to be created for the new map to be displayed.

### DIFF
--- a/open_event/templates/admin/model/create_event.html
+++ b/open_event/templates/admin/model/create_event.html
@@ -39,7 +39,7 @@
       var marker;
       var init_marker;
 function initMap() {
-  var map = new google.maps.Map($("#map"), {
+  var map = new google.maps.Map($("#map")[0], {
     zoom: 8,
     center: {lat: -34.397, lng: 150.644}
   });
@@ -50,27 +50,27 @@ function initMap() {
     map: map
   });
      google.maps.event.addListener(init_marker, 'dragend', function (event) {
-            $("#latitude").value = this.getPosition().lat();
-            $("#longitude").value = this.getPosition().lng();
+            $("#latitude")[0].value = this.getPosition().lat();
+            $("#longitude")[0].value = this.getPosition().lng();
 });
-  $("#submit").addEventListener('click', function() {
+  $("#submit")[0].addEventListener('click', function() {
     geocodeAddress(geocoder, map);
   });
 }
 
 function saveAddress(){
- var address = $("#address").value;
- $("#location_name").value=address;
+ var address = $("#address")[0].value;
+ $("#location_name")[0].value=address;
 }
 
 function geocodeAddress(geocoder, resultsMap) {
-  var address = $("#address").value;
+  var address = $("#address")[0].value;
   geocoder.geocode({'address': address}, function(results, status) {
     if (status === google.maps.GeocoderStatus.OK) {
       var latitude = results[0].geometry.location.lat();
       var longitude = results[0].geometry.location.lng();
-      $("#latitude").value=latitude;
-      $("#longitude").value=longitude;
+      $("#latitude")[0].value=latitude;
+      $("#longitude")[0].value=longitude;
       resultsMap.setCenter(results[0].geometry.location);
        init_marker.setMap(null);
        if(marker==null){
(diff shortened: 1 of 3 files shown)


leto:
This looks like a good start but needs some more work. @aditya1702 the search seems to work, but does the API key really have a space at the end of it? When I search for `Spain` I get
```
{"meta":{"version":1,"status_code":403},"results":{"error":{"type":"KeyError","message":"'XOnuJD0 ' is not a valid key."}}}
```

Also, if I search for something successfully and submit it, then go back and look, the location is not being stored correctly. It always defaults to Oakland, CA

leto:
Just tested this out again. The Geocoding stuff works well. If I put in a city name, it updates the map and the location_name form element. But it is not updating latitude or longitude, so the form cannot be submitted. This is close to working, keep it up! Let me know if you have questions.

leto:
This looks much better. The lat/lng get updated in the form and also get stored in the database. There is a bug in your initMap() function, though. It has a hardcoded lat/lng, so when you edit an event that has a lat/lng, the google map is centered in the wrong location. You should be able to use the values in `$("#latitude")` and `$("#longitude")` instead of the hardcoded values. If no lat/lng can be found, you can continue to use the hardcoded values you have currently.

leto:
Note that I used jQuery syntax above, where ` document.getElementById('foo') ` is the same as `$("#foo")`. Feel free to use that syntax and save yourself some typing

leto:
Nice work! Getting closer. Some bugs still exist:

1) When editing an existing event, the Geocode input textbox is hardcoded to Sydney instead of the existing location name.
2) When editing an existing event, there is no draggable "pin" to modify the lat/lng with the map.
3) When editing an existing event, if you manually type in the current Location Name to the Geocode input textbox, a map pin shows up, but it is not draggable.
4) When creating a new event, no pin shows up until you press "Geocode" and still, the pin is not draggable
5) It would be nice to be able to press Enter in the Geocode text input instead of only being able to click the button

open_event/templates/admin/model/create_event.html
```diff
@@ -24,18 +24,60 @@
...
     <div class="col-md-2"></div>
     <!-- bring in the google maps library -->
-        <script src="http://www.openlayers.org/api/OpenLayers.js"></script>
-        <script src="{{ url_for('static', filename='admin/js/map.js' )}}"></script>
+     <script>
+function initMap() {
+  var map = new google.maps.Map(document.getElementById('map'), {
+    zoom: 8,
+    center: {lat: -34.397, lng: 150.644}
+  });
```
leto: Please add this to the next line so that you drag the initial maker and update the lat/lng before a geocode search, which currently doesn't work:

```
google.maps.event.addListener(init_marker, 'dragend', function (event) { 
$("#latitude").val = this.getPosition().lat(); 
$("#longitude").val = this.getPosition().lng(); 
});
```

Also please start using jquery syntax like above for new code.

leto:
Great work @aditya1702 ! Just add the code I talked about above to `init_marker` so you can drag before geocoding and I will merge this.

open_event/templates/admin/model/create_event.html
```diff
@@ -24,18 +24,86 @@
...
     </div>
     <div class="col-md-2"></div>
     <!-- bring in the google maps library -->
-        <script src="http://www.openlayers.org/api/OpenLayers.js"></script>
-        <script src="{{ url_for('static', filename='admin/js/map.js' )}}"></script>
+     <script>
+      var marker;
+      var init_marker;
+function initMap() {
+  var map = new google.maps.Map($("#map"), {
```
clickmagick: Your code will probably work if you change this to `$("#map")[0]`

leto: Your code will probably work if you change this to `$("#map")[0]`, which will get the DOM object versus the jquery object.

leto:
Looks like everything broke in your latest changes. Please make sure to test your code before asking for a new review.

Read the first few answers here: http://stackoverflow.com/questions/4069982/document-getelementbyid-vs-jquery

Basically, if a function takes the output of `document.getElementById("#foo")` directly and not the `.value()` of that, you can't just replace it with `$("#foo")`, which returns a jquery object. You need to use `$("#foo")[0]` which will return the raw DOM object.

See my comment on line 42 about `#map`

aditya1702:
@leto  Sorry for the mistakes, I will test it properly and submit for review

leto:
Nice work!

Issue #170, opened 2016-01-14
Broken preview for file-uploads


yasoob:
Broken preview for file-uploads. Steps to reproduce:

- Open http://open-event.herokuapp.com/admin/event/upload
- Upload a file which is not an image
- Broken preview in the `Min` column 

Fix:
- Do not try to provide a preview for files which are not of an `image` type.

suheb:
@rafalkowalski @leto @creativepsyco @championswimmer @aditya1702 
Hey guys, what's the expected behavior here?
In my latest commit, I just show the preview depending upon whether it's an image file or not. But both this and #176 can also be fixed if we only allow image files to be uploaded.

leto:
I don't see a use case for uploading files that are not images. Can anybody else? Maybe @mariobehling has some thoughts

creativepsyco:
So we can either restrict uploading or provide the preview image ourselves in case it is broken (similar to FB/GitHub providing a default profile in case you are missing one).

aditya1702:
@leto I agree with @creativepsyco .We can use a default image that gets displayed in absence of an image.

yasoob:
We can only allow uploading `.docx`, `pdf` and image files. I don't think there is any need for any other files apart from these.

suheb:
Looking at the current design, it seems file uploading was intended for images only. For instance, the url ``` /pic/file_name ``` and the preview column in the table.
Let me know how to move forward with this.

leto:
@suheb Let's do pictures only for now, because that was the intention. We should accept .jpg, .jpeg, .gif and .png

suheb:
@leto Done, see my pull request.

suheb:
@yasoob You can close this issue. Fix #177

PR #177, opened 2016-01-19
Fixing: Broken preview for file-uploads #170

suheb:
I have added a field "is_image" in File model and check its value when generating the template to show the preview or not. 
Signed-off-by: Suhaib Khan <suhebjerk@gmail.com>


leto:
Looks good, thanks @suheb!

PR #180, opened 2016-01-24
Fixed #178

### LINKED ISSUE #178: Creating an event with logo does not work.
Steps to reproduce:
1. Click on "Create New Event" on home page.
2. Fill in the required fields.
3. Choose any file for the logo field. This step is important.
4. Click on "Submit".
This gives an internal server error.
Tested on both local machine and <a href="http://open-event.herokuapp.com/admin/">Demo site</a>.
It should be noted that this does not happen when logo field is left empty.

aditya1702:


suheb:
@aditya1702 Ohh I didn't saw you pull request before and created my own. See here: #181
I have also made some improvements in the UI and update_event().

leto:
This no longer applies cleanly (merge conflicts) and #181 does this plus a bit more, so I think we should go with #181 once the merge conflicts get fixed there.

Issue #175, opened 2016-01-17
Allows uploading multiple files with the same name


yasoob:
- Navigate to [http://localhost:8001/admin/event/upload](File Upload) Page.
- Upload a file.
- Edit the contents of the original copy of that same file.
- Upload that same file again with the modified content.
- It uploads without any errors.

The second file over-rights the first file without giving us any clue. The default behavior of the application should not allow us to upload two different files with the same name. This way you can loose your data.

PR #182, opened 2016-01-24
Fixes: Allows uploading multiple files with the same name #175

suheb:
Signed-off-by: Suhaib Khan <suhebjerk@gmail.com>

### DIFF
--- a/open_event/helpers/data.py
+++ b/open_event/helpers/data.py
@@ -2,6 +2,7 @@
 import logging
 
 from sqlalchemy.orm.collections import InstrumentedList
+from sqlalchemy.sql.expression import exists
 from flask import flash, request
 from flask.ext import login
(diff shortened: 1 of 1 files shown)


leto:
Looks like we have merge conflicts, could you update this PR please?

suheb:
@leto Done.

suheb:
@leto I think we're good to go here.

leto:
Thanks for being persistent :+1:

leto:
Am I correct in seeing that http://127.0.0.1:8001/admin/event/upload is only available by typing in the URL? Does any page actually link to it anymore?

rafalkowalski:
@creativepsyco @leto I think that uploading files is not necessary. We had a huge problem in Heroku with storing files, so we decided that we don't need this. I only save a link to database

PR #186, opened 2016-02-17
Added reviews for sessions

SaptakS:
I have added the database model, views and helper python functions so that it may function in the following way.

```
POST - "/sessions/:id/reviews" 
body - { 
"email" : "me@example.org",
"rating": 4.0, // can be from 1.0 to 5.0
"comment": "Lorem Ipsum"
}

RESPONSE:
200 OK

GET - "/sessions/:id/reviews"
Response - "
{
"reviews": [{
          "email" : "me@example.org",
          "rating": 4.0, 
         "comment": "Lorem Ipsum"
         },
        // More reviews
        ],
"average": 3.5
"total": 100
// More paging stuff
}
"
```

### DIFF
--- a/open_event/views/views.py
+++ b/open_event/views/views.py
@@ -1,5 +1,6 @@
 """Copyright 2015 Rafal Kowalski"""
 import os
+import re
 from flask import jsonify, json, url_for, redirect, request, send_from_directory
 from flask.ext.cors import cross_origin
 
@@ -296,20 +297,29 @@ def post_reviews(session_id):
     """Post reviews for sessions"""
     data = json.loads(request.data)
     if "comment" not in data:
-            data["comment"] = ""
-
-    reviews = DataGetter.get_reviews_by_email(data["email"])
-    print reviews.count()
-
+        data["comment"] = ""
+    reviews = DataGetter.get_reviews_by_email_and_session_id(data["email"])
     if reviews.count() > 0:
         DataManager.update_review(data, reviews[0], session_id)
-        return "200 OK\n"
-    elif "email" in data and "rating" in data:  
+        return jsonify(
+                {'status': '200',
+                 'detail': 'Review Updated'})
+
+    elif "email" in data and "rating" in data: 
+        if not re.match(r"[^@]+@[^@]+\.[^@]+", data["email"]):
+            return jsonify(
+                {'status': '400',
+                 'detail': 'Not valid data'})
+
         DataManager.create_review(data, session_id)        
-        return "200 OK\n"
-    else:
-        print "Error: Email and Rating are required\n"
-        return "Error: Email and Rating are required\n"
+        return jsonify(
+                {'status': '200',
+                 'detail': 'Review Added'})    
+    
+    return jsonify(
+                {'status': '500',
+                "title": "The backend responded with an error",
+                "detail": "All required paramters not obtained"})
     
 
 
@@ -322,3 +332,4 @@ def send_pic(filename):
 @app.route('/documentation')
 def documentation():
 	return auto.html()
+
--- a/open_event/views/views.py
+++ b/open_event/views/views.py
@@ -1,6 +1,6 @@
 """Copyright 2015 Rafal Kowalski"""
 import os
-from flask import jsonify, url_for, redirect, request, send_from_directory
+from flask import jsonify, json, url_for, redirect, request, send_from_directory
 from flask.ext.cors import cross_origin
 
 from ..models.track import Track
@@ -9,8 +9,11 @@ from ..models.sponsor import Sponsor
 from ..models.microlocation import Microlocation
 from ..models.event import Event
 from ..models.session import Session, Level, Format, Language
+from ..models.reviews import Review
 from ..models.version import Version
 from ..helpers.object_formatter import ObjectFormatter
+from ..helpers.data import DataManager, save_to_db, delete_from_db
+from ..helpers.data_getter import DataGetter
 from flask import Blueprint
 from flask.ext.autodoc import Autodoc
 from icalendar import Calendar
@@ -277,6 +280,38 @@ def generate_icalender_track(track_id):
 	track.add('image url',matching_track.track_image_url)
 	cal.add_component(track)
 	return cal.to_ical()
+
+@app.route('/api/v1/session/<int:session_id>/reviews', methods=['GET'])
+@auto.doc()
+@cross_origin()
+def get_reviews(session_id):
+    """Returns all session's reviews"""
+    reviews = Review.query.filter_by(session_id=session_id)
+    return ObjectFormatter.get_json("reviews", reviews, request)
+
+@app.route('/api/v1/session/<int:session_id>/reviews', methods=['POST'])
+@auto.doc()
+@cross_origin()
+def post_reviews(session_id):
+    """Post reviews for sessions"""
+    data = json.loads(request.data)
+    if "comment" not in data:
+            data["comment"] = ""
+
+    reviews = DataGetter.get_reviews_by_email(data["email"])
+    print reviews.count()
+
+    if reviews.count() > 0:
+        DataManager.update_review(data, reviews[0], session_id)
+        return "200 OK\n"
+    elif "email" in data and "rating" in data:  
+        DataManager.create_review(data, session_id)        
+        return "200 OK\n"
+    else:
+        print "Error: Email and Rating are required\n"
+        return "Error: Email and Rating are required\n"
+    
+
 
 @app.route('/pic/<path:filename>')
 @auto.doc()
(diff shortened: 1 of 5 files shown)


open_event/views/views.py
```diff
@@ -278,6 +281,38 @@ def generate_icalender_track(track_id):
...
+
+    if reviews.count() > 0:
+        DataManager.update_review(data, reviews[0], session_id)
+        return "200 OK\n"
+    elif "email" in data and "rating" in data:  
+        DataManager.create_review(data, session_id)        
+        return "200 OK\n"
+    else:
+        print "Error: Email and Rating are required\n"
+        return "Error: Email and Rating are required\n"
```
creativepsyco: you should be returning a valid json and not just ok, plus it doesn't have the right status. See [JSONAPI](http://jsonapi.org/format/) to figure out the format. usually it should be of the form:

```json
{
  "errors": [
    {
      "status": "403",
      "source": { "pointer": "/data/attributes/secret-powers" },
      "detail": "Editing secret powers is not authorized on Sundays."
    },
    {
      "status": "422",
      "source": { "pointer": "/data/attributes/volume" },
      "detail": "Volume does not, in fact, go to 11."
    },
    {
      "status": "500",
      "source": { "pointer": "/data/attributes/reputation" },
      "title": "The backend responded with an error",
      "detail": "Reputation service not responding after three requests."
    }
  ]
}
```

creativepsyco:
A few pointers:
* The api should be hosted at: `api/v1/event/:event_id/sessions/:session_id/reviews`, it is no reason to move away from this format
* Pagination stuff is missing, I would recommend, giving 50 reviews and then page the rest. Look for pagination format here: http://jsonapi.org/examples/#pagination
* Data returned is not valid json and not contains the HTTP status code information.
* Migration file is missing. Please use `python manage.py db migrate` to generate a migration file for model changes. See [Readme](https://github.com/fossasia/open-event-orga-server/blob/master/README.md)

creativepsyco:
Also don't use `print` use the logger.info/warn/debug/verbose options.

open_event/views/views.py
```diff
@@ -278,6 +281,38 @@ def generate_icalender_track(track_id):
...
+    """Post reviews for sessions"""
+    data = json.loads(request.data)
+    if "comment" not in data:
+            data["comment"] = ""
+
+    reviews = DataGetter.get_reviews_by_email(data["email"])
+    print reviews.count()
+
+    if reviews.count() > 0:
+        DataManager.update_review(data, reviews[0], session_id)
```
creativepsyco: This seems wrong, you are supposed to change the review for a particular session, but you are seemingly updating the first review in your query after filtering them by email.

A session can have many reviews.
A review must belong to user and to a session.
A user can have many reviews for many sessions.

open_event/views/views.py
```diff
@@ -278,6 +281,38 @@ def generate_icalender_track(track_id):
...
+    return ObjectFormatter.get_json("reviews", reviews, request)
+
+@app.route('/api/v1/session/<int:session_id>/reviews', methods=['POST'])
+@auto.doc()
+@cross_origin()
+def post_reviews(session_id):
+    """Post reviews for sessions"""
+    data = json.loads(request.data)
+    if "comment" not in data:
+            data["comment"] = ""
```
creativepsyco: too much indentation here.

open_event/views/views.py
```diff
@@ -278,6 +281,38 @@ def generate_icalender_track(track_id):
...
+
+    reviews = DataGetter.get_reviews_by_email(data["email"])
+    print reviews.count()
+
+    if reviews.count() > 0:
+        DataManager.update_review(data, reviews[0], session_id)
+        return "200 OK\n"
+    elif "email" in data and "rating" in data:
```
creativepsyco: use proper validation procedures. including validity of emails.

open_event/views/views.py
```diff
@@ -278,6 +281,38 @@ def generate_icalender_track(track_id):
...
+@app.route('/api/v1/session/<int:session_id>/reviews', methods=['POST'])
+@auto.doc()
+@cross_origin()
+def post_reviews(session_id):
+    """Post reviews for sessions"""
+    data = json.loads(request.data)
+    if "comment" not in data:
+            data["comment"] = ""
+
+    reviews = DataGetter.get_reviews_by_email(data["email"])
```
creativepsyco: `get_reviews_by_email_and_session_id`

open_event/views/views.py
```diff
@@ -278,6 +281,38 @@ def generate_icalender_track(track_id):
 	cal.add_component(track)
 	return cal.to_ical()
 
+@app.route('/api/v1/session/<int:session_id>/reviews', methods=['GET'])
```
creativepsyco: `PUT` method for updating existing review?

creativepsyco:
I am also thinking there needs to be some sort of a secret shared key between the server & client to prevent abuse of the API. Let me think about that.

leto:
@creativepsyco yes, very much agreed. API write operations need some kind of API key.  See #187 for my brain dump.

I would say that this PR can get merged before we have API keys figured out but we definitely need API keys before running anything in production or we are just asking for trouble. Once we have API keys implemented on the backend it should be a simple matter of adding a decorator to any API methods that do write operations.

open_event/views/views.py
SaptakS: Yeah.. I got that wrong.. I have to check for both the user and the session. Will make the change asap.

open_event/views/views.py
SaptakS: Won't that already be done while authentication from the client-side?

open_event/views/views.py
SaptakS: The POST method will handle updating existing review.. Should I use PUT for update?

SaptakS:
@creativepsyco The api is hosted in `api/v1/sessions/:session_id/reviews` because every session has a unique session_id. So is it needed to be like this: `api/v1/event/:event_id/sessions/:session_id/reviews`?

open_event/views/views.py
SaptakS: I didn't understand the source part of the json.. Can you please elaborate? I am working on the rest.

SaptakS: For errors, should I use the `@app.errorhandler` of flask? In that case, it will send this error response in case of error in all the others APIs mentioned before. Is that ok?

SaptakS:
Also, which data returned is not a valid JSON? The status or the GET data? For the GET data, I have used the format which has been used in the other APIs.

open_event/views/views.py
creativepsyco: see validators.py, atleast have some pre-conditions, can't depend on the client to have all the checks.

creativepsyco:
return `return "200 OK\n"`

and `"Error: Email and Rating are required\n"` is not valid json, you are returning this from the views.py file, it has to be a json string.

creativepsyco:
Plus has to be the correct status code, even when you are raising errors. 400 for the case of invalid data.

open_event/views/views.py
SaptakS: Ok.

open_event/views/views.py
```diff
@@ -278,6 +281,38 @@ def generate_icalender_track(track_id):
...
+
+    reviews = DataGetter.get_reviews_by_email(data["email"])
+    print reviews.count()
+
+    if reviews.count() > 0:
+        DataManager.update_review(data, reviews[0], session_id)
+        return "200 OK\n"
+    elif "email" in data and "rating" in data:  
+        DataManager.create_review(data, session_id)        
+        return "200 OK\n"
```
SaptakS: Just to clarify, do you mean my response should be something like:
```
return jsonify({
    "status" : 200,
    "detail" : "Review Added Successfully"
})
```

creativepsyco: yes

SaptakS:
I have jsonified the status returns, added email validation and added retrival of rating from database using email and session-id so that a rating with a particular email and session-id is unique.

SaptakS:
Also, can you please clarify if I need to add a `PUT` request function as well because I have already implemented the updating functionality using `POST`. If needed please let me know.

SaptakS:
@leto @rafalkowalski can you please review this PR?

leto:
@SaptakS I need a migration file for the schema, the current PR by itself does not work:

`curl -v http://127.0.0.1:8001/api/v1/session/1/reviews` generates a 500 error:
`ProgrammingError: (psycopg2.ProgrammingError) relation "session_reviews" does not exist`

SaptakS:
@leto I am not sure what you want me to do. The migration file is already there. I have just changed the views so migration file will still be there I guess. Please let me know if I got it wrong.

leto:
What is the name of the migration file? I am talking about running the commands in the "Model updates" section of the README.md, in particular: ` python manage.py db migrate`

which should create a file in `migrations/versions/`. Make sure to `git add` it and make it part of the PR

SaptakS:
Ok. I will check it.

SaptakS:
Closing this PR due to some conflicting code.

PR #188, opened 2016-03-03
count removed from json

mananwason:


leto:
Any reasoning for this?

mananwason:
@leto In the format we were not using count in the api and I have configured the classes such that Gson is able to parse the json data easily but due to introduction of the count object, I would have to make a lot of changes in the client side code and we aren't even using the count object anywhere, so I think we should remove it.

mananwason:
@rafalkowalski :+1:

PR #192, opened 2016-03-08
Validate form only on POST requests

shivamMg:
Remove unnecessary validation of the Event creation form on GET requests. This makes the "required field" messages to appear even though the user has not submitted the form.
Validate form only if it is a POST request.


shivamMg:
@aditya1702 Can you review this.

shivamMg:
Submitting new request with changes on a new branch.

PR #197, opened 2016-03-10
Validate form only on POST request

shivamMg:
Fixes issue at https://github.com/fossasia/open-event-orga-server/issues/196


leto:
Thanks for fixing this! :+1:

Issue #130, opened 2015-12-01
Compress CSS, JS and other static files.


creativepsyco:
Right now, the heroku deployment is slower and a lot of css files are not correctly minified. This should probably change so that access is faster. See image below for clues on where to optimise.

![image](https://cloud.githubusercontent.com/assets/547235/11492056/03f77092-9823-11e5-8848-148055db1865.png)

shivamMg:
@creativepsyco I've taken this on https://github.com/fossasia/open-event-orga-server/issues/191. What do you think?

PR #198, opened 2016-03-18
Minifying javascript and css assets

niranjan94:
Resolves issue #130 by minifying javascript and css resources. 

- Uses [Flask-Assets](http://flask-assets.readthedocs.org/en/latest/), [jsmin](https://pypi.python.org/pypi/jsmin) and [cssmin](https://pypi.python.org/pypi/cssmin) for minification, caching and serving. 
- The resources have been rearranged and grouped into proper subdirectories inside css/js folder.
- Missing semicolons were added in a few javascript files
- leaflet and leaflet.draw was combined into a single file. 

The minification does not happen when in `StagingConfig`, `DevelopmentConfig`, `LocalPSQLConfig`, `LocalSQLITEConfig`.


niranjan94:
Fresh pull request at #217 . Some commits have been squashed.

PR #201, opened 2016-03-18
Added Email validation to forms

aviaryan:
Email entered when creating a new event or adding a new speaker is validated.


leto:
Thanks!

Issue #202, opened 2016-03-18
Logging-in with email


aviaryan:
Like most other web-apps and services, `Open Event Organizer` should also allow logging in with email in addition to username. 

UPDATE : Opened PR #203

aviaryan:
I can't help but wonder why is `email` field not unique here . https://github.com/fossasia/open-event-orga-server/blob/master/open_event/models/user.py#L13 . I believe it has something to do with `roles` but I am not quite sure. Any hints would be lovely. 

```
 email = db.Column(db.String(120))
```

aviaryan:
PR #203 Merged. Closing

PR #203, opened 2016-03-18
Login with email too

aviaryan:
See #202 

One concern maybe that how login is done in case of users with same emails (assuming it is a feature). But there shouldn't be any problems as the query first matches the login (username) and then the email.

### DIFF
--- a/open_event/forms/admin/auth/login_form.py
+++ b/open_event/forms/admin/auth/login_form.py
@@ -10,16 +10,32 @@ class LoginForm(form.Form):
     login = StringField(validators=[validators.required()])
     password = PasswordField(validators=[validators.required()])
 
+    def __init__(self, form):
+        """Init class"""
+        super(LoginForm, self).__init__(form)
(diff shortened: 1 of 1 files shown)


aviaryan:
The second commit takes care of unlikely scenario where a user uses email of another user as his/her username (aka login). Also users with same emails can now login with their email itself (the password plays the distinguisher).

leto:
If I register two new accounts, both with the same email and password, then login with an email and password, the first account is logged into. I think it is a bug that we allow somebody to register with an email that is already registered. If we fix that, the above problem goes away. Do you want to submit a PR for that?

Also, the UI might want to say "Login or Email" so that users know they can use either.

aviaryan:
> I think it is a bug that we allow somebody to register with an email that is already registered.

I agree. PR #227

leto:
Now that email is unique, this is solid

leto:
Thanks!

PR #205, opened 2016-03-19
Use correct repo link in Dockerfile

shivamMg:
Use the original fossasia repo link for cloning repo while building image from Dockerfile.


leto:
Thanks!

PR #207, opened 2016-03-19
fixed #206, duplicate data in iCal export

aviaryan:
See #206 for description

PR #208, opened 2016-03-19
Renamed some fields in exported iCalendar according to specification, fixes #204

### LINKED ISSUE #204: Standardize ical export
In reference to #106, the ical export method should try to use standard fields (see iCalendar specification below) for compatibility with existing calendar services. 
Example - Fields like 'start time' and 'end time' are non-standard and 'dtstart' and 'dtend' should be used in their place.

Some helpful links 
* [Google Calendar ICS export](https://support.google.com/calendar/answer/37118?hl=en)
* [Wikipedia iCalendar](https://en.wikipedia.org/wiki/ICalendar)
* [iCalendar specification](https://tools.ietf.org/html/rfc5545#section-3.8.1.7)

aviaryan:
See #204 for description

(This PR includes #207 )

I still haven't found a *specification name* for fields like email, logo, colour etc. But as these fields are not that important from the calendar point of view, I hope this PR is fine.

### DIFF
--- a/open_event/views/views.py
+++ b/open_event/views/views.py
@@ -249,16 +249,15 @@ def generate_icalender_event(event_id):
 	matching_event=Event.query.get(event_id)
 	if matching_event == None:
 		return "Sorry,the event does not exist"
-	event.add('name',matching_event.name)
-	event.add('latitude',matching_event.latitude)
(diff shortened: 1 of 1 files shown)


leto:
Thanks!

PR #215, opened 2016-03-21
Fix session proposal view

shivamMg:
In the views, URLs `/<event_id>/session/new_proposal` and `/<event_id>/session/new` have functions defined with the same name.

In the templates, since the function names are same there is a conflict and `/<event_id>/session/new` can never be retrieved.

- Rename the function used for creating new session proposals.
- Change template to reflect the same.


leto:
Thanks!

PR #216, opened 2016-03-22
Validate forms only on POST requests

shivamMg:
Worked on the same issue https://github.com/fossasia/open-event-orga-server/issues/196 for other views.


leto:
Thanks!

PR #217, opened 2016-03-22
Minifying javascript and css assets

### LINKED ISSUE #130: Compress CSS, JS and other static files.

niranjan94:
#### What does this PR do?
- Uses [Flask-Assets](http://flask-assets.readthedocs.org/en/latest/), [jsmin](https://pypi.python.org/pypi/jsmin) and [cssmin](https://pypi.python.org/pypi/cssmin) for minification, caching and serving. 
- The resources are rearranged and grouped into proper subdirectories inside css/js folder.
- leaflet and leaflet.draw combined into a single file. 

#### Resolves issue
Resolves #130 opened by @creativepsyco on 1<sup>st</sup> December 2015

#### Behaviour in a development environment
The minification does not happen when using `StagingConfig`, `DevelopmentConfig`, `LocalPSQLConfig`, `LocalSQLITEConfig`.

#### How was this tested
- Manually in a browser


juslee:
Update PR with latest master. @niranjan94

niranjan94:
@juslee on it

PR #218, opened 2016-03-24
Show event switcher dropdown menu only when logged in.

niranjan94:
#### The problem
The event switcher menu is shown even when the user is Unauthenticated. And clicking on any event/link in the dropdown menu takes the user back to the login page. 

So, it's better if the menu is not shown when the user is not logged it. (all the functionality the menu provides, will anyway work only if the user is logged in)

#### What does this PR do?
- Shows the event switcher menu only when the user is logged in

#### How was this tested
- Manually in a browser


championswimmer:
This looks a good issue to resolve. Thanks for putting this in.

leto:
Thanks, this makes much more sense

PR #219, opened 2016-03-24
Redirect user to the intended URL after login

niranjan94:
#### What does this PR do?
When the user accesses a page without logging in and is redirected to the login page, he will be taken back to the intended URL (got from the `next` query string parameter) after he logs in.

#### Resolves issue
Issue #226 opened by @niranjan94 on 26<sup>th</sup> March 2016

#### How was this tested
- Manually in a browser


leto:
Thanks, great to have this

PR #222, opened 2016-03-25
Add scroll feature to Event List table

aditya1702:
This prevents the table from going out of bounds of the div class


aviaryan:
A suggestion, it will be better if a new div is added over `table` and it is given the class `event-list_table`. That way just the table would look scrollable rather than the whole card.

PR #224, opened 2016-03-26
Add __str__ to all models.

### LINKED ISSUE #223: Non-human readable display of labels for QuerySelectField and QuerySelectMultipleField in all forms
WTForm uses the model object's `__str__` or `__unicode__` to get the string to use as the label for `QuerySelectField` and `QuerySelectMultipleField` [as explained here](http://wtforms.readthedocs.org/en/2.1/ext.html#wtforms.ext.sqlalchemy.fields.QuerySelectField). 

But, most of the models don't have a `__str__` or `__unicode__` defined. Hence, WTForm uses the value returned by `__repr__`. Which is of the form `<Object value>`. 

**Example:**
capture

This can be fixed by implementing a `__str__` in all the models and returning a human readable string like the session name/microlocation name/etc ... (returning `self.name` should suffice in most cases)

niranjan94:
#### What does this PR do?
- Implements `__str__` in all the models and returns the human readable name of the entry stored in that model. (`self.name` in most cases).

#### Resolves issue
Issue #223 opened by @niranjan94 on 26<sup>th</sup> March 2016

#### Screenshots
**Before:**
capture
**After:**
capture2

#### How was this tested
- Manually in a browser


leto:
Thanks!

PR #225, opened 2016-03-26
Added 404 error page

niranjan94:
#### What does this PR do?
- Adds a minimal HTML 404 error page when a request is made to a non-existent URL
- Displays a json error message `{"error": "endpoint_not_found"}` is shown if the request has the `Accept` header as `application/json`

#### Screenshot
screenshot 45

#### How was this tested
- Manually in a browser (for the HTML error page)
- Using curl `curl -Haccept:application/json http://localhost:5000/some-page` (for the JSON error message)

### DIFF
--- a/open_event/__init__.py
+++ b/open_event/__init__.py
@@ -2,12 +2,16 @@
 import logging
 import os.path
 import sys
+import json
 
 from flask import Flask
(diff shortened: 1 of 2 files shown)


leto:
Thanks!

PR #227, opened 2016-03-27
Made email unique for a user

aviaryan:
Made email field unique in user model as discussed at https://github.com/fossasia/open-event-orga-server/pull/203#issuecomment-201948310


leto:
Thanks!

PR #229, opened 2016-03-30
Use checkboxes to display MultipleFields

shivamMg:
Use Check-boxes to display `Sessions` or `Speakers` fields in Speaker, Session and Track forms.

[Screenshot (new field)](http://i.imgur.com/z15mu4E.png)

The current forms use drop-down select tag that are not fit for the purpose. Selecting more than one option requires the user to hold down the shift key. [Screenshot](https://i.imgur.com/Mo1mLWz.png)

- Remove `from flask_wtf.file import FileField` from `speaker_form.py`. It was imported but not used.
- Remove `allow_blank=True` option specified in `QuerySelectMultipleField`s. It does not accept any such option [docs](https://wtforms.readthedocs.org/en/2.0.2/ext.html?highlight=queryselectmultiplefield#wtforms.ext.sqlalchemy.fields.QuerySelectMultipleField).

### DIFF
--- a/open_event/forms/admin/session_form.py
+++ b/open_event/forms/admin/session_form.py
@@ -1,6 +1,7 @@
 """Copyright 2015 Rafal Kowalski"""
 from flask_wtf import Form
 from wtforms import StringField, TextAreaField
+from wtforms.widgets import ListWidget, CheckboxInput
 from wtforms.ext.sqlalchemy.fields import QuerySelectMultipleField, QuerySelectField
 from wtforms.validators import DataRequired
 from flask_admin.form.fields import DateTimeField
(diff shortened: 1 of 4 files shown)


shivamMg:
@championswimmer Please take a look :smile:

leto:
This looks good. My only concern is: What happens when there are hundreds of speakers? We may want something like Github's interface for adding a person to a repo: A search field and a list of the people added below. This interface is nice for small numbers of speakers, so I am going to merge it.

shivamMg:
@leto Taken care of it in https://github.com/fossasia/open-event-orga-server/pull/242.

shivamMg:
@leto Something I would like to ask. I noticed "Tracks" has a Sessions field in forms but does not have "sessions" key in the API. So there is no way to retrieve sessions for a track. Should I file an issue and submit a patch?

PR #230, opened 2016-03-30
Specify proper labels in Login and Register forms

shivamMg:
The current Register form uses 'Login' as the field label for Username. The Login form uses 'login' as field label for Username/Email.


leto:
Thanks!

PR #234, opened 2016-04-01
Highlight current event in navbar dropdown

shivamMg:
There is no element that indicates the current event being browsed by the user.

- Highlight the current event being browsed in the "Switch event" drop-down
- Place a divider between events and "Add new event"


leto:
Thanks!

PR #244, opened 2016-04-15
Fixed travis build

aviaryan:
This PR fixes Travis build error.

### Comments

* cb6bcbb - This project was not compatible with Python 3+ so I removed all other python versions. Furthermore assuming that we all are using Python 2.7 and the same will be used on Heroku, I have only kept v2.7. Please correct me if I am wrong. 

See https://travis-ci.org/aviaryan/open-event-orga-server/builds/123264147 for the travis build log of my fork.

PR #245, opened 2016-04-15
Rewrite failing tests

shivamMg:
- Specify `mock2.0.0` to be a requirement in `requirements.txt`. It is being used in validation tests (`test_validations.py`).
- Typecast `hash` to string during user creation in `helpers/data.py`. `User.reset_password` inside models has been defined a string.
- Track count (`len(Track.query.all())`) inside `text_event_track.py` would be `1`.
- Use `assertRaises()` correctly.
- Remove `test_event_start_time_smaller_than_end_time` test from `test_validations.py`. No validation error is raised for start time smaller than end time.

### DIFF
--- a/open_event/helpers/data.py
+++ b/open_event/helpers/data.py
@@ -382,7 +382,7 @@ class DataManager(object):
         password = form.password.data
         user.password = generate_password_hash(password, salt)
         hash = random.getrandbits(128)
-        user.reset_password = hash
+        user.reset_password = str(hash)
 
         user.salt = salt
         user.role = 'speaker'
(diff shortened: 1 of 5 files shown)


aviaryan:
@shivamMg Are you able to run `nosetests` on the terminal and pass all the 21 tests. I am asking this because I had lots of problems running the tests, at one place I had to rename tests/setup.py to something else to fix the issue (it was a nose issue, they even had it open on their github repo https://github.com/nose-devs/nose/issues/54). See #244 for the log.

juslee:
Test fails. Rewrite and make it pass.

shivamMg:
Changes added in https://github.com/fossasia/open-event-orga-server/pull/247

PR #249, opened 2016-04-15
Add more tests for `CustomDateSessionValidate`

shivamMg:
Issue: https://github.com/fossasia/open-event-orga-server/issues/300

### DIFF
--- a/tests/test_validations.py
+++ b/tests/test_validations.py
@@ -30,6 +30,19 @@ class TestValidation(OpenEventTestCase):
             self.event_form = EventForm()
             self.session_form = SessionForm()
 
+    def _session_form_req_url(self, start_time, end_time):
+        """
+        Helper function to initialize Session form with start and end time,
+        and validate it.
+        """
+        validator = CustomDateSessionValidate()
+
+        self.session_form['start_time'].data = start_time
+        self.session_form['end_time'].data = end_time
+        request.url = 'http://0.0.0.0:5000/admin/event/1'
+
+        validator(form=self.session_form, field=None)
+
     def test_event_start_and_end_time(self):
         """Event end time should not be smaller than its start time"""
         with app.test_request_context():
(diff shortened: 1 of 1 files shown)


aviaryan:
> No assertions are being made in the test.

I think in that case the test will fail if an exception is thrown. So the *test* will pretty much work as its purpose is served. 

btw, one reason to remove that test can be that `test_event_end_time_smaller_than_start_time` is very similar to it. But still there is no harm in keeping it.

shivamMg:
> So the test will pretty much work as its purpose is served.

> one reason to remove that test can be that `test_event_end_time_smaller_than_start_time` is very similar to it.

It's not similar to it. `CustomDateEventValidate(object)` includes a [condition](https://github.com/fossasia/open-event-orga-server/blob/master/open_event/helpers/validators.py#L32) that checks if the start time of an event is not smaller than its end time (Event with start time greater than end time is invalid). If it is, it raises a ValidationError. That's why we have a test for it.

If the start time for an event is smaller than its end time, there is nothing to check. So no, `test_event_start_time_smaller_than_end_time` does not serve its purpose.

I have added a few more tests for `CustomDateSessionValidate()`, to check if Session date range is inside Event date range. It seems you have remove `MagicMock()` from your previous commit, any reason.

aviaryan:
> It seems you have remove MagicMock() from your previous commit, any reason.

The test would fail because of that. Also I had no idea why it was there to begin with, other tests like "test_session_start_time_and_end_time_not_inside_event_date_range" were very similar but weren't using it. Simple datetime() function should have sufficed.

juslee:
@shivamMg do an update from master and resolve conflicts.

shivamMg:
PR check at codacy is failing because of duplicating blocks like the ones where we are assigning `start_time` and `end_time` in session_form [ex](https://github.com/fossasia/open-event-orga-server/pull/249/files#diff-48c328f70edd40a80c8994ffec35347bR72). But aren't unit tests supposed to be independent from each other?

juslee:
http://learnpythonthehardway.org/book/ex47.html

> Testing Guidelines
> Follow this general loose set of guidelines when making your tests:
> 
> 1. Test files go in tests/ and are named BLAH_tests.py otherwise nosetests won't run them. This also keeps your tests from clashing with your other code.
> 2. Write one test file for each module you make.
> 3. Keep your test cases (functions) short, but do not worry if they are a bit messy. Test cases are usually kind of messy.
> 4. Even though test cases are messy, try to keep them clean and remove any repetitive code you can. Create helper functions that get rid of duplicate code. You will thank me later when you make a change and then have to change your tests. Duplicated code will make changing your tests more difficult.
> 5. Finally, do not get too attached to your tests. Sometimes, the best way to redesign something is to just delete it and start over

shivamMg:
@juslee Done :+1: I've created a helper function that reduces redundant code. But I did not change test file names from `test_*.py` to `*_tests.py` as described in the first point. I tried it on my computer and nosetests ran fine for both the patterns, but when I did a `python -m unittest discover` it could not run tests with the latter pattern (`*_tests.py`) as unittest requires test files to be of the former pattern. Thought I might ask, if it's still required I will rename all files in one commit.

juslee:
#300 close

PR #250, opened 2016-04-15
Use `Select2Widget` to display Select fields

shivamMg:
Use `Select2Widget` with `multiple=True` to display select fields in Session, Speaker and Track forms.

select2.js is shipped with `flask-admin` so no external libraries were needed.

https://github.com/fossasia/open-event-orga-server/pull/229#issuecomment-205033412


shivamMg:
@leto If you can take a look :smile:. Here's an example [screenshot](https://i.imgur.com/9Jfy6Vf.png)

leto:
Looks good to me

PR #258, opened 2016-04-24
Fix 404 errors for Flask-Admin static assets

shivamMg:
After Flask-Admin upgrade from 1.1.0 to 1.4.0: `vendor/jquery-2.1.1.min.js` is now at `vendor/jquery.min.js`, `vendor/moment-2.8.4.min.js` at `vendor/moment.min.js` and `admin/js/form-1.0.0.js` at `admin/js/form.js`.


aditya1702:
This PR is good! I tested it.It is working :)

PR #271, opened 2016-05-03
Added two test files

aditya1702:


aditya1702:
@leto @rafalkowalski Can this be merged? The three error details are not being loaded

leto:
I can't seem to run the tests:

https://gist.github.com/leto/8a60e28b8ae7a38c6c44cc492df81728

that is after doing a "sudo pip install -r requirements.txt". How exactly are you running the tests?

aditya1702:
@leto I am running it in the same way as you are using python -m unittest discover and that too in vagrant

PR #276, opened 2016-05-06
Made changes to support Flask-DebugToolbar

aditya1702:


niranjan94:
@aditya1702 I think it would be better if you fork this repo and create a branch on that fork instead of creating a branch on the upstream repository. Also @mariobehling said that a peer review process is to be followed. Though we have write access to the repo, we're not supposed to accept/merge our own pull requests.

aditya1702:
@niranjan94 @championswimmer I get it.Should I revert the changes?

niranjan94:
@aditya1702 this is a minor change. I don't think this needs to be reverted now. Let it be. From the next PR we'll all follow the peer review process properly.

PR #309, opened 2016-05-14
Track iCal bugfix and refactored code

### LINKED ISSUE #308: bug: generate_icalender_track view crashing
Also the endpoint is `/api/v1/track/<int:track_id>/export/iCal`
It should be `/api/v1/event/<int:event_id>/tracks/<int:track_id>/export/iCal` as a track is inside event.

aviaryan:
* Fixes #308 
* Fixed a typo in get_sessions_at_event view which caused the view to crash
* Some code was refactored a/c pep8


aviaryan:
Need to write tests for calendar export to make this pass.

PR #322, opened 2016-05-16
return 404 status in API where applicable, prevent api view crashes on invalid input

### LINKED ISSUE #320: Return 404 status code in API when event doesn't exist
When an event doesn't exist, all the api requests for it should have 404 as status code. Currently it returns 200.

```
http http://localhost:5000/api/v1/event/79329/sessions
HTTP/1.0 200 OK
Content-Length: 20
Content-Type: application/json
Date: Mon, 16 May 2016 06:37:28 GMT
Server: Werkzeug/0.11.7 Python/2.7.10

{
    "sessions": []
}
```

### LINKED ISSUE #321: API views crash due to non-existant entity
For example, the following code will crash if speaker_id is non-existant. Some error handling must be implemented and 404 code should be returned if request is invalid.

```python
@app.route('/api/v1/event/speakers/<int:speaker_id>', methods=['GET'])
@auto.doc()
@cross_origin()
def get_speaker_by_id(speaker_id):
    """Return speaker data by speaker id"""
    speakers = Speaker.query.get(speaker_id)
    return jsonify(speakers.serialize)
```

As I am already working on #320 and both the issues are from the same codebase, I will patch them together and send a single PR.

aviaryan:
* Fixes #320 
Return 404 status code with appropriate message when event/speaker/session/page doesn't exist 
* Fixes #321 
API views crash due to non-existant entity - Some API views like the one to get a single speaker would crash if speaker id didn't exist. This has been fixed
* Refactored code at places

All the updates have been done in accordance with the recent discussions on API.

**Note** - This PR contains #352 

PS - codecov/patch fails for this as the whole API code is *untested* [currently](https://codecov.io/gh/fossasia/open-event-orga-server/src/91d6611e6fb659a0d1722831b1d3cd936270f27c/open_event/views/views.py). My aim today was to write tests for them but this issue needed to be worked upon first before writing tests.

### DIFF
--- a/open_event/helpers/data_getter.py
+++ b/open_event/helpers/data_getter.py
@@ -187,14 +187,14 @@ class DataGetter:
     @staticmethod
     def get_event(event_id):
         """Get event by id, None if it doesn't exist"""
-        return Event.query.filter_by(id=event_id).first()
+        return Event.query.get(event_id)
 
     @staticmethod
     def get_session(session_id):
         """Get session by id"""
-        return Session.query.filter_by(id=session_id).first()
+        return Session.query.get(session_id)
 
     @staticmethod
     def get_speaker(speaker_id):
         """Get speaker by id"""
-        return Speaker.query.filter_by(id=speaker_id).first()
+        return Speaker.query.get(speaker_id)
--- a/open_event/helpers/data_getter.py
+++ b/open_event/helpers/data_getter.py
@@ -183,3 +183,18 @@ class DataGetter:
     @staticmethod
     def get_object(db_model, object_id):
         return db_model.query.get(object_id)
+
+    @staticmethod
+    def get_event(event_id):
+        """Check if event with the id exists"""
+        return Event.query.filter_by(id=event_id).first()
+
+    @staticmethod
+    def get_session(session_id):
+        """Check if session with the id exists"""
+        return Session.query.filter_by(id=session_id).first()
+
+    @staticmethod
+    def get_speaker(speaker_id):
+        """Get speaker with id"""
+        return Speaker.query.filter_by(id=speaker_id).first()
--- a/open_event/helpers/data_getter.py
+++ b/open_event/helpers/data_getter.py
@@ -186,15 +186,15 @@ class DataGetter:
 
     @staticmethod
     def get_event(event_id):
-        """Check if event with the id exists"""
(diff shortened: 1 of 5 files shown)


aviaryan:
@juslee Please review. Sorry the codecov/patch tests fails. To make that pass, I will have to write tests for API code which I plan to do in the coming days.

open_event/helpers/data_getter.py
```diff
@@ -183,3 +183,18 @@ def get_association_by_event_and_user(event_id, user_id):
     @staticmethod
     def get_object(db_model, object_id):
         return db_model.query.get(object_id)
+
+    @staticmethod
+    def get_event(event_id):
+        """Get event by id, None if it doesn't exist"""
+        return Event.query.filter_by(id=event_id).first()
```
leto: This is more efficiently written as: `Event.query.get(event_id)` , same goes for the two instances below

leto:
This is a good start. The name `event_status_code` is a bit misleading. If an event exists but not a track, you will return a 200 instead of a 404. I suggest renaming the function `status_code` and if gets passed a positive integer, return 200, if it gets passed None/0, return 404. You can then pass the count of the number of objects in the JSON to `status_code`

aviaryan:
@leto Thanks for your review. I am working on the improvements.

> If an event exists but not a track, you will return a 200 instead of a 404.

Sorry but I don't follow. Can you provide an example ?

Also I wanted to ask, what to return when there are no tracks in `/api/v1/event/<int:event_id>/tracks`. 200 or 404 ? Currently I am returning 200.

juslee:
@shivamMg refer to @aviaryan question.

aviaryan:
Only one doubt remains in the API part. See https://github.com/fossasia/open-event-orga-server/issues/320#issuecomment-220539242

shivamMg:
> Also I wanted to ask, what to return when there are no tracks in /api/v1/event/<int:event_id>/tracks (event_id exists). 200 or 404 ? Currently I am returning 200.

Returning 200 with the following would be correct.

```json
{
    "tracks": []
}
```

aviaryan:
@juslee @shivamMg I have updated the PR with the improvements discussed in #320 and #336 . Also I have updated the first post with more details of this pull. Please review.

PR #329, opened 2016-05-17
Use env var to get app config

### LINKED ISSUE #328: Make app config configurable from an environment var, fix broken configs
Currently the project runs by default on ProductionConfig which has DEBUG disabled. This is not convenient for development purposes as every change in code requires re-running the server.

Solution - Make app config configurable from an environment variable. If no env var is set, the default (ProductionConfig ?) will be used. 

Also there are some configs like DevelopmentConfig, StagingConfig which don't work at all as no SQLALCHEMY_DATABASE_URI is defined. Fix them.

aviaryan:
* Fixes #328
* The env var is APP_CONFIG. Set it to `config.DevelopmentConfig` for development purposes.
* Removed Flask Debug Toolbar from Production

### DIFF
--- a/config.py
+++ b/config.py
@@ -10,14 +10,17 @@ class Config(object):
     CSRF_ENABLED = True
     CORS_HEADERS = 'Content-Type'
     SQLALCHEMY_TRACK_MODIFICATIONS = True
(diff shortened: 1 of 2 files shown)


aviaryan:
@juslee @leto Please review

PR #370, opened 2016-05-24
Added Track View

aditya1702:
#346


juslee:
Check why CI failed. @aditya1702

aditya1702:
@juslee The CI for the gsoc2016 branch fails right now since we are making changes in the view functions and the tests are modeled after the old views

SaptakS:
@juslee we are creating new views accroding to the new workflow. Once thats completed we will write the tests for the new views. Presently it works on the previous views.. So the build fails.

Issue #368, opened 2016-05-23
Add Map


rafalkowalski:
#310 

Map should be added and position from map should be saved to event

aditya1702:
@rafalkowalski @SaptakS Do we need latitude and longitude fields on the event form?

rafalkowalski:
i think only location will be enough

rafalkowalski:
TODO #310 
- [ ] Add possibility to save location name to database, longitude and latitude
- [ ] Remove button GO! it's ugly we need a changing listener if you put Singapore location automatically location should be changed on the map
- [ ] Move Map on the right

PR #374, opened 2016-05-24
Added Map to Event

aditya1702:
#368 
map2


rafalkowalski:
Great! Put Map on the right. 
Fill Location field dynamically.(Moving markerker), Or if you put info to Location automatically marker should be changed
Remove geocode.

aditya1702:
@rafalkowalski I was thinking if we could keep the geocode? It would be a nice feature

rafalkowalski:
Location filed it will work like geocode so we don't need a geocode

rafalkowalski:
@aditya1702 is it ok?

aditya1702:
@rafalkowalski Yes. Although I can seem to move the map to the right. Could you merge it and do it?

PR #382, opened 2016-05-24
Fix Dragging error in Schedule UI and add auto scroll

### LINKED ISSUE #381: Dragging error in Schedule UI
If the session is dragged to the left beyond the calendar, the session is translated.
screenshot from 2016-05-24 17 36 16

As we can see the session lies in an awkward position. We need to prevent this.

niranjan94:
This fixes #381 . Dragging error is fixed and the timeline will now auto scroll while dragging.

### DIFF
--- a/open_event/static/css/admin/scheduling.css
+++ b/open_event/static/css/admin/scheduling.css
@@ -23,6 +23,7 @@
     top: 120px;
     padding: 5px;
     color: white;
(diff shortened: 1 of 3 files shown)


niranjan94:
@SaptakS please review.

niranjan94:
@juslee this can be merged into `gsoc2016`

PR #384, opened 2016-05-24
give distinct names to related api models, fixed #383

### LINKED ISSUE #383: Wrong session swagger docs
The swagger docs for sessions have wrong model schema . This is only happening with sessions, no other models.
```
[
  {
    "id": 0,
    "title": "string"
  }
]
```
screenshot from 2016-05-24 18 10 06

aviaryan:
Since flask-restplus auto documents every model created, so the related model named 'session' created in speaker.py was shadowing the real session model. This was the reason for #383 .
To fix this, I have given proper names to the related models. 

@shivamMg @juslee please review

### DIFF
--- a/open_event/api/languages.py
+++ b/open_event/api/languages.py
@@ -6,7 +6,7 @@ from .helpers import get_object_list, get_object_or_404, get_object_in_event
 
 api = Namespace('languages', description='languages', path='/')
 
-LANGUAGE = api.model('language', {
+LANGUAGE = api.model('Language', {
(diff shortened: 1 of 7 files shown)


shivamMg:
@aviaryan Nice :+1: . Similar names for api models must have overridden each other. Also can you change api model names to title case for languages, levels, microlocations and sponsors. It'll make them uniform since all others are in title case.

aviaryan:
> Also can you change api model names to title case for languages, levels, microlocations and sponsors. 

:+1:

PR #422, opened 2016-05-26
Fix UI bugs on the Scheduler including text overflow.

### LINKED ISSUE #408: Session element overflow and overlap
- The sessions overlap one another.
- The session name is not displayed correctly 

capture

- The session element is overflowing (to the left of) the designated track div.
- The session timings are overflowing due to lack of space

capture2

niranjan94:
This PR fixes #408 . The time is now displayed in a tooltip instead. Text is automatically ellipsized based on available space. Better colors are generated.


niranjan94:
@juslee @rafalkowalski this is ready to be merged.

PR #432, opened 2016-05-27
Modifications to Event map

aditya1702:
#368


aditya1702:
@rafalkowalski Done can be merged

PR #473, opened 2016-05-31
Use `PAGE_PARAMS` to specify page parameters

shivamMg:
- Provide description and type for page params
- Minor changes to documentation

https://github.com/fossasia/open-event-orga-server/issues/464 and https://github.com/fossasia/open-event-orga-server/issues/472


open_event/api/utils.py
```diff
@@ -1,5 +1,10 @@
 from flask_restplus import Model, fields, reqparse
 
+# Parameters for a paginated response
+PAGE_PARAMS = {
+    'start': {'description': 'ID to begin from', 'type': int},
```
aviaryan: `start` is not the ID but the position. Say if event 3 has sessions with ids 3,5,7,9,10 then for `start=3` everything after session_id=7 in event 3 is elligible for return.

open_event/api/utils.py
```diff
@@ -1,5 +1,10 @@
...
+    'limit': {'description': 'Number of results to be retrieved', 'type': int},
```
aviaryan: Max number, max count..

shivamMg:
@aviaryan Thanks for clearing that up. I guess the start parameter shows where relative ids would have been advantageous.

@juslee Please take a look. It just introduces a constant and minor changes to existing code.

aviaryan:
@shivamMg hey, can you add the default values of start and limit (1 and 20 respectively) in the PAGE_PARAMS. 

@juslee To accelerate development, can we merge PRs ourselves if one or two other collaborators have verified the patch.

shivamMg:
@aviaryan Nice catch. Done :+1:

aviaryan:
@shivamMg please resolve the conflicts

PR #487, opened 2016-06-01
'Fix Overlaps' functionality's inefficiency has been resolved

### LINKED ISSUE #481: 'Fix Overlaps' functionality has become much slower
The Fix overlaps functionality seems to have become slower than before causing an approximate of `3` seconds of UI blocking. 

Possible solutions would be, rework of the `collision` function. Profiling might give us more clues into what is causing the CPU load. 

This was first implemented in #427 by @niranjan94 .

niranjan94:
The Fix overlaps functionality seemed to have become slower than before causing an approximate of `3` seconds of UI blocking. 

The problem was due to some events that were broadcasted unnecessarily during session rearrangement. 

Resolves #481 .

@juslee @SaptakS @rafalkowalski this is ready for merging into `gsoc2016`. Please review and merge.


SaptakS:
Seems good. Merging.

Issue #500, opened 2016-06-02
Update user details doesn't work if sign in via google or facebook


rafalkowalski:
- [ ] create UserDetails model during sign up

SaptakS:
@rafalkowalski this bug occurs on logging through fb or gmail,right?

rafalkowalski:
yes

PR #532, opened 2016-06-04
fix: add listener to create user_details table, solve #500

rafalkowalski:
#500 Please review


SaptakS:
Seems good to merge.. Merging.

PR #542, opened 2016-06-04
Fix datepicker and moment JS links

shivamMg:


SaptakS:
Seems good to merge. Merging

Issue #547, opened 2016-06-05
Do not require username, only use email address as login


mariobehling:
The signup pages requires a username. Not sure why this is necessary. Please compare to sign up procedure of eventbrite for example. Simply use email address.

rafalkowalski:
@aditya1702 Please check

aditya1702:
Ok will do

aditya1702:
@mariobehling The register page will take in the user name just for the data but the login page will have only email and password fields. Is this ok?

mariobehling:
@aditya1702 No need for username. Email is the username. Please keep it simple. Why make it complicated? We referred many times to other service Eventbrite, Eventnook, any social media. Please check there again and keep everything super easy for users in future.

aditya1702:
@mariobehling Ok

PR #554, opened 2016-06-05
Use only email for login and signup purposes

aditya1702:
#547 
login1
login2


aditya1702:
@mariobehling please review

PR #558, opened 2016-06-05
Added Node.js/bower info to Heroku doc

niranjan94:
Added information about setting up proper buildpacks for enabling bower on the Heroku app to the Heroku doc.

@rafalkowalski please review and merge.


rafalkowalski:
@niranjan94 awesome!

PR #563, opened 2016-06-05
Fixed track tests

### LINKED ISSUE #562: Tests on development branch failing
This is due to change in track model in #560

aviaryan:
Fixes #562

@aditya1702 @SaptakS please review


aditya1702:
@aviaryan the change had to be made. There was no other alternative. I understand the APIs will fail so in the future whenever I am making any change to the database I will notify you immediately :)

aviaryan:
@aditya1702 that should work. Moreover most changes to models will require only little change in the api and api_tests to fix so we can discuss this over the issue or PR.

PR #574, opened 2016-06-05
add email field to super_admin created by create_db

### LINKED ISSUE #572: Set email in super_admin created by create_db.py
The user created by create_db.py has no email set.
So we can now not login into the server as it requires email to login .

aviaryan:
Fixes #572 

@rafalkowalski @niranjan94 @aditya1702 please review

On a side note, I am surprised how the code is working after login field removed from user model. 
the create_super_admin function is using `user.login` and runs without issues. 
There are other places too which are using user.login and they too work.  Do you guys have any idea ?

### DIFF
--- a/create_db.py
+++ b/create_db.py
@@ -4,7 +4,7 @@ from open_event.helpers.data import DataManager
 
 
 def create_default_user():
-    print "Your login is 'super_admin'"
+    print "Your login is 'super_admin@gmail.com' (super_admin)"
     password = raw_input("Please enter password of Super Admin user")
     DataManager.create_super_admin(password)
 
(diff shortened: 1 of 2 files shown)


aditya1702:
@aviaryan This looks fine.where else user.login is used?

aviaryan:
> And why havent u removed user.login=super_admin?

As I said, the code works without even removing it. Trying to set an undefined property in the model should have thrown an exception, right ?

> This looks fine.where else user.login is used?

* get_event_owner of DataGetter
* some gentelella template files (/gentelella/admin/menu.html , gentelella/admin/profile/edit.html , )

Issue #546, opened 2016-06-05
Sign in with Google sends me to localhost


mariobehling:
http://open-event.herokuapp.com/

http://localhost:8001/gCallback?state=KSWNUPskXcFa92AMQ0NSz5PjPhTWoU&code=4/sLamGaCxmCm5HSPpuZzLNBoX-BPsihmvFc4KaluuHJw#

rafalkowalski:
@mariobehling I will fix it

aditya1702:
@rafalkowalski This is because I have included localhost as the address in my Google Developers Console through which i acquired the client secrets and ids. Let me do this I will add the address for the heroku app

rafalkowalski:
@aditya1702 No! It shouldn't be manually added

aditya1702:
@rafalkowalski I dont think there is any other way. http://stackoverflow.com/questions/23717452/how-to-add-multiple-redirect-uris-for-google-oauth-2

aditya1702:
@rafalkowalski I am not talking about adding it in the code...I am talking about adding it to the list of redirect uris in Google Developers Console. If it is not in that list it wont get redirected even if we add that url to our code.

rafalkowalski:
ok i didn't know

PR #587, opened 2016-06-06
Create redirect URI dynamically(Fix google login)

aditya1702:
#546


aditya1702:
@mariobehling this removes the hardcoded url and creates it dynamically at run time. However as I said before that url must be included in the list of redirect uris of the Google Developers Console through which I obtained the Client details and client secret.
screen1
So when our final url is generated for open-event I can add it to the list. But I have tested the code and its working.
Similar thing for facebook login
screen2

Issue #591, opened 2016-06-06
Update closing_datetime value to database


aditya1702:
Presently the Update button is not saving the closing_datetime value to database
![screen4](https://cloud.githubusercontent.com/assets/15054664/15822194/f227d090-2c11-11e6-88e2-6a83b0e4d962.png)
The url is not created

aditya1702:
@rafalkowalski No it wasnt . Now it is. I will change the value to 'POST' in my pull after which you can merge

PR #593, opened 2016-06-06
Saved closing_datetime value to database

aditya1702:
#591

### DIFF
--- a/open_event/views/admin/models_views/events.py
+++ b/open_event/views/admin/models_views/events.py
@@ -2,9 +2,9 @@ from flask import request, url_for, redirect
 from flask_admin import expose
 from flask_admin.contrib.sqla import ModelView
 from flask.ext import login
-from ....helpers.data import DataManager
+from ....helpers.data import DataManager, save_to_db
 from ....helpers.data_getter import DataGetter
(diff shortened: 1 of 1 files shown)


aditya1702:
@rafalkowalski @SaptakS please review

open_event/views/admin/models_views/events.py
```diff
@@ -60,3 +60,10 @@ def completed_events_view(self):
     def current_view(self):
         events = DataGetter.get_current_events()
         return self.render('/gentelella/admin/event/current_events.html', events=events)
+
+    @expose('/<int:event_id>/update/', methods=('GET','POST'))
```
SaptakS: @aditya1702 shouldn't we change only on POST? Because using GET I believe it would be easily modifiable by anybody and giving permissions would be difficult. @rafalkowalski What do you think?

rafalkowalski: I'm agree @SaptakS

aditya1702:
@rafalkowalski removed 'GET'. can be merged

SaptakS:
Seems good to merge. Merging.

Issue #590, opened 2016-06-06
Check password also while logging in


aditya1702:
The bug is that if we register an email address and password and then do a login , then there is no check on the password. If the email address is correct then the user can log in irrespective of what password he/she has entered.

mariobehling:
Please describe this bug. What does it mean? Any relevant screenshot?

niranjan94:
@mariobehling it means anyone can log into the administration panel by just typing someone's email address and some random text for the password.

PR #599, opened 2016-06-07
Check password also during login

aditya1702:
#590


aditya1702:
@rafalkowalski @SaptakS @niranjan94 please review

niranjan94:
Seems to be good. Merging.

PR #609, opened 2016-06-07
Added custom DateTime field in API

### LINKED ISSUE #601: Fix POST API testing of event and session model by using custom DateTime field
Both Event and session model includes a datetime field which currently can't be tested because it is not possible to write datetime fields as string from SQLAlchemy in a SQLite database (used in Testing). 

However, this issue can be fixed by a using a custom field for DateTime that returns the DateTime object on POST requests. Using a custom DateTime field will also give more control over the datetime 'format'. 

I am currently think about 'YYYY-MM-DD HH:MM:SS' as the DateTime format to be used in the API . 
@niranjan94 @SaptakS Is the format ok w.r.t UI ?

EDIT - Just found out restplus doesn't actually validate the (inbuilt) DateTime fields. It lets it go and psycopg throws the exception. (https://bpaste.net/show/e332f0095399)

aviaryan:
Fixes #601 

DateTime format for API is now (strictly) 'YYYY-MM-DD HH:MM:SS'  .  Any datetime which is not in this format will be rejected by POST API.

@shivamMg @rafalkowalski please review


niranjan94:
@aviaryan this PR has some conflicts. Please resolve them.

aviaryan:
@niranjan94 Resolved

niranjan94:
Seems to be good. Merging.

Issue #641, opened 2016-06-08
Fix case in API namespace descriptions


aviaryan:
Some namespace descriptions are in lowercase ('microlocations') whereas some are in titlecase ('Events'). Make them all titlecase.

Issue #610, opened 2016-06-07
Use JWT for API authentication


aviaryan:
See #603 for previous version

[GSoC docs](https://github.com/fossasia/open-event/blob/master/docs/GSOC2016/rest-api.md) says we should be using JWT for API authorizations. 

I read about JWT today and it looks really good . The main advantage I find of JWT is that it is self-contained. A JWT contains all information about the user which eliminates the need to look in database for every request.

These links should be useful
Intro to JWT https://jwt.io/introduction/
On using JWT https://auth0.com/learn/json-web-tokens/
Flask-JWT package docs https://pythonhosted.org/Flask-JWT/

I plan to implement jwt based authentication in the API using Flask-JWT. 

Note - This issue is an extension to #603 (token based auth). I didn't want to delete what has been said there so opened a new one.

aviaryan:
@shivamMg I have almost done this issue locally. In it, there is a login endpoint `api/v2/login` where user sends username and password and gets the JWT . 
Should I add it to swagger UI or not ?

shivamMg:
I think it should be included in the Swagger UI.

aviaryan:
@shivamMg Okay. I have included it now.

PR #642, opened 2016-06-08
Added JWT based authentication

### LINKED ISSUE #603: Add token based authentication in API
A (access) token based auth should be added side-by-side with the current basic auth in the API.
Here are my views on it, 

token - will be a random string stored in the User model as field `token`

To get token the user will send a post request with the credentials to the following urls - 
1. The current url for UI login is `\admin\login` . The view function responsible for it can store the token as cookie if username-password is correct. (as is already done in PR #600 by @niranjan94 )
2. To get token using API, we can create a url like `\api\v2\login` where user details and token will be returned in case of right username-password. 

Any request can then use the token to access protected methods by setting the header `Authorization` 
as `Token <token>`. 

Once a token is created for a user, it will be returned as-is in all login requests and not be re-created. (This will allow a user to be login'd from android and serverUI at the same time). 
Only on 'forgot password' will the the access token be re-created. 

@niranjan94 @shivamMg @rafalkowalski views ?

aviaryan:
Fixes #603, #610 and includes fix for #641

Expiry time of token is set to 1 day.
Auth url is `api/v2/login`
Send there a POST request with email and password to get access_token . 
```
{
  "email": "email@gmail.com",
  "password": "test"
}
```
```
{
  "access_token": "string"
}
```

Include the token in Authorization header as "JWT _token_" for accessing authorized methods.

Note - I have not removed the basic auth. Both work. This is becasue many good services like GitHub allow their APIs to be accessed by both, access token and username:password authentication. Let's follow their example.

### DIFF
--- a/open_event/__init__.py
+++ b/open_event/__init__.py
@@ -58,7 +58,7 @@ def create_app():
     # set up jwt
     app.config['JWT_AUTH_USERNAME_KEY'] = 'email'
     app.config['JWT_EXPIRATION_DELTA'] = timedelta(seconds=84400)
-    app.config['JWT_AUTH_URL_RULE'] = '/admin/jwtauth'  # not used anyway
+    app.config['JWT_AUTH_URL_RULE'] = None
     jwt = JWT(app, jwt_authenticate, jwt_identity)
 
     admin_view = AdminView("Open Event")
--- a/open_event/__init__.py
+++ b/open_event/__init__.py
@@ -12,12 +12,16 @@ from flask.ext.migrate import Migrate, MigrateCommand
 from flask.ext.script import Manager
 from flask import render_template
 from flask import request
+from flask.ext.jwt import JWT
+from datetime import timedelta
 
 from icalendar import Calendar, Event
 from flask_debugtoolbar import DebugToolbarExtension
 
 from open_event.models import db
 from open_event.views.admin.admin import AdminView
+from helpers.jwt import jwt_authenticate, jwt_identity
+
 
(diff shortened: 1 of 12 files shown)


aviaryan:
@shivamMg @niranjan94 @rafalkowalski please review

niranjan94:
Seems to be good. Can be merged. :smile: ... @rafalkowalski @shivamMg What do you think ?

open_event/api/helpers.py
```diff
@@ -170,31 +171,69 @@ def requires_auth(f):
...
     return decorated
+
+
+@jwt_required()
+def auth_jwt():
+    """
+    A helper function that throws JWTError if JWT is not set
+    """
+    g.user = current_identity
+    pass
```
shivamMg: No need to pass

open_event/__init__.py
```diff
@@ -51,6 +55,12 @@ def create_app():
     app.logger.setLevel(logging.INFO)
     # logging.getLogger('sqlalchemy.engine').setLevel(logging.INFO)
 
+    # set up jwt
+    app.config['JWT_AUTH_USERNAME_KEY'] = 'email'
+    app.config['JWT_EXPIRATION_DELTA'] = timedelta(seconds=84400)
```
shivamMg: Shouldn't it be 86400?

aviaryan: LOL. Mental maths went wrong.

aviaryan:
@shivamMg @rafalkowalski  updated. please merge if satisfied.

PR #643, opened 2016-06-08
Option to Export timeline as PNG

### LINKED ISSUE #512: Option to Export timeline as PNG
An option could be provided to allow the organizer to export the timeline as a PNG. (Which could be printed and distributed to the participants or displayed somewhere at the venue or emailed to all registrants).

niranjan94:
An option to download the timeline as a PNG has been added. _(Which could be printed by the organizers and distributed to the participants or displayed somewhere at the venue or emailed to all registrants)._

Resolves #512.

> The initial plan was to generate a PDF. But that was not feasible, as client-side PDF generation of HTML content is still experimental. For example, [jsPDF](https://github.com/MrRio/jsPDF) supports only basic HTML content and styles are ignored.

@SaptakS @rafalkowalski please review and merge into `development`

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -23,6 +23,9 @@
     "jQuery-Smart-Wizard": "~3.3.1",
     "google-material-color": "~1.2.6",
     "datatables.net": "^1.10.12",
-    "datatables.net-bs": "^1.10.12"
+    "datatables.net-bs": "^1.10.12",
+    "html2canvas": "^0.4.1",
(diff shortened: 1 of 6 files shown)


mariobehling:
Thanks for that! Still, we also need pdf for the schedule. Let's come back to this later. I am not sure, why pdf generation would have to be clientside. We can just generate it on the server.

PR #646, opened 2016-06-08
Send email confirmation link to user on signing up

aditya1702:
screen10
#607


aditya1702:
@SaptakS @rafalkowalski please review

open_event/helpers/data.py
```diff
@@ -478,14 +478,14 @@ def remove_microlocation(microlocation_id):
         flash('You successfully delete microlocation')
 
     @staticmethod
-    def create_user(form):
-        user = User(nickname='asdf',
-                    email=form['email'])
+    def create_user(userdata):
+        user = User(nickname='afwf',
```
rafalkowalski: why every users is afwf?

SaptakS: @aditya1702 @rafalkowalski @mariobehling I think we should remove the nickname field because it isn't being used anyways.. We are logging in using the email id only.

aditya1702:
@SaptakS @rafalkowalski I too thought the same..we should remove the nickname field.So can I remove it?

aditya1702:
@SaptakS @rafalkowalski removed the nickname field and its references anywhere in the code

PR #647, opened 2016-06-08
Use JWT for API Authentication in the Scheduler UI

niranjan94:
Since JWT has been implemented in #642 . I have switched from Basic username:password authentication to the new token-based authentication. 

When a user logs into the admin panel, a token for that user is generated and stored in a cookie named `access_token` which can then be used throughout the panel for any API calls. 

A javascript helper function has been provided in `open_event/static/js/jquery/jquery.codezero.js` for reading cookies. `getCookie('access_token')` would provide the access token. 

@rafalkowalski @aviaryan @SaptakS please review and merge into `development`

### DIFF
--- a/open_event/views/admin/home.py
+++ b/open_event/views/admin/home.py
@@ -6,29 +6,12 @@ from flask.ext import login
 from flask_admin import expose
 from flask_admin.base import AdminIndexView
 from flask.ext.scrypt import generate_password_hash
-from datetime import datetime
-import jwt
 
 from ...helpers.data import DataManager, save_to_db,get_google_auth,get_facebook_auth
 from ...helpers.data_getter import DataGetter
 from ...helpers.helpers import send_email_after_account_create, send_email_with_reset_password_hash
 from open_event.helpers.oauth import OAuth, FbOAuth
 from flask import current_app
-
-def generate_token(user):
-    _jwt = current_app.extensions['jwt']
-    iat = datetime.utcnow()
-    exp = iat + current_app.config.get('JWT_EXPIRATION_DELTA')
-    nbf = iat + current_app.config.get('JWT_NOT_BEFORE_DELTA')
-    payload = {'exp': exp, 'iat': iat, 'nbf': nbf, 'identity': user.id}
-    secret = current_app.config['JWT_SECRET_KEY']
-    algorithm = current_app.config['JWT_ALGORITHM']
-    required_claims = current_app.config['JWT_REQUIRED_CLAIMS']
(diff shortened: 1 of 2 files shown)


open_event/views/admin/home.py
```diff
@@ -6,14 +6,30 @@
...
+import jwt
 
 from ...helpers.data import DataManager, save_to_db,get_google_auth,get_facebook_auth
 from ...helpers.data_getter import DataGetter
 from ...helpers.helpers import send_email_after_account_create, send_email_with_reset_password_hash
-from open_event.models.user import User
 from open_event.helpers.oauth import OAuth, FbOAuth
 from flask import current_app
 
+def generate_token(user):
```
niranjan94: This method is borrowed from Flask-JWT source code.

aviaryan: This seems complicated. 
Maybe you can just modify the /api/v2/login endpoint to look for a `next` url parameter (query param) and make the login form send the request to /api/v2/login.
If the next query param exists, then the method for /api/v2/login can call this intended_url function which can redirect to desired url. If next query param doesn't exist, it can just return the normal access_token json.

niranjan94: Yep. this is complicated. But what you said just gave me an idea.

niranjan94: @aviaryan But we cannot make the form directly post to the API endpoint ... JWT expects a json request with email and password. Form does not give json. We could make form give json by using ajax, but that would further complicate the login. And we also need to access Flask-login and create a logged in session for the user. Doing that from the API would not be right.

niranjan94: Or we could make the login method make a request to the API in the backend and retrieve an access token for us

niranjan94:
@aviaryan please review ... I had foolishly overlooked a simpler way of generating ...

aviaryan:
Hey, the authentication system of the API also checks for current session (using flask-login) if both jwt and basic auth tokens are not present. So JWT is practically not needed for the Scheduler UI. Sorry I am mentioning it now, I added the active-session based API authentication for the swagger UI but the idea that it will work even for the scheduler UI never crossed my mind. :disappointed_relieved: 
Any POST/PUT/DELETE request sent from the orga-server project is automatically authenticated by the current login'd user. 
Again, sorry for not letting you know earlier. :disappointed: 
I guess it will be better if the Auth header is completely dropped.

niranjan94:
Oh... Okay... :smile:... That's alright... No problem... No apologies necessary...

niranjan94:
I'll remove the old code in a new PR

PR #659, opened 2016-06-09
Remove unnecessary authorization header

niranjan94:
From #647:
> @aviaryan: Hey, the authentication system of the API also checks for current session (using flask-login) if both jwt and basic auth tokens are not present. So JWT is practically not needed for the Scheduler UI. Any POST/PUT/DELETE request sent from the orga-server project is automatically authenticated by the current logged in user.  I guess it will be better if the Auth header is completely dropped.

So, that means, we do not need an Authorization header while calling the API from the Admin Panel :tada: :smile: ... 

This PR removes the authorization header that was added in PR #600. 

@SaptakS @rafalkowalski @aviaryan please review and merge.

### DIFF
--- a/open_event/static/js/admin/event/scheduling.js
+++ b/open_event/static/js/admin/event/scheduling.js
@@ -751,9 +751,6 @@ $addMicrolocationForm.submit(function (event) {
         "floor": parseInt($addMicrolocationForm.find("input[name=floor]").val())
     };
     $.ajax({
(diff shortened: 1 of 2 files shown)


aviaryan:
Looks good. But before merging, can you please confirm that you have tested it with a microlocation POST API call from the admin UI; just to be sure. :worried:

PR #661, opened 2016-06-09
Add missing fields to Session model and api

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/627

@aditya1702 @aviaryan Can you please take a look


aditya1702:
@shivamMg looks good. One question: Each session will have speakers...so shouldnt we make the fields for speaker rather than a session?

shivamMg:
@aditya1702 Mario created an issue to make the organizer server schemas (models and api) to be consistent with OTS schemas: https://github.com/fossasia/open-event-orga-server/issues/624

aviaryan:
@shivamMg why is audio field.String. That field will contain a url to the audio, right ?
Also please check that the GET API returns all the new fields, and update sample data is utils_post_data. 
Thanks

shivamMg:
@aviaryan The schema at OTS doesn't exactly say it is a string ([eg](https://github.com/OpenTechSummit/2016.opentechsummit.net/blob/gh-pages/programm/json/sessions.json#L1207)), so I thought of keeping it like that. I doesn't really make sense to do the same here though, since we are not storing audios at server right now. I've defined it as a URL now and renamed it `audio_url` to make it similar to others.

aviaryan:
@shivamMg @aditya1702 In PR #666, it seems that we are uploading all types of files to server. :confused: So both PRs conflict.

In my opinion, slides (and possibly audio) can be stored on server but definelty not video. A youtube link would be better. (it can be embedded in each session page and will look awesome). Also the video   (and audio) will be uploaded by the organizer after the session so they should have no problem uploading to youtube/other_site and copying the link. 

@mariobehling @rafalkowalski your views ?

aditya1702:
@aviaryan I think that is a good idea...I can remove the video part. But for the slides and the audio I think it would be better if we store it.

SaptakS:
@aviaryan :+1:

mariobehling:
@aditya1702 @SaptakS @aviaryan Please give the option either to define a URL or upload media. Both is good.

aviaryan:
For video, I will definetly go with url (youtube). 
For audio, we can allow storing on audio on server but that won't work in the long run and we will have to purchase storage (AWS). Thus better let audio be stored as url now (archive.org can be good, has direct links. There is an "[audio tech](https://archive.org/details/audio_tech)" category for this. )
For documents,  there is not much problem uploading to server but for consistency, we can go with url for now. 

We can make uploading files (to AWS or another good service) possible in a later milestone.

@shivamMg @aditya1702 Let's decide on one thing and merge this PR. :smile:

mariobehling:
@shivamMg Of course we will run this service on a paid server. Please don't worry about storage etc. Just offer all options for all media either store on server or provide URL. This could also be an option for a paying customer only in future.

aditya1702:
@aviaryan @shivamMg @mariobehling I would go for storing media on the server in folder. During the event if a user wants to upload a file from his /her collection, he can do this easily. So I prefer going for folder

aviaryan:
After discussing with @aditya1702 , we agree on having video as external link and audio/slides as uploads. 
@shivamMg If you agree, please update this PR .

mariobehling:
where did you discuss this? Sorry, I do not know how to make it more clear. We need all media be uploadable and alternatively as a URL.

aviaryan:
@mariobehling I think I got this wrong the first time. The need is therefore to have both url and upload options in those fields. Sorry about that.

shivamMg:
@mariobehling @aviaryan OK. I'll make a new PR.

PR #675, opened 2016-06-10
Show login info in swagger ui

### LINKED ISSUE #669: Show login info in swagger UI
Since POST, PUT and DELETE requests work from the swagger ui using current login'd user as authorizer, it will be better if the show some info of active user in the UI. 
If no user is login'd, we can show links to register and login, and use next param in url to come back to swagger UI after either of the processes.

aviaryan:
Fixes #669

#### Screenshots

(click for larger view)

log1

log2

@SaptakS @niranjan94 please review


niranjan94:
Looks good. Merging.

PR #678, opened 2016-06-10
Fixed requires_auth, don't let faulty credentials breach

### LINKED ISSUE #677: requires_auth function should abort if a means of authentication fails
Suppose a user tried to send a JWT based authentication and his credentials where wrong, then the `requires_auth` function should abort and don't check active session.
This was my intention all along but there was a logic issue and somehow the tests weren't able to catch it.

aviaryan:
Fixes #677


niranjan94:
Looks good. Merging.

PR #700, opened 2016-06-11
Fix root URL crash when no user is logged in.

### LINKED ISSUE #695: UI crashed when no user is logged in and root url is opened for the first time
This looks because in data_getter , the events list is fetched for the user without checking if he is logged in. 

```python
# line 237
    @staticmethod
    def get_published_events():
        events = Event.query.join(Event.roles, aliased=True).filter_by(user_id=login.current_user.id) \
            .filter(Event.state == 'Call for papers')
        return events

    @staticmethod
    def get_current_events():
        events = Event.query.join(Event.roles, aliased=True).filter_by(user_id=login.current_user.id)\
            .filter(Event.state != 'Completed')
        return events
```

@rafalkowalski @SaptakS please look

crash

niranjan94:
Resolves #695 .

@aviaryan @rafalkowalski please review and merge into `development`

### DIFF
--- a/open_event/helpers/data_getter.py
+++ b/open_event/helpers/data_getter.py
@@ -225,7 +225,7 @@ class DataGetter:
         return events
 
     @staticmethod
-    def get_published_events():
+    def get_all_published_events():
         events = Event.query.filter(Event.state == 'Published')
         return events
 
--- a/open_event/views/views.py
+++ b/open_event/views/views.py
@@ -35,7 +35,7 @@ app = Blueprint('', __name__)
 def get_main_page():
     """Redirect to admin page"""
     call_for_papers_evs = DataGetter.get_call_for_speakers_events()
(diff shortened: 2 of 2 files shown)


shivamMg:
:+1: Merged. `get_published_events` had two definitions inside the same class. This would be a fix for that. If I understand correctly, datagetter functions with user_id in use still don't have is_authenticated check, so the issue shouldn't be closed.

aviaryan:
Well it seems that urls for completed events and current events  (like  http://localhost:5000/admin/events/current) don't allow entry without login. So maybe that is not a problem. @niranjan94 What do you say ?

niranjan94:
@aviaryan you're right. All the other datagetter functions that use the logged in user's ID are only called by view that require the user to be logged it. So, it wouldn't be an issue.

PR #714, opened 2016-06-11
feat: migrations fixed

rafalkowalski:


string-areeb:
Culprit of broken migration No.1

PR #729, opened 2016-06-11
Full custom validation in API

### LINKED ISSUE #717: Full custom payload validation for API
API (POST/PUT) currently uses a validate_payload func for validating custom fields and it works quite well. 
I suggest we should take this idea and make all fields custom. (String, Integer, Float) are the ones that remain. Then we can create a custom decorator like `api.expect`

Why ?
1. api.expect doesn't allow `null` (None when translated to Python) value as input. This is a **huge problem**. (specially for integers)
2. No need of running two validations. (api.expect and validate_payload). 
3. Consistent validation errors structure. 

Also we can return all validation errors at once instead of the first found validation error.

@shivamMg What do you think ?

aviaryan:
Fixes #717 

Purpose: Gain control over input validation and errors. 

#### Fixed in this PR

* null values in the payload of a post/put request no longer give validation errors.
* if required=True for a string field, don't accept an empty value for it.
* also don't accept null value for any field if required = True

@shivamMg  please review

### DIFF
--- a/open_event/api/custom_fields.py
+++ b/open_event/api/custom_fields.py
@@ -12,6 +12,12 @@ class CustomField(Raw):
     """
     Custom Field base class with validate feature
     """
+    __schema_type__ = 'string'
+
+    def __init__(self, *args, **kwargs):
+        super(CustomField, self).__init__(**kwargs)
+        # can be used for managing params
+
     def format(self, value):
         """
         format the text in database for output
@@ -48,7 +54,6 @@ class EmailField(CustomField):
     """
     Email field
     """
-    __schema_type__ = 'string'
     __schema_format__ = 'email'
     __schema_example__ = 'email@domain.com'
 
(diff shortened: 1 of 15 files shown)


open_event/api/custom_fields.py
```diff
@@ -133,3 +135,52 @@ def validate(self, value):
...
+    """
+    __schema_type__ = 'integer'
+    __schema_format__ = 'int'
+    __schema_example__ = 0
+
+    def validate(self, value):
+        if value is None:
+            return self.validate_empty()
+        if type(value) == int and value >= 0:
+            # No <0 int values right now
```
shivamMg: Better to call it `PositiveInteger` then.

aviaryan: @shivamMg 
Probably all the Integer that we will use have `>0` condition. So I didn't want to name every field as `PositiveInteger`. I have updated the code and added a param `positive`. `positive` is True by default, and can be made false if ever a negative integer field is needed.

PR #733, opened 2016-06-11
UI enhancements

SaptakS:


open_event/static/css/custom.css
```diff
@@ -4185,6 +4185,7 @@ fieldset {
 
 .actionBar a {
   margin: 0 3px;
+  float:right;
```
rafalkowalski: please don't do this, change in different file

Issue #747, opened 2016-06-12
Make POST and PUT requests functionable from DAO


aviaryan:
Make it such that calling DAO.method with payload is the only statement required to successfully execute POST and PUT requests.
Why ?
This allows us to internally do requests on the API. Or in other words, send a *VERB* request when inside another request. 

Also this is a required step to proceed #728

PR #757, opened 2016-06-12
Made DAOs in API independent/self-sufficient

### LINKED ISSUE #746: Create a Data Access Object for Event
This will make it consistent with other service models . Event's DAO will have the same methods as other DAO. 

This is also required to proceed #728

aviaryan:
Fixes #746 and #747 

This PR makes the DAO independent meaning if you have the params of a request, you can now use the DAO to successfully execute any type of request and get the return value.
So it will add the ability to internally call an API and execute any type of request.

### DIFF
--- a/open_event/api/events.py
+++ b/open_event/api/events.py
@@ -5,9 +5,9 @@ import custom_fields as fields
 from open_event.models.event import Event as EventModel, EventsUsers
 from open_event.models.user import ADMIN, SUPERADMIN
 from .helpers import get_object_list, get_object_or_404, get_paginated_list,\
-    requires_auth, validate_payload, update_model
+    requires_auth, update_model
 from utils import PAGINATED_MODEL, PaginatedResourceBase, PAGE_PARAMS, \
-    POST_RESPONSES, PUT_RESPONSES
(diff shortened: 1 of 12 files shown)


aviaryan:
@shivamMg please review

open_event/api/events.py
```diff
@@ -38,14 +38,59 @@
...
 
-# Helper function to fix datetime event payload
-# TODO: Make ServiceDAO more versatile so that event can use it
-def fix_payload(data):
-    data['start_time'] = EVENT_POST['start_time'].from_str(data['start_time'])
-    data['end_time'] = EVENT_POST['end_time'].from_str(data['end_time'])
-    data['closing_datetime'] = EVENT_POST['closing_datetime'].from_str(
-        data['closing_datetime'])
-    return data
+class EventDAO(ServiceDAO):
```
shivamMg: It would be better to create a BaseDAO, and then extend it to EventDAO and ServiceDAO. What do you say?

aviaryan: Sounds good. BaseDAO can be used for a parent-level (or independent) model.

shivamMg:
@aviaryan Nice work.

PR #772, opened 2016-06-13
Export event as json (zip)

### LINKED ISSUE #758: Feature to export event as json
Parent issue: #728 

GET request to `\events\<id>\export\json` will export the event as a set of json files archived in a zip and start its download. 

Files in the zip will be like 
```
event.json
sessions.json
speakers.json
microlocations.json
... and so on
```

The information in these JSON files will be the same as what is returned by corresponding GET requests.

aviaryan:
Fixes #758 

URL is `api/v2/events/<id>/export/json` . Going to it will start the download as `event<id>.zip`

@mariobehling @shivamMg please review

PS - I have not added this to the Swagger spec as I don't think this is an API. (It is more like a download link)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ app.db
 # Static files at 'open_event/static/admin/lib'
 lib/
 node_modules
+
+# event export files
+static/exports/event*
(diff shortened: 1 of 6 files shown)


shivamMg:
Looks cool. Merging :+1:

mariobehling:
Good job! Please implement this in the UI as well: https://github.com/fossasia/open-event-orga-server/issues/780

Issue #770, opened 2016-06-13
Create a doc regarding what each role is allowed to do


aditya1702:
#623

PR #775, opened 2016-06-13
Create documentation regarding roles

aditya1702:
#770


aditya1702:
@rafalkowalski @mariobehling I have tried created the list of URLs each role can access.

PR #809, opened 2016-06-14
Removed Login error

aditya1702:
The login error because of user.salt was there because on logging in through Google or Facebook no password is created for the user. I have added a password creation template for the users who choose to login through Google and Facebook.
new_password
#805


aditya1702:
@rafalkowalski @SaptakS can be merged. Please review

PR #884, opened 2016-06-16
Event wizard: S5 wrong session/speaker form stored.

niranjan94:
There was a small error in the javascript causing only required title, name and email to be required irrespective of the user's choice. 

@SaptakS please review and merge


SaptakS:
Seems good.

Issue #882, opened 2016-06-16
On location bottom column: Show live events and cities only


mariobehling:
It looks like currently all events no matter what stage, also draft and past, show up on the location bottom column.

* [x] Ensure only cities show up (sorted by cities with the most events on top)
* [x] Only show live events

![screenshot from 2016-06-16 08 56 37](https://cloud.githubusercontent.com/assets/1583873/16108001/a15edd5c-33a0-11e6-8cd4-55ce4472c6ae.png)

mariobehling:
At the moment it is still the address that shows up, but the goal is that only the city to shows up.

rafalkowalski:
@mariobehling can i assume that city is always first?

mariobehling:
@rafalkowalski No, you need to extract that from the address. There is a library with all kinds of cities and countries available that is related to OpenStreetMap. Not sure, what was the link though. It is not as simple as you might think.

rafalkowalski:
@mariobehling @SaptakS @niranjan94 
I don't have an idea how can i solve this issue  without adding additional field in event creation city. Searching city in text is not efficient. I wasted too much time to think about it. So i need to add a city field to step 1 If you have an idea please inform me

niranjan94:
@rafalkowalski location is a compulsory field for an event to be made live. So, if we have location, we have lat and long. We can get city from that.

niranjan94:
We can make a call to 
`https://maps.googleapis.com/maps/api/geocode/json?latlng=40.714224,-73.961452` and get the city if and when required. 

Or, we could event create a `city` field, and make a call to that API during event creation/update and store the city

PR #892, opened 2016-06-16
feat: locations in footer

rafalkowalski:
#882


mariobehling:
What does this merge do? I only see that locations are not showing up now.

PR #903, opened 2016-06-16
Add a check list to the specific event dashboard

### LINKED ISSUE #848: Add a check list to the specific event dashboard
How it should work:

**Checklist**
* [x] Each wizard step gets a row.
    * [x] Basic Details
    * [x] Sponsors
    * [x] Session, Tracks, Rooms
    * [x] Call for Speakers
    * [x] Session Forms

**Color Changes and Emoticons**
* [x] If the all essential elements have been filled in the widget background turns green and there is a tick at the end
* [x] If some parts are missing, but some basic information is there, the widget background turns yellow and there must be a small friendly warning element
* [x] If basic information is missing, the widget background turns red
* [x] If form is not getting used, e.g. "Call for papers is turned off", the widget background turns gray

**First step**
* Implement a simple widget without colors and make a row and edit button for each element

**Second step**
* Define for each form, what should be included to make it yellow or green.

**Third step**
* Document it in /docs

**Forth step**
* Implement it

Here an example. No need to do exactly the same, just use our design.

screenshot from 2016-06-15 14 29 24

niranjan94:
Resolves #848

#### Screenshot:
capture

@SaptakS @rafalkowalski @mariobehling please review and merge


niranjan94:
I'll check the tests

mariobehling:
Ok, great. Let's define more details of this feature step by step and add tests later.

mariobehling:
Please fix travis build first.

PR #927, opened 2016-06-17
Create separate script for populating database

shivamMg:
@niranjan94 Please review and merge. I've created a separate `populate_db.py` that is run when `create_db.py` is executed. And it can be executed separately too.

### DIFF
--- a/create_db.py
+++ b/create_db.py
@@ -3,18 +3,9 @@ import sys
 
 from open_event import current_app
 from open_event.models import db
-from open_event.models.role import Role
-from open_event.models.service import Service
-from open_event.models.permission import Permission
 
-from open_event.models.track import Track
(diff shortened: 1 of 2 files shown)


mariobehling:
Where is the relevant issue? Where is the documentation about this?

shivamMg:
@mariobehling It did not introduce any new functionality or patch. Creating roles and permissions required running `create_db.py` which couldn't have been done at the heroku app. So I split the create_db.py file into two, to separate creation of roles. It was a minor issue so I didn't think of creating a new issue.

mariobehling:
Ok, thanks. Got it. If this is too small for an issue, there was a closed issue, that you could have reopened and added this info to. Or alternatively provide sufficient info in the PR. Thank you.

Issue #631, opened 2016-06-07
Session Wizard Step 1: Add Session Details


mariobehling:
Wizard Parent Issue: https://github.com/fossasia/open-event-orga-server/issues/602
![screenshot from 2016-06-08 00 20 24](https://cloud.githubusercontent.com/assets/9530293/15870966/edd3f7a4-2d0f-11e6-8150-a8a457d4ea04.png)

The session details need to be in a logic order and reflect the idea for the event. Some events require less information, some more. 

* [x] Prerequisite is https://github.com/fossasia/open-event-orga-server/issues/638 The organizer therefore needs to be able to edit the session form for each event.
* [x] Prerequisite is https://github.com/fossasia/open-event-orga-server/issues/638 The event organizer can choose which form elements are required.

The standard form for sessions should include:
* [x] Title
* [x] Subtitle
* [x] Short Abstract
* [x] Long Abstract
* [x] Session Type (this choice depends on the session type defined by organizers, e.g. workshop, talk etc. https://github.com/fossasia/open-event-orga-server/issues/635)
* [x] Track (this choice depends on the tracks defined by organizers https://github.com/fossasia/open-event-orga-server/issues/636)
* [x] Possibility to upload slides
* [x] Ability to add multiple links with + button and choose type (e.g. Project Website, Twitter, Facebook, Github etc.)
* [x] Other Comments for Organizers (will not be shown to public)

Other
* [x] Please take out description of menu points. It is redundant, e.g. "details about sessions" etc.

SaptakS:
Menu point description removed https://github.com/fossasia/open-event-orga-server/pull/667

mariobehling:
Please implement all fields in the list of this issue and follow the sorting order of the sort order of the list.

SaptakS:
@mariobehling ohk. Will do it.

SaptakS:
> Ability to add multiple links with + button and choose type (e.g. Project Website, Twitter, Facebook, Github etc.)

@mariobehling what exactly do you mean by these? Will this be available by default or should this be added from custom form as well?

mariobehling:
@SaptakS We already implemented this option in step 5 of the Create Event Wizard. Just ensure anything that is marked as "Include" and "Required" is handled exactly as marked by the organizer.

SaptakS:
Closing this issue. Follow up for showing comments only to organizers only https://github.com/fossasia/open-event-orga-server/issues/1086

Issue #602, opened 2016-06-07
Create Session Wizard


SaptakS:
To add Speaker details and add/invite other speakers for a session, I guess it is best to have a session Wizard.

The wizard will have 3 steps:-
- [x] Add session details https://github.com/fossasia/open-event-orga-server/issues/631
- [x] Add speaker/self details https://github.com/fossasia/open-event-orga-server/issues/632
- [x] Invite other speakers for session #792 

At the beginning, the session will be added with a "pending" state to the DB. The organizer can either "accept" or "reject" the session.

The Speaker can also edit the session and Invite Speakers later.

This is a rough design workflow
![screenshot from 2016-06-07 12 09 09](https://cloud.githubusercontent.com/assets/9530293/15848149/b7c57a50-2ca8-11e6-92e4-7459f830d475.png)

@rafalkowalski @aditya1702 @mariobehling @juslee What do you think?

mariobehling:
@SaptakS Looks good. Speakers can of course also just "answer the call for papers" and propose a session without getting an invite. Apart from that organizers should be able to add sessions as well. Do you have a screenshot how things should look like?

SaptakS:
@mariobehling It isn't ready yet... I will provide a screenshot as soon as I get it done. Thanks. :)

SaptakS:
@mariobehling 
![screenshot from 2016-06-08 00 20 24](https://cloud.githubusercontent.com/assets/9530293/15870966/edd3f7a4-2d0f-11e6-8150-a8a457d4ea04.png)
![screenshot from 2016-06-08 00 20 29](https://cloud.githubusercontent.com/assets/9530293/15870967/edd6786c-2d0f-11e6-862e-c12f7fcb330c.png)
![screenshot from 2016-06-08 00 20 33](https://cloud.githubusercontent.com/assets/9530293/15870968/ede14f30-2d0f-11e6-9eec-c8873e3ef36d.png)

mariobehling:
I am not sure, if the speaker details should just be kept in the profile. Did you have a look at the speakers form of the last FOSSASIA?

mariobehling:
Start and End time in first step does not make sense and order does not make sense entirely. Let's create sub issues for this. https://github.com/fossasia/open-event-orga-server/issues/631

SaptakS:
@mariobehling what order do you suggest? And I don't understand what you mean to say by
" if the speaker details should just be kept in the profile "

mariobehling:
[...]

PR #930, opened 2016-06-17
Improved Session-UI, added form validation

SaptakS:
#602 #631 

session-wizard-step1
session-wizard-step2


SaptakS:
@mariobehling @niranjan94 please review.

PR #940, opened 2016-06-17
API: Nested fields show null if not set

### LINKED ISSUE #937: If nested fields are null, show null instead of their structure
Take for example Sessions API. If `track` is None (not set), it shows 

```json
  "title": "string",
  "track": {
    "id": null,
    "name": null
  },
  "video": "string"
```

It would be better if it shows 

```json
  "title": "string",
  "track": null,
  "video": "string"
```

@shivamMg @mariobehling What do you think ?

aviaryan:
Fixes #937


aviaryan:
@shivamMg  please review

shivamMg:
@aviaryan Done!

PR #947, opened 2016-06-17
Added creator field in Event API

### LINKED ISSUE #942: Add creator field in Event API
Points - 

* Creator field cannot be modified by PUT
* Creator field cannot be set by POST. It is auto-set at POST request.

aviaryan:
Fixes #942 
Also fixes the way event-user relationship was handled when an event is created. It is now updated to the current permissions system. 

@shivamMg please review


shivamMg:
Looks good :+1: Merging.

Issue #1003, opened 2016-06-18
Implement view to create a Super Admin


aditya1702:

mariobehling:
Where and how do you plan to implement this? There are edit buttons in the user menu: https://open-event.herokuapp.com/admin/users/ Maybe here?

aditya1702:
@mariobehling I am thinking of displaying a button on the header next to the Browse Events..something like"Become a Super Admin". And as for the edit buttons I will implement them so that a super_admin can edit the status of the users to admin.(#1039)

aditya1702:
@mariobehling should I proceed with the above implementation

mariobehling:
@aditya1702 It would be good to have some kind of a mockup or screenshot to understand better, what you want to achieve. Users can not simply "become" super admins. It should be edited in `/admin/users/` in the option section. There is already an edit button. This button could open a simple pop-up for now with tickboxes and information on roles, e.g. see below please.

User Roles
* [ ] Verified User (has access to all user spaces and is granted roles in events according to event policies)
* [ ] Admin (has access to all functions of the system, but cannot change roles of Admins and Super Admins)
* [ ] Super Admin (has access to all functions of the system and admin area)

aditya1702:
@mariobehling I think we can close this. The Super_Admin will be already created when the database is created initially for our app. Thus the only part left is for thesuper_admin to create more admins which I have already implemented here: #1039

PR #1025, opened 2016-06-19
Display Admin link only when user is super_admin

aditya1702:
#1003


aditya1702:
@mariobehling please review

Issue #764, opened 2016-06-12
Prepare footer 


rafalkowalski:
@mariobehling what do you think? Currently it's an empty space 
![screen shot 2016-06-13 at 00 13 30](https://cloud.githubusercontent.com/assets/2479753/15994127/c5bc139c-30fb-11e6-8775-469c18376293.png)

mariobehling:
Yes, very good! So let's think, which links we have:
Column 1: 
* [ ] How it Works
* [ ] Pricing
* [ ] Mobile Apps
* [ ] Sitemap

Column 2:
* [ ] List of Event Categories (automatically generated by our top categories)

Column 3:
* [ ] List of Event Locations (automatically generated by our top locations)

Column 4:
* [ ] Support 
* [ ] Twitter
* [ ] Facebook
* [ ] Google+
* [ ] YouTube
* [ ] GitHub

Footer 1:
* [ ] About
* [ ] Blog
* [ ] Help (directs to Support)
* [ ] Press
* [ ] Security
* [ ] Developers
* [ ] Privacy
* [ ] Cookies

Footer 2:
* [ ] Dropdown menu with Interface Languages

rafalkowalski:
@mariobehling can you tell me what categories mean?

mariobehling:
Good point. Let's better separate categories into a) Types and b) Topics and add "Types" in the Column 2. Also compare with Eventbrite.

![screenshot from 2016-06-14 11 00 53](https://cloud.githubusercontent.com/assets/1583873/16037164/98e238c2-321f-11e6-8031-8fd578f8a1d9.png)

mariobehling:
Unfortunately the footer design is not like in the screenshot, e.g. text is too dark to read and width of bottom footer is different.

mariobehling:
Please check the responsiveness. There are a number of issues with the implementation in smaller resolution.

* [ ] The footer font is too dark and difficult to read
* [ ] At points the country bar overlaps
* [ ] in smaller resolutions are errors (Please compare to Eventbrite, grab the side of your browser and see how the footer of Eventbrite adapts to the size.

Compare medium small resolution
![screenshot from 2016-06-20 11 20 40](https://cloud.githubusercontent.com/assets/1583873/16189198/64f97e1c-36d9-11e6-9b6e-a48d97303fd3.png)

Compare small resolution
[...]

PR #1045, opened 2016-06-20
footer ui improved

rafalkowalski:
#764 

screen shot 2016-06-20 at 09 28 49


rafalkowalski:
@mariobehling can we close #764?

Issue #639, opened 2016-06-08
Implement Simple Signup for Not logged in users, who are signing up for events


mariobehling:
Think of a potential speaker, who answers the call for papers. Flow is as follows
1. Potential speaker receives a link to the call for papers, goes to the call for papers page and clicks on "submit session" button.
2. Without logging in the potential speaker (as anonymous user) should be able to 
- [x] fill out form "sessions" 
- [x] be prompted for email and password to sign up or sign in if he/she clicks "save draft"
- [x] Result: If speakers enters email address, he/she will receive a confirmation mail with the requirement to confirm the email. This is just like a sign up.

SaptakS:
@mariobehling I am working on it.

mariobehling:
There is a related issue and discussion about email and signup process here as well: https://github.com/fossasia/open-event-orga-server/issues/965

SaptakS:
@mariobehling , @niranjan94 said he would be working on it. If he doesn't, I will take it up.

SaptakS:
@mariobehling as said in #965 we will be having only a single form instead of a wizard. So I think this issue then only will require the last point. Right?

SaptakS:
@mariobehling This is the implementation that I have created: -
* The user is not logged in and creates a session. If he is not already registered, we register him, assign role as speaker and send an email with a randomly generated password
![screenshot from 2016-06-19 17 28 26](https://cloud.githubusercontent.com/assets/9530293/16177172/866b5584-3643-11e6-99fb-585a27f7c335.png)
* He can then verify is registration and login.

SaptakS:
Also, I am not sure what to do if the speaker is already registered but is not logged in while filling up form.

mariobehling:
@SaptakS Login questions are outlined here: https://github.com/fossasia/open-event-orga-server/issues/965 already.

SaptakS:
@mariobehling I think this [PR](https://github.com/fossasia/open-event-orga-server/pull/1041) already solves the purpose. Because as far as I understand, a non registered speaker would fill up session form only from `/e/ID/cfs` page. And to fill form from `/event/ID/sessions/create` ,  I think he needs to be logged in. What do you think @niranjan94 ?

mariobehling:
Sounds good. Please give me a sample set up and link to test it.

niranjan94:
http://open-event.herokuapp.com/e/3/cfs/

mariobehling:
[...]

PR #1047, opened 2016-06-20
Simple login for new session by unregistered speaker

SaptakS:
**[Not Ready to Merge]**
Steps done:

*    Speaker will be allowed to fill in the form without signing in. He/she presses save.
*    The session would be stored in an "unverified" state.
*    If the speaker is already registered, he/she is shown the message to login and verify the session.
*    If the speaker is not registered, a new account with a random password is created and an email verification is sent with the newly create password. Then, a message is shown "Email verification for account has been sent. Please login to complete the process."


SaptakS:
#639

SaptakS:
working on the build error.

PR #1054, opened 2016-06-20
Added profile image upload code, Shown image in session details page

SaptakS:
* Using S3 to save uploaded profile image and slides in.
* Display image profile picture in session details page..
session_details

### DIFF
--- a/open_event/helpers/data.py
+++ b/open_event/helpers/data.py
@@ -165,8 +165,7 @@ class DataManager(object):
         speaker_img = ""
         if speaker_img_file != "":
             speaker_img = upload(speaker_img_file,
-                                 'events/%d/sessions/%d/avatar' % (int(event_id), int(new_session.id)))
-        print speaker_img
+                                 'events/%d/speaker/%d/photo' % (int(event_id), int(new_speaker.id)))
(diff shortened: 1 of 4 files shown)


open_event/helpers/data.py
```diff
@@ -166,6 +162,13 @@ def add_session_to_event(form, event_id, speaker_img_name, state="pending"):
                               user=login.current_user if login and login.current_user.is_authenticated else None)
         new_session.speakers.append(new_speaker)
         save_to_db(new_session, "Session saved")
+        speaker_img = ""
+        if speaker_img_file != "":
+            speaker_img = upload(speaker_img_file,
+                                 'events/%d/sessions/%d/avatar' % (int(event_id), int(new_session.id)))
```
aviaryan: This is speaker's photo. So why not store in `events/id/speakers/id/photo` ? Is speaker not available ?
Also this method will fail if there are many speakers in a session.

SaptakS: True.. Should replace it by speaker id.. My mistake.

SaptakS:
@aviaryan @mariobehling please review.

PR #1068, opened 2016-06-20
Minify Flask's HTML output to speed up page load

niranjan94:
Currently the HTML Output generated by Flask has loads of whitespace and linebreaks which adds up to the page's size. 

This PR would minify the HTML Output of flask. (For, example, the home page's size was reduced by 40% after enabling this. Other pages' size reduction would vary)

(there was no issue created as this is very a small and a non-breaking enhancement)

@rafalkowalski @mariobehling please review and merge .

_We could go for a better cache-based implementation in future._

### DIFF
--- a/config.py
+++ b/config.py
@@ -16,6 +16,7 @@ class Config(object):
 
 class ProductionConfig(Config):
     DEBUG = False
+    MINIFY_PAGE = True
     try:  # you don't want production on default db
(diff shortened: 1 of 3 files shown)


mariobehling:
Good initiative and much needed, but I am missing an issue for this. Please always create issues for all work. It can be very brief if needed.

Issue #949, opened 2016-06-17
Admin: Sessions use admin style table


mariobehling:
In the user space we will adapt the design more and more to make it more suitable for normal users. In the admin area the table design is most suitable. 

Please implement the same table design as in the other sections of the admin area in the sessions section at `/admin/sessions/`.

Instead of "Upcoming/Past" use the following tabs

* [x] All Sessions
* [x] All Pending
* [x] All Accepted
* [x] All Rejected
* ~~Create Session~~

And implement the following table head columns:

* [x] Event
* [x] Title (of Session)
* [x] Speaker
* [x] Date of Submission
* [x] Date of Event
* [x] Organizer
* [x] Options

mariobehling:
I think "Create Session" - we need to take it out as it currently does not fit into the structure with different forms etc. So, let's keep it more simple for now.

mariobehling:
![screenshot from 2016-06-19 16 31 21](https://cloud.githubusercontent.com/assets/1583873/16177954/b9345dce-363b-11e6-8d52-ff9bab459122.png)

* [ ] Please in all views always move "Show xx Sessions" next to tabs
* [ ] Move "Search" on the left in tabs row
* [ ] Show 25 entries as standard

mariobehling:
Unfortunately dropdown menu does not seem to work anymore.
![screenshot from 2016-06-21 11 58 38](https://cloud.githubusercontent.com/assets/1583873/16225594/ae58e33a-37a7-11e6-9b5a-9f510109884e.png)

mariobehling:
Still not working in live system.

rafalkowalski:
@mariobehling it's working now

PR #1084, opened 2016-06-21
Show xx sessions and options added 

rafalkowalski:
#949 @mariobehling @niranjan94 
Displaying name changed
screen shot 2016-06-21 at 11 58 25 
Options added
screen shot 2016-06-21 at 11 59 11


mariobehling:
Still not working for me in live system. It worked for a moment and then again does not work.

PR #1092, opened 2016-06-21
Only show session times defined in the event

### LINKED ISSUE #976: Scheduler: Only show session times defined in the event
In the first step of the wizard the user adds a time for the event. Only show the available time in the scheduler, where the session can be entered.

Might be related to https://github.com/fossasia/open-event-orga-server/issues/953#issuecomment-226933080

niranjan94:
In the first step of the wizard the user adds a time for the event. Only the available time in shown in the scheduler. 

Resolves #976. 

@mariobehling @rafalkowalski @SaptakS please review and merge

### DIFF
--- a/open_event/static/js/admin/event/scheduler.js
+++ b/open_event/static/js/admin/event/scheduler.js
@@ -42,6 +42,24 @@ var time = {
     format: "YYYY-MM-DD HH:mm:ss"
 };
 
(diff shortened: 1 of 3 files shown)


mariobehling:
Please check in live system. Getting internal error to create an event.

niranjan94:
@mariobehling will check

PR #1103, opened 2016-06-21
Made social_links importable/exportable

### LINKED ISSUE #1099: Make social_links of an event importable/exportable
Export social links of an event in the event zip and then further import them from the same.

aviaryan:
Fixes #1099 

@shivamMg @mariobehling  please review

PS - I added a testcase to test successful import of OTS samples. It was taking around 60 secs even on TravisCI's server. Also it didn't add much variety to the test cases. So I commented it. I wonder if it was the right decision.


shivamMg:
All the process of creating the archive, deserializing json and then saving objects takes quite some time, and that too with around 240 kB of ots data. The PR looks good. Merging it :+1:

PR #1150, opened 2016-06-23
more activities logs added

rafalkowalski:
@aviaryan please review


aviaryan:
@rafalkowalski record_activity should be added in relevant API code too to capture logs of data update through API. I will merge it and add API logging and other logs that are needed. :smile:

mariobehling:
@aviaryan Please also take this issue into account: https://github.com/fossasia/open-event-orga-server/issues/1159 We should have all API information in one place and reference it instead of hardcoding it in every place where we need it.

Issue #1135, opened 2016-06-22
It's not possible to update profile details without upload new image


rafalkowalski:
Upload Image is required

PR #1155, opened 2016-06-23
Enable profile to be edited without image to be uploaded

aditya1702:
#1135


aditya1702:
@rafalkowalski @SaptakS Please review

PR #1162, opened 2016-06-23
Fix: ticket and booked_ticket using create_db.py

SaptakS:


SaptakS:
@niranjan94 @aditya1702 review

Issue #1035, opened 2016-06-19
Connect user profiles to Google and Facebook and get details


mariobehling:
If a user logs in using add details and profile photo from the accounts to Open Event.

aditya1702:
@mariobehling I updated the title of my PR  to Facebook also. Do we need anything else in this issue or can it be closed?

mariobehling:
This is a good first step, but we need to connect the apps more permanently. At  the moment it looks like people can change their info. You can compare how twitter adds apps in the settings. Let's follow up in milestone 8 and focus on bug hunting more right now.

PR #1169, opened 2016-06-23
Connect and display user's Google plus image and name

aditya1702:
#1035 
profile_img
profile_img2


aditya1702:
@mariobehling @SaptakS Please review

PR #1222, opened 2016-06-25
API validation for custom session and speaker forms

### LINKED ISSUE #989: Validate speaker and session API according to custom forms
Custom forms are used by the organizer to define which fields they want and require in their event.
The API should validate payload for sessions and speakers according to that custom form.

@SaptakS Can you please explain how form data is being stored in the customform model.

SaptakS:
Fixes #989 

@aviaryan @shivamMg please review.

### DIFF
--- a/open_event/api/sessions.py
+++ b/open_event/api/sessions.py
@@ -7,6 +7,7 @@ from open_event.models.microlocation import Microlocation as MicrolocationModel
 from open_event.models.speaker import Speaker as SpeakerModel
 from open_event.models.session_type import SessionType as SessionTypeModel
 from open_event.helpers.data import record_activity
+from open_event.helpers.data_getter import DataGetter
 
 from .helpers.helpers import get_paginated_list, requires_auth, \
     save_db_model, get_object_in_event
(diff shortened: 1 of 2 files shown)


open_event/api/sessions.py
```diff
@@ -146,6 +149,13 @@ def update(self, event_id, service_id, data):
         return obj
 
     def create(self, event_id, data, url):
+        session_custom = json.loads(DataGetter.get_custom_form_elements(event_id).first().session_form)
+        for key in SESSION_POST:
+            if key in session_custom:
+                if session_custom[key]['require'] == 1:
+                    SESSION_POST[key].required = True
+                elif session_custom[key]['require'] == 0:
```
aviaryan: setting required=False isn't required because all field in API models are required=False by default.

aviaryan:
@SaptakS It will be better if you make changes in `validate` method as it will affect both POST and PUT requests. The current version works only for POST requests.

open_event/api/sessions.py
SaptakS: But once we have converted to True and in next request if it is not required, then we need to turn it back to False, right?

aviaryan: @SaptakS you can clone the model in the function. Btw, this just works so forget what I said.

PR #1231, opened 2016-06-25
Fix import of event

aditya1702:
Earlier the function importEvent() was not being called. So I brought it in the same file


aditya1702:
@aviaryan Please review

Issue #1220, opened 2016-06-25
Admin: Implement Trash for deleted items


mariobehling:
Sometimes accounts get compromised and errors happen on the side of the user. It is always good to be able to get back deleted items. Therefore please implement "trash" for admins. When a used deletes an item, it will be deleted from the user account. However, the items is moved to trash of admins and gets deleted from trash after 30 days automatically. Please implement:

* [x] Events that are deleted -> move to trash and show up under a tab "All Deleted Events" in `admin/events/
* [x] Users that are deleted -> move to trash and show up in `/admin/users/` with status "deleted"
* [x] Sessions that are deleted -> move to trash and show up in `/admin/sessions/` under tab "All Deleted Sessions"
* [x] For all items, in options also show the delete option. Any item that is deleted from the trash is deleted forever.

PR #1268, opened 2016-06-27
Implemented Admin Trash

aditya1702:
#1220 
trash1
trash2
trash3

For the normal user the item deleted will not be shown in the tables. However it wont be deleted from the database.
trash4
trash5

Each item has a property in_trash which will be set to True as soon as the user deletes the item(session, event, user). And it will be shown in the trash of the admin.
trash6
trash7
trash8

However once it is deleted from the trash it will be permanently deleted from the database and a message will flash "Your item has been permanently deleted"
trash12
trash10


aditya1702:
@mariobehling I think this is how you wanted the top bar to look. The trash feature is working. Just the "30 days" time slot is to be implemented. I will make that commit as soon as possible :)

aditya1702:
@mariobehling Please review

mariobehling:
@aditya1702 The screenshot below "However once it is deleted from the trash it will be permanently deleted from the database and a message will flash "Your item has been permanently deleted" is the wrong location in the "user space". This message should appear in the admin space. Please check.

mariobehling:
Ok, let's test this on the system. Please also test this in detail on the live system and write tests. Thank you.

PR #1304, opened 2016-06-28
Add Custom-form-error to Speaker and Session APIs

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1223

@aviaryan Please review.

### DIFF
--- a/open_event/api/__init__.py
+++ b/open_event/api/__init__.py
@@ -19,6 +19,7 @@ from helpers.errors import (
     NotAuthorizedError,
     PermissionDeniedError,
     ValidationError,
+    CustomFormError,
     InvalidServiceError,
     ServerError,
 )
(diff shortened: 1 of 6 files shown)


aviaryan:
@shivamMg Is there any place Custom Forms Error will be returned ? If it won't be used, I don't think there is a need to define it. 
Also I think it is complex to decide whether to return a validation error or custom forms validation error on failure. So we won't be able to use it. What do you say ?

shivamMg:
@aviaryan Hehe, I created them because you requested for them. Anyways, there really doesn't seem to be a difference between validation error and custom form error, except that the latter should be used in session and speaker apis. So I don't know, if it isn't doing much we can just leave it.

aviaryan:
@shivamMg Sorry I realized it only after seeing this PR. Really sorry

Issue #1319, opened 2016-06-29
Improve top bar and start page


mariobehling:
Someone changed the size of the font on the top bar. It is way to small now. 

* [x] Please match the size.

![screenshot from 2016-06-29 14 50 55](https://cloud.githubusercontent.com/assets/1583873/16452467/f6b61f82-3e08-11e6-8cab-81cf44c6dcab.png)

The "Create Event" button on the start page also changed the looks over time. 

* [x] Please adjust is to match the looks as well.

![screenshot from 2016-06-29 14 52 48](https://cloud.githubusercontent.com/assets/1583873/16452506/2f271268-3e09-11e6-8f2e-8a92b403b6a7.png)

SaptakS:
@mariobehling on it. Plus we need to make it responsive as well. Someone didn't add the classes to make it responsive so it is not working correctly in small devices.. Will work on this.

SaptakS:
@mariobehling 

**medium screen resolution**
![screenshot from 2016-06-30 07 59 29 copy](https://cloud.githubusercontent.com/assets/9530293/16475161/f24a3cb2-3e99-11e6-9838-7650da145234.png)

**Small screen resolution**
![screenshot from 2016-06-30 07 59 51](https://cloud.githubusercontent.com/assets/9530293/16475163/f8296a54-3e99-11e6-90bb-32aae24e430d.png)

**Create Event Button**
![screenshot from 2016-06-30 07 59 29](https://cloud.githubusercontent.com/assets/9530293/16475177/0b4f513e-3e9a-11e6-8b47-ea3466ad9717.png)

SaptakS:
The measurements are exactly same as that in eventbrite.

mariobehling:
Please also fix the space between top bar and other areas. This is also in your screenshot. Not sure, why you do not see it.

![screenshot from 2016-07-01 02 29 57](https://cloud.githubusercontent.com/assets/1583873/16515369/ef58a088-3f71-11e6-87f7-c7c573840d0d.png)

SaptakS:
> Please also fix the space between top bar and other areas.

Didn't understand which space you are talking about.

mariobehling:
Thank you, good improvement. Still errors in some resolutions.

![screenshot from 2016-07-01 12 05 56](https://cloud.githubusercontent.com/assets/1583873/16518648/469b4df2-3f84-11e6-99ec-1b6b08580891.png)

SaptakS:
[...]

Issue #703, opened 2016-06-11
Implement "My Sessions" Dashboard


mariobehling:
Parent Issue: https://github.com/fossasia/open-event-orga-server/issues/622

The URL`/events/mysessions` - shows all sessions the user/speaker has submitted to all "his/her" events. Please implement as shown below.
![screenshot-www eventbrite com 2016-06-11 10-40-20](https://cloud.githubusercontent.com/assets/1583873/15984230/78e2d87c-2fc4-11e6-8391-e86e099512c7.png)

niranjan94:
I'll work on it right away .. :+1:

niranjan94:
@mariobehling this is the UI prototype
![capture](https://cloud.githubusercontent.com/assets/2404372/15985427/3525a9ac-300a-11e6-81ed-95225326bfb3.PNG)

mariobehling:
Looks good. Thanks! Please do not forget to add "Manage Sessions" on the top right dropdown menu.

mariobehling:
One thing here left: Please take out "admin" from the URL.

niranjan94:
@mariobehling so, you want admin to be removed from all URLs ?

mariobehling:
Yes, please. Details here as well https://github.com/fossasia/open-event-orga-server/issues/622

niranjan94:
Alright ... Will do

niranjan94:
Opened a separate issue for that here #724. Working on it now. Closing this issue.

PR #1336, opened 2016-06-30
Top bar

SaptakS:
@mariobehling please review #1319 

**medium screen resolution**
screenshot from 2016-06-30 07 59 29 copy

**Small screen resolution**
screenshot from 2016-06-30 07 59 51

**Create Event Button**
screenshot from 2016-06-30 07 59 29

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,5 +1,3 @@
 {
-    "directory": "open_event/static/admin/lib",
-    "proxy": "http://172.31.1.4:8080",
-    "https-proxy": "http://172.31.1.4:8080"
+    "directory": "open_event/static/admin/lib"
 }
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,3 +1,5 @@
 {
-    "directory": "open_event/static/admin/lib"
+    "directory": "open_event/static/admin/lib",
+    "proxy": "http://172.31.1.4:8080",
+    "https-proxy": "http://172.31.1.4:8080"
 }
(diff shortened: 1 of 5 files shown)


.bowerrc
```diff
@@ -1,3 +1,5 @@
 {
-    "directory": "open_event/static/admin/lib"
+    "directory": "open_event/static/admin/lib",
+    "proxy": "http://172.31.1.4:8080",
```
niranjan94: I think the proxy needs to be removed ?

open_event/helpers/data_getter.py
```diff
@@ -178,10 +178,10 @@ def get_sessions_of_user(upcoming_events=True):
         """
         if upcoming_events:
             return Session.query.filter(Session.speakers.any(Speaker.user_id == login.current_user.id)).filter(
-                Event.state != 'Completed')
+                Session.start_time >= datetime.datetime.now())
         else:
             return Session.query.filter(Session.speakers.any(Speaker.user_id == login.current_user.id)).filter(
-                Event.state == 'Completed')
```
niranjan94: The previous logic was correct. It is supposed to be sessions in upcoming **events** and past **events**. See the original issue #703 .

.bowerrc
SaptakS: My mistake.. sorry..

open_event/helpers/data_getter.py
SaptakS: But I think upcoming event means the one whose date is after the current date. And here state=completed shows the events which are live. So I think the new logic is correct. @mariobehling what do you think?

niranjan94: I meant, the date comparison has to be against event and not session.

SaptakS: But if session is in an event which is at a date later to current date, then wouldn't the session will also have a date later to current date?

PR #1343, opened 2016-06-30
Update Import functions

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1327

@aviaryan Please review. You can test with the OTS sample set. fossasia samples still need some work (https://github.com/fossasia/open-event-orga-server/issues/1187).


open_event/api/sessions.py
```diff
@@ -137,19 +137,31 @@ def fix_payload_post(self, event_id, data):
...
+        if st_data:
+            session_type = SessionTypeModel(
+                name=st_data['name'],
+                length=st_data['length'],
+                event_id=event_id
+            )
+            save_to_db(session_type, 'SessionType saved')
+            payload['session_type'] = session_type
+
         return ServiceDAO.create(self, event_id, payload, url, validate=False)
```
aviaryan: If you do it this way, then a session type will be created for each session. So if we have just 2 session types in the whole event and there are 100s of sessions, then 100s of duplicate session types will be created. This is not we want.

Also what about `session_type_id` in POST model. I see it is now not being used.
I suggest you leave the sessions API as it was and rather make change in the import_helpers to import session types from sessions json.

samples/ots16/sessions.json
```diff
@@ -151,7 +151,7 @@
             "name": "Erdgeschoss, Saal / Ground Floor"
         }, 
         "session_type": {
-            "id": 5, 
+            "id": 5, "length": 0.3, "length": 0.3,
```
aviaryan: double length , there is one more above

open_event/api/events.py
```diff
@@ -122,9 +130,23 @@ def create(self, data, url):
...
+            pass
+        data = BaseDAO.validate(self, data, model)
+        return data, social_links
+
     def update(self, event_id, data):
-        data = self.validate(data)
+        data, social_links = self.validate(data)
         payload = self.fix_payload(data)
+        for link in social_links:
+            save_to_db(link, 'Save social link')
```
aviaryan: This will create duplicate social links in the event. It will be better if we do it such that first all existing social links of an are deleted and then new social links are set.

aviaryan: Also social_links are not being linked to event (by setting event_id attr)

aviaryan:
I can work from here and fix these issues on my fork (and then send a PR). 
https://github.com/aviaryan/open-event-orga-server/tree/remove-sl-st

shivamMg:
@aviaryan Wait

shivamMg:
@aviaryan Please review.

open_event/api/events.py
```diff
@@ -122,9 +136,28 @@ def create(self, data, url):
...
+        return data, social_links
+
     def update(self, event_id, data):
-        data = self.validate(data)
+        data, social_links = self.validate(data)
         payload = self.fix_payload(data)
+        for link in social_links:
+            sl, created = get_or_create(
+                SocialLinkModel,
+                **link
```
aviaryan: I think there is a need to add event_id here (like 'create')

open_event/api/sessions.py
```diff
@@ -137,19 +137,33 @@ def fix_payload_post(self, event_id, data):
     def update(self, event_id, service_id, data):
         data = self.validate(data, event_id)
         data_copy = data.copy()
+        st = data_copy.get('session_type')
```
aviaryan: maybe field 'session_type_id' can be removed from POST model

shivamMg: I've removed it now.

open_event/api/sessions.py
```diff
@@ -137,19 +134,33 @@ def fix_payload_post(self, event_id, data):
...
         data_copy = self.fix_payload_post(event_id, data_copy)
         data = self._delete_fields(data)
         obj = ServiceDAO.update(self, event_id, service_id, data, validate=False)
         obj.track = data_copy['track']
         obj.microlocation = data_copy['microlocation']
         obj.speakers = data_copy['speakers']
-        obj.session_type = data_copy['session_type']
+        obj.session_type = st
```
aviaryan: st has a `dict` . I think we will have to create a SessionType object first (using get_or_create). 
Also it will be a good idea to update the sample test data (utils_post_data) since event and sessions POST/PUT API has changed.

open_event/api/events.py
```diff
@@ -122,9 +137,23 @@ def create(self, data, url):
...
+        del data['social_links']
+        data = BaseDAO.validate(self, data, model)
+        return data, social_links
+
     def update(self, event_id, data):
-        data = self.validate(data)
+        data, social_links = self.validate(data)
         payload = self.fix_payload(data)
+        for link in social_links:
+            sl = get_object_or_404(SocialLinkModel, link['id'])
```
aviaryan: Since social_links have been added to the PUT API, so it should be possible to add more social links to the event using a PUT request. But this PUT routine is not capable of that. I think it won't even allow deleting existing social links. 

I think it will be tough to properly include social links in PUT/POST API of Event (as it is a list). What about modifying the import functions to create social links using the event json. Same can be done for session types by using the sessions.json. Then we won't need to modify the API at all. (except adding length field in Session.SessionType GET API)

aviaryan:
@mariobehling @shivamMg 
Now that separate (sub-)APIs for social links and session types have been created, do we still have to remove social_links and session_types from sample json ?
Also this PR is obsolete now so it can be closed.

PR #1402, opened 2016-07-04
Create Social Links and Session Types API

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1390 [Can be merged]

Define SocialLink API for GetList, Post, Put and Delete methods. GET endpoint to fetch a single social-link does not have use case, so it seemed unnecessary to include it @aviaryan What do you think.


aviaryan:
> Define SocialLink API for GetList, Post, Put and Delete methods. GET endpoint to fetch a single social-link does not have use case, so it seemed unnecessary to include it @aviaryan What do you think.

Yes both Get Paginated and Get Item methods are unnecessary here so it's best we exclude them.

aviaryan:
Looks good. Merging

PR #1405, opened 2016-07-04
Sanitize output of rich text editors before saving to DB

### LINKED ISSUE #1388: Revision should not be created if no change
Revision is created even if you open an event for edit and then save it back without making a change. 
This should be ignored.
@niranjan94 @mariobehling What do you think ?

Screenshot: An empty revision
empty_revision

niranjan94:
Also resolves #1388. #1388 was due to the presence of a few additional whitespace characters in the output from the rich text editor. The sanitization fixes it. 

The sanitizer:

1. Allows only a small set of whitelisted tags (basic formatting options provided in the text editor)
1. Allows only a small set of whitelisted attributed (style attribute)
1. Allows only a small set of inline css styles (font weight, text decoration, text alignment)
1. Linkify's the text and also adds the `rel=nofollow` and `target=_blank` attributes to the link.

@mariobehling @rafalkowalski please review and merge/


niranjan94:
Will fix the failing tests.

Issue #1419, opened 2016-07-04
Save floor value in Microlocation


SaptakS:
On creating microlocation, floor values are not being saved

SaptakS:
Fixed.

Issue #1418, opened 2016-07-04
Creates new Session Type, Microlocation and Track on updating/editing


SaptakS:
After creation of event, on editing session type, microlocation and track information, a new object is created rather than updating the old one.

SaptakS:
Fixed.

Issue #1040, opened 2016-06-19
Sessions get deleted if the event is edited.


niranjan94:
In the current implementation of the event-edit functionality, all the tracks, session_types, microlocations are cleared and recreated again. 

This causes all the linked sessions to be removed as well.

shivamMg:
@niranjan94 The current implementation is supported by dynamic forms. Like for instance you can specify any number of tracks, rooms, session types, etc in the event create/edit form. During edit-event, the server removes the older entities and creates the new ones coming from the forms. The newer entities (when saved to db) are not the same as the old ones, even if the user does not edit them, but they do not lose their link with the event. The new entities created are linked to the same event. Can you give some instructions to reproduce the bug. The sessions don't seem to lose their link.

niranjan94:
1. Import the OTS Sample
1. Publish it
1. Goto /e/<event_id>/sessions - You'll see all the sessions with tracks
1. Now, goto /events/<event_id>/edit 
1. Don't have to change anything. Just press save. 
1. Goto /e/<event_id>/sessions - You'll see all the tracks, no sessions.
1. Goto /events/<event_id>/sessions - You'll see all the sessions, but the track field will be empty

niranjan94:
When the tracks are recreated, their `ID`s change. But the sessions are still having the old non-existant `track_id`s stored in them

shivamMg:
Point 4, `/events/<event_id>/edit` is giving a 500 for that event id. Attribute error for speaker form in events view. Is it working for you?

niranjan94:
What is the event ID ?

shivamMg:
The imported event has twelve.

niranjan94:
Check event 1

niranjan94:
http://open-event.herokuapp.com/e/1/sessions/ 
http://open-event.herokuapp.com/events/1/edit/

shivamMg:
Lol, can't login with my registered id, internal server error. Can't register a new one either, internal server error there too.

shivamMg:
Alright, it's working now, I mean the bug's reproducible. Working on it.

PR #1420, opened 2016-07-04
Bug fixes in Event Wizard Editing

### LINKED ISSUE #1416: Editing CFS in Event Wizard doesn't save information
If we edit the information for CFS in event wizard once it has been saved, it is not updating.

SaptakS:
Fixes #1416 , #1418 , #1419 

@mariobehling please review.

### DIFF
--- a/open_event/helpers/data.py
+++ b/open_event/helpers/data.py
@@ -993,6 +993,9 @@ class DataManager(object):
         custom_forms_value = form.getlist('custom_form[value]')
 
         # save the edited info to database
+        for session_type in session_types:
+            delete_from_db(session_type, "Session Type Deleted")
+
         for index, name in enumerate(session_type_names):
             if not string_empty(name):
                 session_type, c = get_or_create(SessionType,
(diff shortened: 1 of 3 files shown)


open_event/helpers/data.py
```diff
@@ -1009,6 +1015,9 @@ def edit_event(request, event_id, event, session_types, tracks, social_links, mi
                                                event_id=event.id)
                 db.session.add(social_link)
 
+        for track in tracks:
+            delete_from_db(track, "Tracks Deleted")
```
niranjan94: No. When you do this, after edit the tracks are recreated, their IDs change. But the sessions are still having the old non-existent track_ids stored in them. This same issue was there before (#1040) and was fixed by @shivamMg.

niranjan94:
Don't delete all the tracks, microlocation, session types etc, and recreate them during edit. It's not the right way. All the existing relations get messed up too. See https://github.com/fossasia/open-event-orga-server/pull/1077/commits/11803324a1561ff51d94c49c2131412bc81d564e .

SaptakS:
@niranjan94 true... Didn't think about it.. Will send a patch right away.. Thanks.

PR #1430, opened 2016-07-05
Fixed bugs and enhancements in Event API copyright field

### LINKED ISSUE #1429: Import fails if copyright is null or not set in Event API payload
This was reported here https://github.com/loklak/loklak_server/issues/642#issuecomment-230478572
Copyright was added to Event API here https://github.com/fossasia/open-event-orga-server/pull/1425
Also some code can be optimized using DAOs.

aviaryan:
Fixes #1429 

I removed copyright from tests.api.utils.create_event because copyright is an optional field and so we will find more bugs if we don't set copyright from the start. (like update bug [mentioned](https://github.com/fossasia/open-event-orga-server/issues/1429#issuecomment-230504135) in issue)

### DIFF
--- a/open_event/api/events.py
+++ b/open_event/api/events.py
@@ -9,8 +9,7 @@ from open_event.models.role import Role
 from open_event.models.user import ORGANIZER
 from open_event.helpers.data import save_to_db, update_version, record_activity
 
-from .helpers.helpers import get_paginated_list, requires_auth, parse_args, \
-    update_model, get_object_or_404
+from .helpers.helpers import get_paginated_list, requires_auth, parse_args
 from .helpers.utils import PAGINATED_MODEL, PaginatedResourceBase, \
     PAGE_PARAMS, POST_RESPONSES, PUT_RESPONSES, BaseDAO, ServiceDAO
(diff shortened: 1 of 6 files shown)


aviaryan:
@shivamMg please review

shivamMg:
@aviaryan Ok I understood it. For "copyright": null, the copyright attr in Event was being stored as None. While it should be stored as `<EventCopyright None>`, that is an EventCopyright object will all attributes as None. 

I made some changes in your branch so there is not need to check if event copyright exists during a PUT request (and then create or update it). I did not understand how to put all the so made a commit over your commits and made PR here: https://github.com/fossasia/open-event-orga-server/pull/1433. I've added some comments to explain the changes.

aviaryan:
@shivamMg 

>  I made some changes in your branch so there is not need to check if event copyright exists during a PUT request. I did not understand how to put all of it to describe it, so I made a PR here: #1433

You could have just said to make it so such that a copyright will always exist for an event after an event has been created. :smile: 

---

Also see https://github.com/fossasia/open-event-orga-server/pull/1433#issuecomment-230659983

PR #1433, opened 2016-07-05
copyright api fix

shivamMg:
Fixes provided by @aviaryan : https://github.com/fossasia/open-event-orga-server/pull/1430 with some changes mentioned [here](https://github.com/fossasia/open-event-orga-server/pull/1430#issuecomment-230558837).

One thing, I've updated the tests as well, so create_event binds a copyright to it. This is to reflect the apis. If you give a copyright:null in a post request it would create an empty `<EventCopyright None>` object and bind it to event (after the changes in this pr). But since we are directly creating an event in the database (not through apis) it should have an `<EventCopyright>` object, event if it's empty. 

You can test them first locally.

edit: One question, why do we have a validate param for DAO create and update. Aren't they already being validated.

### DIFF
--- a/open_event/api/events.py
+++ b/open_event/api/events.py
@@ -9,8 +9,7 @@ from open_event.models.role import Role
 from open_event.models.user import ORGANIZER
 from open_event.helpers.data import save_to_db, update_version, record_activity
 
-from .helpers.helpers import get_paginated_list, requires_auth, parse_args, \
-    update_model, get_object_or_404
+from .helpers.helpers import get_paginated_list, requires_auth, parse_args
 from .helpers.utils import PAGINATED_MODEL, PaginatedResourceBase, \
     PAGE_PARAMS, POST_RESPONSES, PUT_RESPONSES, BaseDAO, ServiceDAO
 from .helpers import custom_fields as fields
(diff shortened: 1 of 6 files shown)


aviaryan:
Some comments on the changes:

I think rather than changing `create_model` function which is used by the whole API, you can just change `payload['copyright'] = CopyrightDAO.create(payload.get('copyright'), validate=False)` to 
`payload['copyright'] = CopyrightDAO.create(payload.get('copyright', {}), validate=False)` . 

Also you have to make sure that copyright for event is created when an event is created through the UI. (since update of API won't work on event without copyright)

Also if a copyright field is not set in payload in update API, then it should not affect the current copyright of event. So a better way is - 
```
if payload.get('copyright'):
    CopyrightDAO.update(event.copyright.id, payload['copyright'])
```

aviaryan:
> One question, why do we have a validate param for DAO create and update. Aren't they already being validated.

The payload is automatically validated inside the create and update functions (see microlocations API). The validate param is to control whether to do validations or not. This is because there are times when more changes to payload/validate methods are required like sessions api, speakers api, event api. So it is to conrol that validate doesn't runs again (optimize time).

aviaryan:
> Also if a copyright field is not set in payload in update API, then it should not affect the current copyright of event. So a better way is -
```
if payload.get('copyright'):
    CopyrightDAO.update(event.copyright.id, payload['copyright'])
```

Sorry a more correct way is - 

```
if 'copyright' in payload:
    CopyrightDAO.update(event.copyright.id, payload['copyright'] if payload['copyright'] else {})
```

I guess a similar thing needs to be done in `create` to get empty dict when payload['copyright'] == null

shivamMg:
>I think rather than changing `create_model` function which is used by the whole API, you can just change `payload['copyright'] = CopyrightDAO.create(payload.get('copyright'), validate=False)` to 
`payload['copyright'] = CopyrightDAO.create(payload.get('copyright', {}), validate=False)` . 

This is definitely better :+1: 

> Also if a copyright field is not set in payload in update API, then it should not affect the current copyright of event. 

I've a little doubt about this. If the field is not set, doesn't it imply that the programmer wants it null instead.

aviaryan:
> I've a little doubt about it. If the field is not set, doesn't it imply that the programmer wants it null instead.

If programmer wants a field null, he can just set it null. Skipping a field in PUT should imply that he wants to not change that field. 

>  I know we don't handle it like that right now (if a field is not set during PUT, we don't update it). Just a question.

We do. If you have a look at how update_model function works, then it updates only the fields that are included in the payload. Although the payload validation will give a "Required field missing" error if a field is missing in PUT API. #1436 

I will merge this PR now.

Issue #1410, opened 2016-07-04
Add input validation in Wizard


aviaryan:
Right now sponsors url field accepts any input. (no URL validation)
Same goes for ticket url field in Step 1. 
I suggest we should add client side input validation. We are already including bootstrap validator in our dependencies.

In this issue, I will add input validation to wizard wherever possible.

aditya1702:
@aviaryan I could help with this issue

niranjan94:
@aviaryan @aditya1702 I think now that the URL validators have been added, all fields of the wizard are validated if and when required. This issue could be closed.

PR #1446, opened 2016-07-06
Added URL validators to Event wizard steps

aditya1702:
#1410

### DIFF
--- a/open_event/templates/gentelella/admin/event/wizard/step-1.html
+++ b/open_event/templates/gentelella/admin/event/wizard/step-1.html
@@ -5,8 +5,9 @@
                    value="{{ social_link.name | default('Twitter', true) }}">
         </div>
         <div class="col-sm-8 input-group">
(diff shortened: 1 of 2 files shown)


aditya1702:
@niranjan94 @rafalkowalski Please review

open_event/templates/gentelella/admin/event/wizard/step-1.html
```diff
@@ -182,8 +183,9 @@
         <div class="item form-group">
             <label>Ticket URL
             </label>
-            <input name="ticket_url" class="form-control"
-                   value="{{ event.ticket_url | default('', true) }}"/>
+            <input type="url" pattern="https?://.+" name="ticket_url" class="form-control"
```
aviaryan: rather than using a pattern, why not use type=url as you have done above (social link) .

aditya1702:
@aviaryan Its done for Event url so I also did for the ticket url

Issue #1456, opened 2016-07-07
Copy event does not copy Sponsors, Tracks, Rooms and Session Types


aditya1702:
If u create a copy of an event it only copies event details but does not copy the Sponsors, Tracks, Rooms and Session Types

PR #1463, opened 2016-07-07
Copied tracks, session types and rooms also during copying of an event

aditya1702:
#1456


aditya1702:
@SaptakS @rafalkowalski Please review

PR #1465, opened 2016-07-07
Form Validation in Event Wizard corrected

### LINKED ISSUE #1461: Make floor for rooms compulsory
Floors for rooms need to be compulsory. For now, it gives internal error. Will be adding form validation.

SaptakS:
Fixed #1461 #1462 

@mariobehling @rafalkowalski please review

### DIFF
--- a/open_event/static/js/admin/event/event_wizard.js
+++ b/open_event/static/js/admin/event/event_wizard.js
@@ -13,18 +13,22 @@ $(document).ready(function () {
     });
 
     $(document).on("click", ".save-event", function () {
-        $wizardForm.submit();
+        if(!validate()) {
+            $wizardForm.submit();
+        }
     });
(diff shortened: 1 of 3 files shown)


rafalkowalski:
@SaptakS do we need fllor field as required?

SaptakS:
Yes.. That is how the DB schema is right now.. Else it gives an internal error..

rafalkowalski:
@SaptakS @mariobehling i think floor field can't be required, we need to create event in short time, we don't need to wonder which floor we should put. So in database this field is wrong implemented

SaptakS:
@rafalkowalski For creating an event, floor field is not required.. It is required only when you create a microlocation.

rafalkowalski:
@SaptakS ok

PR #1472, opened 2016-07-08
Added functionality for after event mails

aditya1702:
#1362


aditya1702:
@mariobehling @SaptakS Please review

PR #1482, opened 2016-07-08
API endpoints added

SaptakS:
@championswimmer @mananwason added API endpoints `/events/ID/event/` to return the same values as `events/ID/`. Please review.

### DIFF
--- a/open_event/api/events.py
+++ b/open_event/api/events.py
@@ -257,7 +257,7 @@ class EventWebapp(Resource):
         return event
 
 
-@api.route('')
+@api.route('/')
 class EventList(Resource, EventResource):
     @api.doc('list_events', params=EVENT_PARAMS)
     @api.marshal_list_with(EVENT)
--- a/open_event/api/events.py
+++ b/open_event/api/events.py
@@ -227,6 +227,36 @@ class Event(Resource):
         return event
 
 
+@api.route('/<int:event_id>/event')
+@api.param('event_id')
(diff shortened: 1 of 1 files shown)


SaptakS:
Will get the build fixed ASAP.

shivamMg:
@SaptakS No need to create a new class again. You can just add a new path to the `@api.route` decorator.

```python
@api.route('/<int:event_id>', '/<int:event_id>/event')
```

But then why do we need it. It doubles the get, delete and put APIs for Events.

championswimmer:
@shivamMg 
When we used a file based API (like saving the jsons to a folder on github), then event/:id is a folder, so we save the event.json in event/:id/event.json

The API matching that would make it consistent. That's it

shivamMg:
@SaptakS You need to define only the `get` method for the class. As @championswimmer said, only the GET method is required. The reason we cannot just add a new path (like I said above) is because routes are defined for Resource classes and not individual methods.

```python
@api.route('/<int:event_id>/event')
@api.param('event_id')
@api.response(404, 'Event not found')
class EventWebapp(Resource):
    @api.doc('get_event_for_webapp') # should be unique
    @api.marshal_with(EVENT)
    def get(self, event_id):
        """Fetch an event given its id.
        Alternate endpoint for fetching an event.
        """
        return DAO.get(event_id)
```

SaptakS:
@championswimmer @shivamMg please review

PR #1494, opened 2016-07-09
Samples/update audio

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1454

@aviaryan Please review.

### DIFF
--- a/samples/ots16/sessions.json
+++ b/samples/ots16/sessions.json
@@ -1325,7 +1325,7 @@
         "video": null
     }, 
     {
-        "audio": "audio/5058.mp3", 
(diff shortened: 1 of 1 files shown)


samples/ots16/event.json
```diff
@@ -20,38 +20,6 @@
     "organizer_name": "FOSSASIA",
     "privacy": "public",
     "schedule_published_on": "2016-05-04 09:00:00",
-    "social_links": [
-        {
```
aviaryan: Why did you remove social_links ? Import JSON files should reflect the APIs and event GET API has social_links so I think we should keep it. 
Also the export of event will have social_links field in event.json (as they are based on APIs ) .

shivamMg: They seemed redundant as Social links were already in social_links.json, where they are actually created from. OK, I'll reset commit.

PR #1505, opened 2016-07-09
floors made non compulsory

SaptakS:
Enabled saving of event with microlocations having no floor information
#1461


SaptakS:
@niranjan94 @mariobehling review

SaptakS:
@mariobehling @aditya1702 is already writing the tests for the event wizards as a part of separate issue. So I think this PR can be merged.

Issue #1549, opened 2016-07-11
Normal user can see unpublished events


shivamMg:
A user can see details view of an unpublished event even if he is not member of staff or has permission to read event.

aditya1702:
@shivamMg Just decorate the view with @can_access I think

aditya1702:
I will do this if u dont mind :)

PR #1563, opened 2016-07-11
Prevent users from viewing details about other unpublished events

aditya1702:
#1549


aditya1702:
@shivamMg Please review

PR #1568, opened 2016-07-11
Merge migrations

shivamMg:
@rafalkowalski @niranjan94 We have multiple heads right now. here's the merged migration file.


shivamMg:
Wrong branch

PR #1574, opened 2016-07-11
Set in_trash as False during initialization

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1572


shivamMg:
Merging because it's urgent. It only fixes the bug.

Issue #1334, opened 2016-06-30
Update & format sample data from Loklak eventbrite.com scraper


jig08:
Currently, the `JSON` files extracted using the _Loklak eventbrite.com scraper_ in [github.com/fossasia/open-event-orga-server/samples/EventBriteInfo_LoklakScraper](https://github.com/fossasia/open-event-orga-server/tree/master/samples/EventBriteInfo_LoklakScraper) are outdated and not formatted as desired.

PR #1583, opened 2016-07-12
[PATCH] Updating & Formatting sample data from Loklak eventbrite.com scraper wrt #1334

niranjan94:
@rafalkowalski @mariobehling please review and merge


aviaryan:
@niranjan94 I think these samples won't work on import. So please add sessions.json, speakers.json etc with an empty array. ( `[ ]` )

niranjan94:
@aviaryan okay .. will do

Issue #1588, opened 2016-07-12
validate issue in Upload field


aviaryan:
Upload field fails to validate relative paths like `/images/logo.png` .
I noticed this issue when working on the import event zip with images feature #1449  . 
As I am already doing it, I will solve this also.

Related - #1454

Issue #1449, opened 2016-07-06
Import/Export needs to add all images (not embed them from external services)


mariobehling:
Currently we offer the import and export of images and include a link to the image. There are limits to this approach. If we import an image link to another website, the site might not allow us to show lots of images and block us. You can see this at event 55 in the system (http://open-event.herokuapp.com/e/55/).

Importing an event, should actually import all the images and import and export files should have all images in the actual image folder.

On the other hand we still want our Android app and Web app to be able to download updates from the website and we want to define a URL to do that.

How should we deal with this?

I think we could solve it this way: We add the images into the export/import file and also define API end points in one of the JSON files (maybe events.JSON) and hence the Android app will know where to get the files.

aviaryan:
Sounds good. We can create a folder named "images" in the zip. 
At the root it can have event-level data like logo.png, background.png etc. Then there can be folders inside it for sponsors and speakers. Those folders will have images related to sponsors and speakers respectively. 

> On the other hand we still want our Android app and Web app to be able to download updates from the website and we want to define a URL to do that.

I don't understand what here is meant by "updates" . Aren't the web and android apps supposed to use th e APIs to get data from server. So if something in an event is updated, its GET API data will be updated.

mariobehling:
> Aren't the web and android apps supposed to use the APIs to get data from server. So if something in an event is updated, its GET API data will be updated.

The Android app can use the API, but for the web app it is not sure. So, let's proceed that way and include the pictures in the zip itself. As other parts of the project (apps, and event parser) depend on this, it is very high priority. Please proceed asap.

aviaryan:
I will start working on this using Celery for background upload/download jobs. I plan to first do the import of images part and then the export part.

PR #1594, opened 2016-07-12
Upload media from import event zip and direct links to own server

### LINKED ISSUE #1592: Show progress info in Import Event dialog
Currently import event dialog only shows a "PENDING" message when event is being imported. This can be replaced with more informational messages that will give an idea on how much time the process is going to take.

aviaryan:
Fixes #1592 and #1588 
Fixes the import part of #1449 

upl

A sample import zip with both included media (stored in zip) and media with absolute urls has been attached below. It is a small event so it won't take much time to import. 
[EventBriteInfo_LoklakScraper.zip](https://github.com/fossasia/open-event-orga-server/files/359878/EventBriteInfo_LoklakScraper.zip)


aviaryan:
To include images, audio etc in zip, you first store the resource file in the zip and then refer it as the following if the logo.png is in the images folder inside the zip. The preceeding '/' is necessary. 

```json
{
"logo": "/images/logo.png"
}
```

PR #1610, opened 2016-07-13
Run Export Event as Celery background task

### LINKED ISSUE #1500: Run import/export events as background tasks using Celery (& Redis)
Parent issue: #1449 

Importing/Exporting of events can take a large amount of time. So it is not a good idea to block the application's main thread while importing/exporting an event.
We can use Celery for running these tasks in the background. Celery requires Redis which is available as REDISTOGO extension on Heroku. Locally redis can be setup from http://redis.io/download 

Celery http://www.celeryproject.org
Redis http://redis.io/

aviaryan:
Fixes #1500 (Import part is already done)

I'll do export event with media ( #1449 ) in the next PR.

export_event

### DIFF
--- a/open_event/api/exports.py
+++ b/open_event/api/exports.py
@@ -1,9 +1,11 @@
-from flask import send_file, make_response
+import os
+from flask import send_file, make_response, jsonify, url_for, current_app
 from flask.ext.restplus import Resource, Namespace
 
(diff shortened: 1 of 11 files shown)


aviaryan:
@shivamMg @niranjan94 please review

Issue #1642, opened 2016-07-14
Add local images to the OTS sample set


aviaryan:
Ability to import images and other media from the zip was added here https://github.com/fossasia/open-event-orga-server/pull/1594

The OTS sample set should include its images locally and the OTS json files should link to the local images.

mariobehling:
Please take the images from the OpenTechSummit https://github.com/OpenTechSummit/2016.opentechsummit.net/tree/gh-pages/programm

There are all here already and I hope we can do this quickly.

aviaryan:
@mariobehling Ok. will proceed.

mariobehling:
What about the JSON files that are not needed anymore? Which ones are integrated?

aviaryan:
If you are referring to https://github.com/fossasia/open-event-orga-server/issues/1327, then currently all json files are needed. After sub-apis were created for social links and session types, no decision was made on whether to remove the json files or not. See https://github.com/fossasia/open-event-orga-server/pull/1343#issuecomment-230527099
So currently all json files are integrated.

mariobehling:
@aviaryan Please delete un-needed JSON files, that offer duplicate data. Which ones are those? We are keeping samples here now https://github.com/fossasia/open-event

PR #1644, opened 2016-07-14
Added images to OTS sample set

aviaryan:
#1642 

The OTS.zip now is 
[ots16.zip](https://github.com/fossasia/open-event-orga-server/files/364707/ots16.zip)


mariobehling:
Ok, thanks. What about the audio files?

Issue #1555, opened 2016-07-11
Make first wizard step 1 into "one column" layout


mariobehling:
Please make the style more similar to Eventbrite into one column.

mariobehling:
Please make the design closer to Eventbrite. Create subissues if necessary.

Especially regarding:
* [ ] location and when a map shows up (This might require a subissue. The problem with calling Google Maps will be that as soon as we have a few users, we will have to pay. So better implement it in the way Eventbrite does).
* [ ] adding image
* [ ] make the same style as with background image below for the logo
* [ ] Add sections "Tickets" and move Ticket URL into this section

mariobehling:
Good progress. Please also
* [ ] Move column to center
* [ ] Move "next, save, make live" to bottom right hand of the actual column (not the page)
* [ ] Move event license to very last point in section "Additional Details" and rename to "License of Event Data and Content"

SaptakS:
Ok. Working on it.

PR #1645, opened 2016-07-14
Step-1 of Event Wizard Improved

SaptakS:
#1555 

fireshot capture 44 - create event - open event - http___127 0 0 1_5000_events_create_


mariobehling:
* [ ] Please move column to center
* [ ] Move "next, save, make live" to bottom right hand of the actual column (not the page)

PR #1676, opened 2016-07-16
Added a sample version.json file

mananwason:


mananwason:
Thanks for the merge @niranjan94 but for the future please let the checks to be completed first otherwise the travis builds can fail. Cheers !

niranjan94:
@mananwason right now, the sample json files are not being used in any of the test cases. :smile: That's why I merged without waiting for the build to complete :sweat_smile:

mananwason:
Okay np @niranjan94 😄

PR #1681, opened 2016-07-16
restructured directory to mock api structure

mananwason:


aviaryan:
@mananwason Can you elaborate the reason for this change ?
I think having this way will be confusing for a new-comer as these files serve as examples for both import/export feature and API.

mananwason:
@aviaryan Since we are trying to emulate the api via these json files, I added this directory to make it look like an endpoint.

PR #1685, opened 2016-07-17
Fossasia16 sample set [Under Discussion]

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1187

Include images as well. Updating json wrt the files.

@mariobehling @aviaryan @rafalkowalski Please review.


niranjan94:
@mariobehling should these samples along with so many media files be on version control repository ? So many media files also increases the repository download size. I think it's better to have the sample sets outside of the project repository.

niranjan94:
@shivamMg could you add these to https://github.com/fossasia/open-event along with the other samples there ?

Issue #1691, opened 2016-07-17
Can't delete events in trash


rafalkowalski:
Internal server error

PR #1706, opened 2016-07-18
Resolved error due to Event delete from trash

aditya1702:
#1691


aditya1702:
@SaptakS @rafalkowalski Please review

Issue #1674, opened 2016-07-16
Implement ticketing, payments, donations as modules


mariobehling:
As we are adding ticketing now, please make it optional.
* [x] Implement an admin page "modules" at `admin/modules`
* [x] On admin page give the option to activate and deactivate modules (only show ticketing if module is activated), implement following modules:
  * [x] ticketing
  * [x] payments
  * [x] donations

aditya1702:
@mariobehling Is this ok?
![1](https://cloud.githubusercontent.com/assets/15054664/16900671/efc41dc0-4c48-11e6-9cda-7b256ad337c4.png)
![2](https://cloud.githubusercontent.com/assets/15054664/16900672/efc68952-4c48-11e6-901b-ec65e48cb303.png)

mariobehling:
Well, I think this is a misunderstanding. Please check out "modules" pages at Drupal and "Plugins" pages at Wordpress. We need a page to activate/deactivate modules.

Apart from that we also need a page to show all activities of ticketing, but this is not part of this issue.

aditya1702:
@mariobehling I have implemented the following. If the module ticketing and payments is turned ON then we will show Free and Paid in the type menu
![2](https://cloud.githubusercontent.com/assets/15054664/17085673/befdb7c4-51fb-11e6-954f-6b4c58ab1fee.png)
![1](https://cloud.githubusercontent.com/assets/15054664/17085680/cb11a14c-51fb-11e6-87d3-0f050050ac23.png)
If donation is also enabled in the modules then we will also show Donation in the type box. Is this ok?

Presently I am working on the #1841 issue.

mariobehling:
@aditya1702 Hope you understand the entirety of a "payment module". This is not just if there could be the option of a paid ticket, but implementing the payment system including emails that get send out and tickets that are stored on user accounts. I recommend you have a look at eventbrite and eventnook and open relevant issues. 

Just to be sure: The module payments entails the whole payment process. Please outline how you plan to implement this.

aditya1702:
@mariobehling Yes I understand that the payment system will involve more. This is just a small implementation of the modules function. We will outline the system in more detail

PR #1746, opened 2016-07-19
Implement Admin Modules[Do Not Merge]

aditya1702:
#1674


aditya1702:
@mariobehling I have implemented it as you explained. The db part is left though. The template is shown above. If its good I will continue further :)

mariobehling:
Thanks!

mariobehling:
Please add the db part in a separate PR.

Issue #1758, opened 2016-07-20
Implement a different way to show progress in Import/Export jobs


aviaryan:
https://github.com/fossasia/open-event-orga-server/issues/1500#issuecomment-232501877

Import/Export jobs can take some time and user should be able to do other things in that scenario. 
As pointed by @mariobehling in #1500, 

> @aviaryan It is all a good idea, but I see that you put out a lightbox window while you generate the export and therefore the user has to wait until it is finished?
I do not think this is a good solution.
Please propose a better solution, e.g. we can have a progress bar in the event table somewhere and when it is finished there is a download link. In the meantime the user can also go to other pages and check later, if the download has finished.
Where is the best place to have this progress bar and download link in the table?

One solution I see is to send an email to user with the download link/event link when their export/import job have been completed. The lightbox (popup) can give the message to the user "Email will be sent to you when this is done". Then the user will have the choice whether to close the lightbox or not. If they close it, they can do other tasks and will still get notified through email.

mariobehling:
@aviaryan Close the lightbox on start export and show a progress bar. See how mailchimp does it.

mariobehling:
Step 1: After clicking what is exported, the "Export Page" can show an export in progress

![screenshot from 2016-07-20 09 00 33](https://cloud.githubusercontent.com/assets/1583873/16977806/a249a076-4e58-11e6-8551-b21f02ef3673.png)

Step 2: As soon as the export is finished and ready for download
* [ ] a) Send out an email
* [ ] b) Show export button on export page

![screenshot from 2016-07-20 09 01 18](https://cloud.githubusercontent.com/assets/1583873/16977830/cad9fa72-4e58-11e6-8841-ae55ca68ebc7.png)

(Compare to mailchimp)

mariobehling:
Here how the email looks like at mailchimp:
![screenshot from 2016-07-20 10 22 37](https://cloud.githubusercontent.com/assets/1583873/16979893/f0c66e18-4e63-11e6-9d61-ca43c743f99e.png)

Issue #1217, opened 2016-06-25
Import/Export: Add all files to Export


mariobehling:
Related to Import Issue: https://github.com/fossasia/open-event-orga-server/issues/1216

Currently we only export the JSON files. Please also implement the export of:
* [x] all image files
* [x] all audio files
* [ ] add log activity in /admin/logs

the-dagger:
Shouldn't be too hard @mariobehling. 
But we won't be needing audio files for the Android app. 
Also could you let me know what you mean by logs? 
Cheers.

mariobehling:
@the-dagger Ok, agree. Logs is related to the admin area. Please look at the online installation and check it out yourself at ` /admin/logs` (as mentioned in the issue).

aviaryan:
Since the need is that the exported zip can contain both direct links to resources on our server and embededded media files in the zip, I suggest we allow a config to be sent with an event export request. 

```js
{
"images": True,
"audio": False,
"video": False,
"slides": False
}
```

- From the above setting, the exported zip will have image links as relative and all other links (audio, video) as absolute. Likewise only images will be included in the zip. 
- If all the keys in the above setting are set to True, then everything will be included in the zip and all links will be relative. 

The android app/user can set the preferences according to their need and the zip will be generated accordingly. 
The Orga-server UI can show a form with checkboxes upon starting export of an event and so allow the user to select their need.

mariobehling:
> This is ofcourse believing that someone will not require both absolute links to a media object and the media object itself at the same time. 

@aviaryan Let's think of the use case, that an app would come with all the data, but would update later version, e.g. updated speaker images or updated/additional audio.

How should we handle this? Should we define the API endpoints as well?

aviaryan:
@mariobehling That is an interesting question. 
In other issue #1702, we have been discussing about versioning on the server so that the client apps can download only the data that is updated.
There we are describing versions of APIs like `speakers_ver`, `sessions_ver` and so. 
If the version num on server is more/different than one on the client, then fetch that API.
This works for the text data part but the problem is with images which need to be locally stored. 

How to know which images have been updated so that only they can be download ?
The best way for this is to check the change in URLs. But as the storage schema is currently set to use a fixed url for a fixed resource (e.g. event 1 speaker 42 photo = `/events/1/speakers/42/photo`) so the url will not change. 
[...]

PR #1759, opened 2016-07-20
Added Export zip with media feature

aviaryan:
#1217 and #1449 

export_options
export-options-done

PS - Regarding dismissing the lightbox and showing a progress bar in page etc., I don't have prior knowledge on how to do that so it might take some time. I have opened #1758 as a follow up issue for that. 

**Example** of an exported zip [event106.zip](https://github.com/fossasia/open-event-orga-server/files/372869/event106.zip)

### DIFF
--- a/app/api/exports.py
+++ b/app/api/exports.py
@@ -1,6 +1,6 @@
 import os
 from flask import send_file, make_response, jsonify, url_for, current_app
-from flask.ext.restplus import Resource, Namespace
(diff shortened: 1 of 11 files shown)


mariobehling:
@aviaryan We do not show the export button on the "manage events" page anymore. Please check my comments in https://github.com/fossasia/open-event-orga-server/issues/1758#issuecomment-233855809 where to add the export feature.

Issue #1663, opened 2016-07-15
Add a versions GET API in Event


aviaryan:
Parent #1702 

Versions GET API will basically have the last update date of an API.
Example - https://github.com/fossasia/open-event/blob/master/otsJson/event/1/version

This can be useful for the android app as they can decide whether to refresh data from the server or not. 

@shivamMg @mariobehling views ??

shivamMg:
@aviaryan :+1: Would be a nice feature.

aviaryan:
@mariobehling @shivamMg Do we still plan to add version information to Event API now that E-Tag has been done ? #1749 
Following the discussions at #1702, I think we have planned on keeping both (versions and E-Tag) but need to confirm so that I can start working on this issue. 
I plan to add `versions` information in the Event GET API as is being reflected by https://github.com/fossasia/open-event/pull/41 .

mariobehling:
Ok, on a note about our discussions: I am starting to loose track of the discussion across multiple issues. The best practices outlined was always to create smaller issues, that can be solved within a day and followed up easily. So, let's try to go back to this. We can have parent issues that help to understand the bigger picture. Please help to make it easier.

Regarding the question about E-Tag and Event-API, I think the version info in the event API is important when we create JSON files for import/export. So, yes we still need it.

In howfar do we need to document this? Please do the necessary.

PR #1778, opened 2016-07-21
Add version in Event API, fixed bugs

### LINKED ISSUE #1769: Multiple version data being stored for each update in database
For each update of an event (wheter in sessions or speakers or tracks etc), the old version data of event is being duplicated and stored with the column increment as a new field.
This should not happen and only one row with latest version information should be stored for an event.

This screenshot will help to understand what I mean to say - 
duplicate_versions

aviaryan:
Fixes #1769 and #1663 

Due to changes in database definition (foreign key constraint), db upgrade might give error. Run the following in sql shell to fix.
- The first statement removes all version data of events which have been deleted permanently. (#1663)
- The second statement removes all past version data for an event. (#1769)

```sql
DELETE FROM versions WHERE NOT EXISTS ( SELECT * FROM events AS T1 WHERE T1.id = versions.event_id );
delete from versions where id not in (select max(id) from versions group by event_id);
```

@niranjan94 @shivamMg please review


mariobehling:
@aviaryan I am merging this. Please check that the deployment on Heroku is working. Thank you.

aviaryan:
@mariobehling Done.

PR #1802, opened 2016-07-22
search query fixed

### LINKED ISSUE #1780: User don't see a query phase when puts input to search
main

http://127.0.0.1:8001/explore/united-states/events?query=blabla
explore open event

Query phase should be visible in search form

rafalkowalski:
#1780 @niranjan94 @SaptakS  please review

### DIFF
--- a/app/views/admin/home.py
+++ b/app/views/admin/home.py
@@ -181,7 +181,7 @@ class MyHomeView(AdminIndexView):
     def browse_view(self):
         params = request.args.items()
         params = dict((k, v) for k, v in params if v)
-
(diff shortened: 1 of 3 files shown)


app/views/admin/home.py
```diff
@@ -181,7 +181,7 @@ def forbidden_view(self):
     def browse_view(self):
         params = request.args.items()
         params = dict((k, v) for k, v in params if v)
-
+        print params
```
niranjan94: Please remove necessary print statements that were used while debugging before making a PR :)

Issue #1805, opened 2016-07-23
Accept button should not be shown if session already accepted


SaptakS:
Accept button should not be shown if session already accepted. Same goes with reject button.
![screenshot from 2016-07-23 14 42 16](https://cloud.githubusercontent.com/assets/9530293/17077180/4387f680-50e4-11e6-84ec-c7d2c3a2d3d8.png)

PR #1818, opened 2016-07-23
Fixed accept/reject button condition

SaptakS:
#1805 

screenshot from 2016-07-24 03 51 29

@niranjan94 @rafalkowalski review


mariobehling:
Good job thinking this through!

PR #1826, opened 2016-07-24
Added feature to empty trash every 30 days

aditya1702:
#1220


mariobehling:
Did you implement A) "Delete items from trash after 30 days automatically" or B) "Delete trash every 30 days"? We need A).

aditya1702:
@mariobehling I have implemented it in the following way: If an item for example an event is shifted to trash the date gets stored in the trash_date column. Then after 30 days from the date on which it got stored in the trash that event gets deleted. Same applies for the users and sessions.

mariobehling:
Ok, thank you.

PR #1828, opened 2016-07-24
Event wizard: sessions, speakers, call for speakers enhancements

### LINKED ISSUE #1740: Simplifying Wizard steps 3, 4, 5
Currently we offer options to turn on or off options 3 and 4 in the wizard. There are some dependencies that are difficult to handle and explain though, e.g. if sessions are switched on, but no speaker submission is possible (there is no button in the public event page) if call for speakers is switched off currently.

What is the solution for this? I see the only solution is to combine wizard points 3+4+5. 

If you have a better solution, please propose it. If not, please combine the three steps and implement the necessary changes in the dashboard check list.

### LINKED ISSUE #1811: Can there be events without tracks. Logically there is always one track
If we switch off tracks, what happens?

* Is there still a schedule?
* So, basically no tracks means there is just one track correct?
* But if there is one track, then should we keep this option "tracks" and take out the option to switch it off?
* What about concerts and other events that do not have other event parts?

I am unclear if we should have the option to switch on/off the event wizard option in step 3.

### LINKED ISSUE #1813: Wizard step 3 name: Track option disappeared in forms
The track option does not show up in forms anymore. We had a dedicated logic to show the track option or not. If we just keep tracks always, we could simplify this and always have the track option in the forms list. Described here https://github.com/fossasia/open-event-orga-server/issues/1811

If we keep the logic as is, then the track should sometimes show up and sometimes not.

At the moment it never shows up.

screenshot from 2016-07-23 20 26 41

niranjan94:
Resolves #1740
Resolves #1811
Resolves #1813

@mariobehling @rafalkowalski please review and merge


niranjan94:
Willl fix the tests.

mariobehling:
Build is failing and codecov/codacy not running. Please check.

niranjan94:
@mariobehling travis build and codacy has been fixed. 

As for the decrease in coverage, it is due to the fact that few lines of code are skipped when the session & speakers are turned off. I'll get the coverage back up in one of my next PRs.

PR #1829, opened 2016-07-24
Wizard/add ticketing sys

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1095
https://github.com/fossasia/open-event-orga-server/issues/1112

@aditya1702 Please review and merge.

I've added a "type" field to the ticket describing its type (free, paid, donation). I haven't added currency option for now (https://github.com/fossasia/open-event-orga-server/issues/1095). It shouldn't block issue on modules though (https://github.com/fossasia/open-event-orga-server/issues/1674). 

@aditya1702 Also, I commented out ticket settings in template to work on the above issues. The modules support can now be added.

Edit: Also, I haven't added support multiple tickets like in eventbrite, if it's required I'll add it later.

screenshot from 2016-07-24 17 41 26


mariobehling:
Codacy/Codecov tests not running. Please check.

app/models/ticket.py
```diff
@@ -25,16 +26,18 @@ class Ticket(db.Model):
 
     def __init__(self,
                  name,
-                 quantity,
-                 event,
+                 type,
```
shivamMg: This won't cause any issue. Builtin function "type" will not conflict with the "type" attribute here. This "type" field will be used by referencing it like `ticket_obj.type`. We are already using another "type" field in the Events model.

shivamMg:
Fixed Codacy issues.

PR #1886, opened 2016-07-26
Added query params in Sessions API to filter by time

### LINKED ISSUE #1878: Add api endpoint to get dates for the event.
@shivamMg @niranjan94 We need an endpoint to get all the dates of an event. Processing it while downloading sessions takes a massive hit on the performance of the app. We should do it on the server itself. 
When showing the schedule in the app, we need sessions sorted by dates, so I first tried processing dates and getting sessions by these dates, app was freezing due to the db operations. Then I tried with this dates endpoint, it seems to work.

But the best way would be to have an endpoint that gives sessions according to date as well or else something like https://github.com/fossasia/open-event/blob/master/testapi/event/1/eventDates should be okay as well.

Let me know what do you think @mariobehling @shivamMg @niranjan94

aviaryan:
Fixes #1878 (https://github.com/fossasia/open-event-orga-server/issues/1878#issuecomment-235225737)

### DIFF
--- a/app/api/helpers/query_filters.py
+++ b/app/api/helpers/query_filters.py
@@ -4,6 +4,7 @@ from sqlalchemy import or_, func, and_
 
 from app.helpers.helpers import get_date_range
 from app.models.event import Event
+from app.models.session import Session
 from custom_fields import DateTime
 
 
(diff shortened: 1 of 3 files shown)


aviaryan:
@mananwason @niranjan94 please review

PR #1900, opened 2016-07-27
Use HTTPS for socketio

shivamMg:


app/templates/gentelella/admin/base.html
```diff
@@ -78,7 +78,7 @@
         setATagsInNotifMenuDropDown();
         var namespace = '/notifs';
 
-        var socket = io.connect('http://' + document.domain + ':' + location.port + namespace, {reconnection: false});
+        var socket = io.connect('https://' + document.domain + ':' + location.port + namespace, {reconnection: false});
```
niranjan94: No. Don't do this either. It should use either https or http depending on what the page was loaded over. Because https might not necessarily be available on all installations

niranjan94: http://stackoverflow.com/a/17336519

Issue #1931, opened 2016-07-29
500 error if not ticket price is set


aditya1702:
Right now the ticket_price gets set to '' if its not given any value. But since ticket_price is an integer the string cannot be stored in an integer column. Hence we make the ticket_price option compulsory. That is if anyone chooses the paid option he/she will not be able to save the event without setting the ticket_price.

mariobehling:
What is the description? What is the proposed solution?

aditya1702:
@mariobehling Sorry about no description. I have given my views above

PR #1932, opened 2016-07-29
Set a default ticket price if no value is selected

aditya1702:
#1931


mariobehling:
Sorry, this is not peer-reviewed and there is no reason to have a standard ticket price. This has to be reverted therefore.

Issue #1050, opened 2016-06-20
Research and propose best practices for storing files


mariobehling:
Follow up to https://github.com/fossasia/open-event-orga-server/issues/1020

In the [documentation about S3,](https://github.com/fossasia/open-event-orga-server/blob/development/docs/AMAZON_S3.md#using-the-s3-storage-module-in-code) it is mentioned that a solution to keep data distinct is to use subfolders with IDs. Is this the final solution? Please research how other services store data, answer open questions and propose implementations of storage capabilities.

* [ ] What happens if files with the same data (or better e.g. same name like two speakers but same name) are downloaded? 
* [ ] What happens if we use the "copy" function to copy an event?
* [ ] What happens if we want to use the same file e.g. a logo in different events?

mariobehling:
@niranjan94 Could you please help to think this through and implement it according to best practices, e.g. any images, audio etc - all files should be stored on the server. We cannot depend on images that get called from other websites. This was already an issue for files that got imported and where added in wizard step 1. This is also a security risk and what if other sites are down? Please ensure a same-origin policy. This issue is the cause for a number of other open issues and questions and it is a very high priority by now.

@aviaryan I really wish you would have followed up on this in an earlier milestone.

niranjan94:
@mariobehling alright ... I'll research and give a proposal for implementation by tonight ...

niranjan94:
As of now, all the media that is to be added (logo, background image, slides, video, audio, sponsor logo, speaker image, etc) are being requested from the user as file and uploaded. We're not accepting file URLs as we cannot depend on external services. 

The only place where we are accepting URLs right now are in the API and import. For that we could:

1. Accept URLs. Verify if the file exists at the URL and then save else fail and throw an error to the user. We will download the file from the URL and upload it to S3 so that we don't have to depend on that URL anymore. (and/or) 
2. Let the user upload the file when he/she is making an API request and file will be saved in S3.

For import we could, 

1. Accept URLs. Verify if the file exists at the URL and then save else fail and throw an error to the user. We will download the file from the URL and upload it to S3 so that we don't have to depend on that URL anymore. (and/or) 
2. Have the user keep the files in the zip itself and give a path to the file in the URL field which we can use to save the file to S3. 

---

> What happens if files with the same data are downloaded?

I do not understand this. @mariobehling could you explain ? 

> What happens if we use the "copy" function to copy an event?

**Suggested.** the file must also be duplicated to reflect the new event ID. 
[...]

PR #1933, opened 2016-07-29
Implemented better storage schema

### LINKED ISSUE #1926: Implement a better storage schema
As a result of discussion at #1050, we found out that there were 2 issues with current implementation.

1. guessable urls - URLs right now follow a pattern and so are prone to scripts. 
2. low SEO - The filename of resource on server is something like `audio`, `video` etc. These don't describe the media well and thus have less SEO.

Solutions - 

1 - 
> https://github.com/fossasia/open-event-orga-server/issues/1050#issuecomment-231822943
What we can do is use the current method of setting a KEY for an object
eg. events/1/sessions/1/audio and then encrypt the KEY using app.secret_key as the key.
Then add the encrypted string to the KEY. 
So events/1/sessions/1/audio becomes events/1/sessions/1/audio/<ENCRYPTED> where ENCRYPTED is encrypted form of events/1/sessions/1/audio .
(ENCRYPTED can be base64 encoded to make sure it is url friendly)
As app.secret_key is supposed to be secret so this will be safe from scrapers.
Also it won't have the disadvantage of having problems with keeping only distinct data on the server. (as ENCRYPTED will be same for same KEY)

2 - 
Store the original filename of the media file after the url. eg > 
`events/1/sessions/1/audio/<ENCRYPTED>/<filename>` . As far as atomicity of data is concerned, there is a method in boto (s3 lib) to delete all files inside a key. So we can delete everything inside `events/1/sessions/1/audio/<ENCRYPTED>/` at once easily.

aviaryan:
Fixes #1926 ( and #1050 )

### DIFF
--- a/app/helpers/storage.py
+++ b/app/helpers/storage.py
@@ -1,8 +1,10 @@
 import os
-from shutil import copyfile
-
+from base64 import b64encode
+from shutil import copyfile, rmtree
+from flask.ext.scrypt import generate_password_hash
 from boto.s3.connection import S3Connection
 from boto.s3.key import Key
+from werkzeug.utils import secure_filename
 
 from app.settings import get_settings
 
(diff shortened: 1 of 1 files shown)


aviaryan:
@niranjan94 please review

PR #1955, opened 2016-07-30
Heroku One Click Deploy Patched

phoenix1796:
The previous Heroku deploy using the button did not work properly because it didn't list out all the steps like in doc/heroku.md 
I have added the following :
- Addons:
    - Postgres:hobby-dev
    - Redis:hobby-dev
- Buildpacks (In same order as mentioned in the docs)

(Above changes are in app.json)
This should fix the one click heroku button deploy

### DIFF
--- a/app.json
+++ b/app.json
@@ -25,10 +25,10 @@
   ],
   "buildpacks":[
       {
-          "url":"heroku/nodejs"
+          "url":"heroku/python"
       },
       {
-          "url":"heroku/python"
+          "url":"heroku/nodejs"
       }
     ],
   "scripts": {
--- a/app.json
+++ b/app.json
@@ -19,6 +19,18 @@
       "generator": "secret"
     }
   },
+  "addons":[
+      "heroku-postgresql:hobby-dev",
+      "heroku-redis:hobby-dev"
+  ],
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -19,6 +19,18 @@
...
+  "addons":[
+      "heroku-postgresql:hobby-dev",
+      "heroku-redis:hobby-dev"
+  ],
+  "buildpacks":[
+      {
+          "url":"heroku/nodejs"
+      },
+      {
+          "url":"heroku/python"
```
niranjan94: Could you swap the order of the build packs ? Python first and nodejs next. I'll update the docs.

phoenix1796:
@niranjan94 I have added the buildpacks in the same order as that written in docs/heroku.md
this is the correct order because the last buildpack decides the process types in the procfile

niranjan94:
@phoenix1796 yes. You're right about the last build pack determining the process type. But that's why we have a 'Procfile' in the project root to tell heroku what the process is. 

Right now both our heroku deployments have python first and nodejs next. So, could you swap the order?

phoenix1796:
@niranjan94 Done , Buildpacks swapped : )

PR #2020, opened 2016-08-02
Make SocketIO integration in Configs explicit

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1999

@niranjan94 Please review.


niranjan94:
Do the same check in the base view file and run the socket io javascript only if socket io is enabled.

shivamMg:
@niranjan94 Done!

PR #2032, opened 2016-08-03
Fix locale error on certain linux hosts, install redis at docker build

### LINKED ISSUE #2031: Locale error with Ubuntu 14.04 in Docker build
Docker build gave locale error when running on Ubuntu 14.04 x64 as the host. This didn't happen when I tested the build on Travis's server.

```
>> RUN localedef -v -c -i en_US -f UTF-8 en_US.UTF-8; 
usr/share/i18n/locales/en_US:7: non-symbolic character value should not be used
```

aviaryan:
Fixes #2031 

In addition to that, I also added the step of installing redis server in docker build. This is to avoid downloading and installling redis each time the container is run. (this will save time)


aviaryan:
@niranjan94 please review

Issue #1970, opened 2016-07-31
Offer different Payment Methods in Wizard Step 1


mariobehling:
Related to https://github.com/fossasia/open-event-orga-server/issues/1958

Different options include
* [x] Payment through Paypal
* [x] Payment through Stripe (please take care, that every organizer needs to set up a Stripe account. We cannot use a global Stripe account)
* [x] Offline payments

![screenshot from 2016-07-31 12 28 13](https://cloud.githubusercontent.com/assets/1583873/17275970/98890430-571a-11e6-8ef8-7b2abab568d4.png)

PR #2091, opened 2016-08-06
Add options for organizer to choose between payment methods

aditya1702:
#1970


aditya1702:
@niranjan94 Please review

mariobehling:
Please read the issue carefully. There is no requirement to choose - between - but organizers can give the option for different payment methods at the same time.

niranjan94:
@aditya1702 Also, please squash all your commits. Unnecessary merge commits.

Issue #2014, opened 2016-08-02
Event image or sample image not showing up in sessions photo


mariobehling:
Currently there is still a placeholder on the MySessions page.

![screenshot from 2016-08-02 20 23 57](https://cloud.githubusercontent.com/assets/1583873/17340124/344cd542-58ef-11e6-8849-d53e923e82f3.png)

mariobehling:
Thanks, this is a good start. Two issues with the implementation:
* [ ] Images are not reduced to thumbnail size. The image in the screenshot is a whopping 2.3MB and 1688x843px.
* [ ] The height of the images should be same as the height of the text area. If the width of the image area needs to be adapted for that, then ok.

![screenshot from 2016-08-05 08 16 48](https://cloud.githubusercontent.com/assets/1583873/17428078/e7340570-5ae5-11e6-967d-c87e101f0cd6.png)

SaptakS:
I will check this out.

SaptakS:
> The image in the screenshot is a whopping 2.3MB and 1688x843px.
Since the image uploaded is of that size, I don't think there is a way unless we save all the images in the backend after cropping and compressing which I am not sure about.

> If the width of the image area needs to be adapted for that, then ok.
In case of adapting the image according to the height, should I distort the image or should I rather let it be cropped?

mariobehling:
@SaptakS Please check how this is done in most CMS and just follow what they are doing. How are different sizes implemented in Wordpress? Please do research yourself.

SaptakS:
@mariobehling the size of the image remains same for all the pages. The thing that the CMS do is while uploading they provide a maximum and minimum image size and resolution. But the image size remains same in all pages.. It is just adjusted using css. Which I will do.

mariobehling:
@SaptakS This is not entirely correct. Wordpress, Drupal image modules and others for example have different options for Thumbnails etc.

mariobehling:
This is all related to https://github.com/fossasia/open-event-orga-server/issues/2019. Maybe implement this issue first.

SaptakS:
@mariobehling yes we can create at max 2 sizes - 1 for thumbnail and 1 for fuller image. But what I am trying to say is we need to do that while uploading the image itself. I will look into it.

mariobehling:
@SaptakS You can always keep the original and keep different sizes as well. And it is usually possible to generate images later as well, e.g. if we change the settings described in #2019

SaptakS:
[...]

PR #2093, opened 2016-08-07
Event image in Mysession List

SaptakS:
#2014

- [x] The height of the images should be same as the height of the text area. If the width of the image area needs to be adapted for that, then ok.
- [x] Default Placeholders resized to thumbnails

- [x] Transparent/white placeholder removed ( [Issue comment](https://github.com/fossasia/open-event-orga-server/issues/1845#issuecomment-237301067) )


niranjan94:
@SaptakS I think a fresh PR would be better ... The commits seem to have gotten messed up

SaptakS:
@niranjan94 let me check..

Issue #2016, opened 2016-08-02
Create a subtab in the admin content section for system images


mariobehling:
Follow up/Related to https://github.com/fossasia/open-event-orga-server/issues/1845#issuecomment-236981065

* [x] Create a subtab "System Images" at https://open-event-dev.herokuapp.com/admin/content/
* [x] List all topics and subtopics 
   * [x] Add a standard image for each topic/subtopic
   * [x] Enable administrator to change that image -> Create an upload field for an image
* [x] Add help text about which size (width/height) is required
* [ ] Add field form field for copyright and origin information next to each image as well

SaptakS:
@niranjan94 can you help me with the uploading the new image?

mariobehling:
@SaptakS Right now the implementation still shows images for topics, but subtopics are still not implemented. There should be much more different subtopics.

We discussed this 16 days ago already here: https://github.com/fossasia/open-event-orga-server/issues/1845#issuecomment-236990088

Please also add a dedicated photo for each subcategory. If subcategory is not chosen, use the photo of the main category.

![screenshot from 2016-08-02 20 04 32](https://cloud.githubusercontent.com/assets/1583873/17339498/79a69c20-58ec-11e6-9af3-5c35616f858b.png)

mariobehling:
There are also a number of pages with images of other sizes missing and while all pages in the admin section have a green save button on the bottom right, it is different here.

I think I mentioned it before: Please look how other systems are doing it, e.g. Wordpress and Drupal.

There are different "category" sizes of images, e.g. Thumbnail, Medium Size, Large Size. If more different sizes are needed e.g. in Drupal you can define your specific custom size. A size that I see we need additionally is for example "Icon".

The strategy would be  to list which pages use "Thumbnails" and then make the thumbnail size configurable.

I see you are missing pages like event pages as well and config pages that have very small icons.

Looking at how you got things doing it seems you did not have investigate how others are doing things. In Wordpress you would see the following though, which would already get us moving forward.

![screenshot from 2016-08-18 17 19 00](https://cloud.githubusercontent.com/assets/1583873/17779545/546ffb4e-6568-11e6-86e8-c552dfe60c56.png)

mariobehling:
Ok, my last comment should have gone into this issue https://github.com/fossasia/open-event-orga-server/issues/2019#issuecomment-240758846

I deleted it and reposted it there. Thanks.

SaptakS:
[...]

PR #2105, opened 2016-08-07
System Image Tab added in admin/content/

SaptakS:
#2016 

screenshot from 2016-08-08 01 24 09

Back end code to yet to be implemented


mariobehling:
Thank you. Please also list the subtopics and include specific images for subtopics.

PR #2111, opened 2016-08-08
Tickets: Hide feature. Lock Price and Ticket Name.

shivamMg:
Fixes: https://github.com/fossasia/open-event-orga-server/issues/2110 and https://github.com/fossasia/open-event-orga-server/issues/2044#issuecomment-237481910

screenshot from 2016-08-09 09 31 56

screenshot from 2016-08-09 09 33 12

The Price and Name of the Ticket are locked (made read-only) if the Ticket has already placed orders.

@mariobehling @niranjan94 Please review and merge.


mariobehling:
@shivamMg "Hide Ticket" should not stop ticket sales. If the organizer wants to stop ticket sales he/she already has the option of changing the dates until tickets can be sold. "Hidden Tickets" could still be sold through link sharing, which is another feature that needs to be implemented. 

For now, please ensure "Hidden Tickets" could still be sold, but they are not shown publicly.

niranjan94:
> "Hide Ticket" should not stop ticket sales. If the organizer wants to stop ticket sales he/she already has the option of changing the dates until tickets can be sold. "Hidden Tickets" could still be sold through link sharing, which is another feature that needs to be implemented. For now, please ensure "Hidden Tickets" could still be sold, but they are not shown publicly.

@shivamMg is this ready for merging ?

shivamMg:
@niranjan94 Yes it is.

Issue #2023, opened 2016-08-02
Implement control of service fees for ticket sales


mariobehling:
Implement control of fee settings as a subtab in `https://open-event-dev.herokuapp.com/admin/settings/` as follows:
* [x] List each supported currency
* [x] Below currency, implement a form field for percentage fee, e.g. 2.50% per ticket
* [x] Below percentage field implement a form field for a maximum fee per ticket, e.g. maximum 19.90 USD per ticket
* [x] Only show this tab if ticketing and payments module are enabled

See a sample how Eventbrite charges fees: 
https://www.eventbrite.com/support/articles/en_US/Troubleshooting/how-much-does-it-cost-for-organizers-to-use-eventbrite?lg=en_US

aditya1702:
@niranjan94 Already working on 2nd point of #1920 , so would like to help in this

aditya1702:
@mariobehling Is this proper?
![1](https://cloud.githubusercontent.com/assets/15054664/17464574/aae90702-5cff-11e6-819a-3767764d691d.png)
![2](https://cloud.githubusercontent.com/assets/15054664/17464575/aafa805e-5cff-11e6-8f87-e096e5cbf8f3.png)

mariobehling:
It looks good, but I cannot see that this is using percentage and currency a bit. Could you reduce the width of the fields and add behind the first field "%" and behind the second field the currency e.g. USD, EUR, R$ etc.?

mariobehling:
* [ ] Currently, changes in ticket fees are not saved, e.g. AUS Dollar cannot be changed.
* [ ] When data is saved, the page redirects to first tab instead of ticket fees tab itself
* [ ] What happens when maximum fee is 0? Expected behavior: There should not be a maximum fee in that case meaning: Whatever is the ticket price, the fee is always the percentage listed.

aditya1702:
@mariobehling Changes in the fees are being saved. How did you get the bug in the first point?

mariobehling:
![screenshot from 2016-08-13 11 37 45](https://cloud.githubusercontent.com/assets/1583873/17642525/7877a612-614a-11e6-9c41-cae0e9ee1cee.png)

Changing it to 50 AUS does not change it and after save it is back.

aditya1702:
@mariobehling Yes I have found the issue. Will make a PR asap

aditya1702:
[...]

PR #2122, opened 2016-08-08
Implement control of fees in ticketing system

aditya1702:
#2023


aditya1702:
@mariobehling @niranjan94 Please review

mariobehling:
@niranjan94 please review.

niranjan94:
Looks good. The UI could be improved, but that can be taken care of later as it's not a priority at this point. we need the functionality right away .

PR #2127, opened 2016-08-09
Unpredictable non-increment-able URLs for public event pages

### LINKED ISSUE #1850: Propose and implement a better event url handling
Currently we give events a number that is incrementing and we use this number in the URL. The problem with this is, that we release the info about how many events are on the system and it is easy to guess URLs.

Please propose and implement a better URL handling of events.

niranjan94:
Resolves #1850. 

**Example:**
`http://localhost:5000/e/e6905e43/`


niranjan94:
@mariobehling @rafalkowalski @SaptakS please review and merge

PR #2158, opened 2016-08-11
Display PayPal and/or stripe depending on the currency chosen and availability

### LINKED ISSUE #2141: What are the currencies that we're going to support ?
capture

Since we're providing two payment Gateways, we'll have to make sure all the currencies that we provide are supported in both the payment gateways. 

The Forex column represents the supported currencies in various Forex APIs. (as we'll need to convert to display the sales on an event-wise basis and for charging the fee)

@mariobehling please share your thoughts regarding this.

niranjan94:
Resolves #2141 ([See comment](https://github.com/fossasia/open-event-orga-server/issues/2141#issuecomment-238960373)).

@mariobehling @rafalkowalski please review and merge


mariobehling:
Travis failing. Please check.

PR #2161, opened 2016-08-11
Display Ticket Fees tab only when ticketing and payments are enabled

aditya1702:
#2023


aditya1702:
@niranjan94 @mariobehling Please review

Issue #2132, opened 2016-08-09
Add possibility to catch callback notification when app is generated


rafalkowalski:
If Web app or Android app are generated user should receive notification.

PR #2163, opened 2016-08-11
Possibility to create user Notification by api added

rafalkowalski:
#2132  @niranjan94 @SaptakS please review

### DIFF
--- a/app/api/notifications.py
+++ b/app/api/notifications.py
@@ -16,7 +16,7 @@ api = Namespace('notifications', description='Notifications', path='/')
 
 NOTIFICATION = api.model('Notification', {
     'id': fields.Integer(required=True),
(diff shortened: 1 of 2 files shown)


app/api/notifications.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+# Create DAO
+class NotificationDAO(ServiceDAO):
+    version_key = 'notifications_ver'
+
+    def create_user_notify(self, payload):
+        user = DataGetter.get_user_by_email(payload['email'])
+        DataManager().create_user_notification(user, payload['action'], payload['title'], payload['message'])
+        return user
```
aviaryan: you are returning user here whereas API response is set to NOTIFICATION. So response won't be returned correctly.

Issue #2175, opened 2016-08-12
Don't hardcode server name in app


aviaryan:
Open Event should be able to run anywhere. So we should avoid hard-coding the domain name in the code wherever possible. Right now I see 2 places where domain has been hardcoded.

- [x] app.config['SERVER_NAME']  - This value has been commented because of the same reason. #1852 

- [x] get_latest_heroku_release function - App name has been hardcoded to open-event.. We should use HEROKU_APP_NAME environment variable there. 

There may be more places I might have missed.

PR #2176, opened 2016-08-12
Don't harcode app name of heroku

### LINKED ISSUE #1852: SERVER_NAME in app.config should be generic
Right now the `SERVER_NAME` in app.config is set to "open-event-dev.hrokuapp.com" because of which if the code is hosted in any other domain, it is unable to recognize the URL and gives 404.

I think we can set the value using environment variable.

aviaryan:
Fixes #1852 and #2175

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -332,6 +332,16 @@ def update_sent_state(sender=None, body=None, **kwargs):
 def set_secret():
     current_app.secret_key = get_settings()['secret']
 
+
+@app.before_first_request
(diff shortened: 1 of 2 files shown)


mariobehling:
Please resolve branch conflicts.

niranjan94:
@aviaryan looks like the server name has been set in https://github.com/fossasia/open-event-orga-server/pull/2184

aviaryan:
@niranjan94 Ok. I have removed that part.

PR #2184, opened 2016-08-12
Allow SERVER_NAME to be set dynamically

aditya1702:
#1852


aditya1702:
@niranjan94 @SaptakS Please review

Issue #2179, opened 2016-08-12
Ensure social links are taken care of in migrations


mariobehling:
Currently social links often disappear after updates. Please ensure they are part of migrations.

mariobehling:
Social links still getting lost after updates.

mariobehling:
![screenshot from 2016-08-24 08 08 10](https://cloud.githubusercontent.com/assets/1583873/17920054/4bbcc8a6-69d2-11e6-8237-778863291a86.png)

PR #2198, opened 2016-08-13
Social links fixed in settings model

rafalkowalski:
#2179 @niranjan94 please review

### DIFF
--- a/app/settings/__init__.py
+++ b/app/settings/__init__.py
@@ -39,12 +39,10 @@ def set_settings(**kwargs):
     else:
         setting = Setting.query.first()
         if not setting:
-            print "adsadsa"
             setting = Setting(**kwargs)
         else:
-            "dsadsadsa"
-            for k, v in kwargs.iteritems():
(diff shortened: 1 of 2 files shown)


app/settings/__init__.py
```diff
@@ -37,7 +37,14 @@ def set_settings(**kwargs):
                                     maximum_fee=ticket_maximum_fees[i])
             save_to_db(ticket_fee, "Ticket Fees settings saved")
     else:
-        setting = Setting(**kwargs)
+        setting = Setting.query.first()
+        if not setting:
+            print "adsadsa"
```
niranjan94: Could you remove this print statement ?

app/settings/__init__.py
```diff
@@ -37,7 +37,14 @@ def set_settings(**kwargs):
...
+            setting = Setting(**kwargs)
+        else:
+            "dsadsadsa"
```
niranjan94: Could you remove this string ?

Issue #2202, opened 2016-08-13
Event details page not found


rafalkowalski:
For example http://open-event-dev.herokuapp.com/e/122/ returns 404 code

PR #2203, opened 2016-08-13
Searching page fixed

rafalkowalski:
I've changed id to identifier event
#2202 @niranjan94 please review


SaptakS:
@rafalkowalski please fix the travis build.

tests/api/utils_post_data.py
```diff
@@ -40,7 +40,8 @@
         "privacy": "public",
         "start_date": "2016-06-29T19:30:00",
         "timezone": "UTC"
-    }
+    },
+    "identifier": "23c0644e"
```
aviaryan: @rafalkowalski I don't think identifier will be a part of Event POST API input payload as it is machine set.

PR #2219, opened 2016-08-14
Maximum fee should be equal to service fee if not set by user

aditya1702:
#2023


aditya1702:
@niranjan94 Please review

PR #2241, opened 2016-08-16
Admin: User Role Permissions

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/1990

screenshot from 2016-08-16 20 56 08

@aviaryan Please review and merge.


mariobehling:
Why are you following a different design on each tab? Please make it consistent. 

Look at the Drupal permission table and follow that please. We can use this as an example.

Issue #2171, opened 2016-08-12
Simplify Admin Sections and Description and enhance Admin Dashboard


mariobehling:
Let's keep it more simple, please:
* [x] In Admin section rename tab heading as follows: 
  * Manage All Users -> Users; 
  * Manage All Permissions -> Permissions; 
  * See System Messages -> System Messages; 
  * Manage All Reports -> Reports;
  * Manage All Settings -> Settings; 
  * Manage All Modules -> Modules; 
* [x] Move Logs to become a subtab in "Report" and implement as follows
  * [x] Subtab a) System Logs
  * [x] Subtab b) Server Logs (currently Heroku) 
* [x] In admin dashboard in `/admin/` take out "Recent Events"
* [x] In admin dashboard implement the following widgets
  * [x] Add Widget "Events" with the information: Number of Events a) Live, b) Draft, c) Past
  * [x] Add Widget "Users" with information: a) System Roles: Number of Super Admins, Admins and Registered Users and b) Event Roles: Number of Organizers, Co-Organizers, Track Organizers, Attendees of all events
  * [x] Add Widget "Sessions" with number of Accepted and Draft Sessions
  * [x] Add Widget "Messages" with information "Number of sent out emails in the last 24 hours, last 3 days, last 7 days, last month and number of all emails sent out since installation.

mariobehling:
The admin dashboard is leaving white spaces between sections. Please implement without white spaces. Compare to event dashboard for ideas.

![screenshot from 2016-08-16 10 43 08](https://cloud.githubusercontent.com/assets/1583873/17693292/676f891a-639e-11e6-91f6-93cd7e1cccb1.png)

mariobehling:
* [ ] Number of event organizers and other event roles is 0. It should not be.

![screenshot from 2016-08-16 10 46 01](https://cloud.githubusercontent.com/assets/1583873/17693351/b0b5ae6a-639e-11e6-865a-516c6a383232.png)

rafalkowalski:
@aditya1702 I will do this

mariobehling:
Thank you for fixing this.

Currently the panel shows 108 organizers and 11 co-organizers. The event section shows - including live, draft, past events - 99 events.

Please explain how the numbers are calculated.

![screenshot from 2016-08-17 09 29 48](https://cloud.githubusercontent.com/assets/1583873/17728060/7cc376d6-645d-11e6-8927-f38a5ff06799.png)

[...]

PR #2243, opened 2016-08-16
Performance improved in admin panel 

rafalkowalski:
#2171 @niranjan94 please review


shivamMg:
Nice. Also takes care of redundant code.

PR #2244, opened 2016-08-16
Add info tooltip to prevent maximum fee from being set as 0

aditya1702:
#2023


aditya1702:
@niranjan94 @mariobehling Please review

PR #2245, opened 2016-08-16
Flow of organizers to make monthly payments of service fees

### LINKED ISSUE #2025: Propose and implement flow of organizers to make monthly payments of service fees
Let's assume our ticketing has a monthly payment fee of 2.5%. 
* [x] Then we need to sent out an email on the first day of each month with the percentage of the income generated by the organizer with a two week period for the organizer to pay. 
* [x] If the money has not been received by the 15th of the month, we need to send out another email as a reminder.
* [x] For invoicing we should use the payment gateways of paypal and stripe. The gateway/API info should be entered in subtab "Payment Gateways" in `admin/settings`

Please propose and implement a flow to make this happen.

niranjan94:
Resolves #2025.


rafalkowalski:
@niranjan94 please fix tests

PR #2248, opened 2016-08-16
Facebook Open Graph Tags

shivamMg:
https://github.com/fossasia/open-event-orga-server/issues/2145

@niranjan94 Please review and merge.


app/__init__.py
```diff
@@ -96,6 +97,9 @@ def create_app():
     app.config['SQLALCHEMY_RECORD_QUERIES'] = True
     # app.config['SERVER_NAME'] = 'http://127.0.0.1:8001'
     # app.config['SERVER_NAME'] = 'open-event-dev.herokuapp.com'
+
+    # Facebook APP ID
+    app.config['FB_APP_ID'] = os.getenv('FB_APP_ID', '')
```
niranjan94: No. Don't set it here. Allow the admin to change it at `/admin/settings/`.

shivamMg: Turns out it's App ID is same as Facebook Client ID in Settings model.

niranjan94:
@shivamMg also add the same Facebook App ID field along with the social links on the super admin settings panel.

mariobehling:
In the light, that we still have a lot of bugs and basics not working I do not understand why we need to spend time on this kind of social media issue.

shivamMg:
will submit another PR.

PR #2259, opened 2016-08-17
Fixed docker deployment

### LINKED ISSUE #2121: Make sure open-event-server runs with Docker
I fixed Docker build issue in #1997 . Build works fine now but `docker run` i.e. running the app still has problems (The server won't run). Today when going through docker docs and testing it on another project, I realized that our Dockerfile is highly outdated and very inappropriate compared to current standards.

For example, we should be using [docker-compose](https://docs.docker.com/compose/) as orga-server app is complex and comprised of many components.

* App [Python, NodeJS]
* Celery workers
* Redis
* Postgres Database

The Dockerfile should be simplified too and it should not install everything in one container. i.e. for example postgres should not be installed in `App` container. 
We should use `volumes` to make sure that data (postgres database) remains preserved across builds. 
I am sure more improvements can be done too.

---

This is a parent issue and will be closed only when orga-server is perfectly running on Docker.

aviaryan:
Fixes #2121 

The server can now be run through Docker. Redis, Celery, Postgres etc everything is working as expected.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,44 @@
+.git
+.dockerignore
+
+#####################
+# .gitignore copied
+#####################
+
+.vagrant/
+.DS_Store
+*.pyc
+*.pyo
+env
+env*
+dist
+build
+*.egg
+*.egg-info
+_mailinglist
+.tox
+.cache/
+test.db
+app.db
+.idea/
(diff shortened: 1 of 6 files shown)


mariobehling:
Good job!

PR #2264, opened 2016-08-17
Admin Permissions

shivamMg:
Resolves: https://github.com/fossasia/open-event-orga-server/issues/2263 and https://github.com/fossasia/open-event-orga-server/issues/1990#issuecomment-240421953

@niranjan94 Please review.


mariobehling:
@shivamMg User Roles are System Roles. I do not understand why there are different tabs now. It would be great if you could be more transparent about what you are doing and follow the logic of the existing event roles tab and design the table in the same way.

shivamMg:
@mariobehling Sorry, I actually thought [this](https://github.com/fossasia/open-event-orga-server/issues/1990#issuecomment-240422593) meant applying the System Roles and Event Roles logic to User permissions by creating a separate tab for it. I'll submit another PR with the fix.

Issue #2273, opened 2016-08-18
Implement a way to resend confirmation email


obnoxious23:
After creating a new user I did not receive a confirmation email. Thus I can login, but cannot i.e. publish events. Besides, I get this notification while editing events:

>  Your account is unverified. Please verify by clicking on the confirmation link that has been emailed to you.

Expected behavior: A way to resend the confirmation email next to the notification mentioned above. Most likely in form of a suitable button.

aditya1702:
@obnoxious23 Where did you try the email thing? On the main app? It might not have been sent because we have run out of emails

mariobehling:
@aditya1702 Yes, this is the case. Question here is: In case an email was lost, how can a user request a new confirmation mail. Check how twitter and other social networks deal with this to understand this feature.

aditya1702:
@mariobehling Ok.

mariobehling:
@aditya1702 I think part of this process should be that the user gets a visual confirmation after he/she clicks on the accept link in the email.

Currently the user clicks on the email and is directed to the homepage.

Expected behaviour, the user is directed to the contact settings page and gets a message "Thank you. Your new email is now confirmed."

PR #2298, opened 2016-08-19
Provide a way to resend email confirmation to the users

aditya1702:
#2273


aditya1702:
@mariobehling @niranjan94 Please review

Issue #2172, opened 2016-08-12
Implement analytics plugins and add tab in admin settings section for configuration


mariobehling:
In order to add analytics capabilities to the server we need key and/or code options for each page. Please implement the following:
* [x] Support for Google Analytics
* [ ] Support for Piwik
* [x] Add keys/config options as a **subtab** "Analytics" in `admin/settings/`

![screenshot from 2016-08-12 11 53 00](https://cloud.githubusercontent.com/assets/1583873/17619129/b52d8238-6083-11e6-9d0a-cf36a0e47a30.png)

mariobehling:
Analytics ID seems to be hard-coded. There is an un-known ID in the code as below.

  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-58955678-3', 'auto');
  ga('send', 'pageview');

SaptakS:
@mariobehling I am already working on it..

mariobehling:
The user is already clear that he/she is in the settings tabe. Please keep the names short as we will be lacking tab space:
Analytics Configurations -> Analytics

Same would be good with images
Image Configurations -> Images

SaptakS:
@mariobehling as I understand to use Piwik we need to install it first and then use it. Can you please guide me about it? I haven't really used Piwik before.

mariobehling:
Let's postpone this then. Thank you.

mariobehling:
Closing this and will open an issue for Piwik when time has come.

PR #2347, opened 2016-08-23
Made Analytics configurable

SaptakS:
#2172 

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -39,6 +39,7 @@ import sqlalchemy as sa
 
 from nameparser import HumanName
 import stripe
+from app.settings import get_settings
 from app.helpers.flask_helpers import SilentUndefined, camel_case, slugify, MiniJSONEncoder
 from app.helpers.payment import forex
 from app.models import db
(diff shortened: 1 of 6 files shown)


app/__init__.py
```diff
@@ -165,6 +166,13 @@ def get_locations_of_events():
 
     return dict(locations=get_locations_of_events)
 
+
+@app.context_processor
+def get_key_settings():
```
niranjan94: Just noticed, The settings are already being loaded into a `settings` variable ... (you can find that by scrolling down ... around line `200` ) ... You can use that variable

app/templates/gentelella/admin/analytics.html
```diff
@@ -1,10 +1,12 @@
+{% macro render_analytics(test='') %}
 <script>
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
   (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
   })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
-  ga('create', 'UA-58955678-3', 'auto');
+  ga('create', '{{ key_settings.analytics_key }}', 'auto');
```
niranjan94: You could use `{{ settings.analytics_key }} `` instead

niranjan94:
@SaptakS please see inline comments

Issue #2019, opened 2016-08-02
Create a subtab "Image Options" in Admin/Settings, implement central config location and display options in subtab


mariobehling:
In a subtab "image options" in https://open-event-dev.herokuapp.com/admin/settings/ we can later implement detailed image settings. As a first step list the hardcoded image settings of the different areas with pixel width and height, resolution and compression and define a central place where all image configs are stored.

* [ ] Define and implement place of image config

Display info for config on subtab image options including:
* [ ] Front page event images
* [x] Speaker images on event page 
* [x] Event image in MySessions https://open-event-dev.herokuapp.com/events/mysessions/
* [x] Speaker Images in event dashboard e.g. https://open-event-dev.herokuapp.com/events/85/speakers/
* [ ] Any other location where images are appearing

mariobehling:
Looking at the current implementation at http://open-event-dev.herokuapp.com/admin/settings/, I am unclear if these are maximum widths and if there is any ratio implementation. Are images cropped etc?

Please also sort the subtabs in a logic order e.g. System Tokens, Microservices, Image Configurations, Payment Gateways, Ticket Fees

mariobehling:
There are also a number of pages with images of other sizes missing and while all pages in the admin section have a green save button on the bottom right, it is different here.

I think I mentioned it before: Please look how other systems are doing it, e.g. Wordpress and Drupal.

There are different "category" sizes of images, e.g. Thumbnail, Medium Size, Large Size. If more different sizes are needed e.g. in Drupal you can define your specific custom size. A size that I see we need additionally is for example "Icon".

The strategy would be  to list which pages use "Thumbnails" and then make the thumbnail size configurable.

I see you are missing pages like event pages as well and config pages that have very small icons.

Looking at how you got things doing it seems you did not have investigate how others are doing things. In Wordpress you would see the following though, which would already get us moving forward.

![screenshot from 2016-08-18 17 19 00](https://cloud.githubusercontent.com/assets/1583873/17779545/546ffb4e-6568-11e6-86e8-c552dfe60c56.png)

SaptakS:
@mariobehling  I was also thinking of doing something similar because my implementation would have required saving too many copies of the image. I will get it done asap. Thanks for clearing the concept.

mariobehling:
Thank you. Good step forward!

Please also add all other pages that have images now. What comes to my mind:
* [ ] Profile Image and Speaker Image: The option here should only allow one pixel parameter as height = width
[...]

PR #2373, opened 2016-08-24
Aspect ratio, Profile Info and Logo Info added

SaptakS:
#2019 

fireshot capture 45 - settings - open event - http___localhost_5000_admin_settings_

@mariobehling please review.


mariobehling:
Please add the following info
* [ ] "Maintain Aspect Ratio" -> "Standard Aspect Ration is .... . Tick to maintain ratio."
* [ ] Add quality settings in  %
* [ ] Actually in the way you showing the config there is no need to enable the user to choose Icon/Thumbnail/Large - cause the user cannot change the size anyways as it would change the style, correct? 
   * [ ] So simply list the all pages that use a certain size below "Maintain Aspect Ratio"
   * [ ] Also add here information where on the page the image is located, e.g. the public event pages have the top background image, but there are also speaker images.
* [ ] Profile/Speaker images - I am not aware of any area where there is a large size. If there is none, then take out the configuration for that, please.

mariobehling:
I think for profile/speaker images there are three sizes, but not the large one:
* [ ] Icon, e.g. top right profile image
screenshot from 2016-08-24 19 08 13


* [ ] Small, e.g. Event Dashboard speakers table http://open-event-dev.herokuapp.com/events/9/speakers/
screenshot from 2016-08-24 19 08 30

* [ ] Thumbnail, e.g. on public event page and on profile page
screenshot from 2016-08-24 19 09 07

SaptakS:
@mariobehling will complete the things right away..

SaptakS:
fireshot capture 46 - settings - open event - http___localhost_5000_admin_settings_

@mariobehling please check..

mariobehling:
Thank you. Looks good. So, let's try out how this works in the live system. Cheers.

Issue #2335, opened 2016-08-22
User icon coming from external service


mariobehling:
It looks like the user icon comes from Google instead from our own service.

![screenshot from 2016-08-21 22 12 15](https://cloud.githubusercontent.com/assets/1583873/17844371/f6ad398a-6838-11e6-9fd5-2a15f13fa9a8.png)

![screenshot from 2016-08-21 22 11 35](https://cloud.githubusercontent.com/assets/1583873/17846907/80b3004e-684b-11e6-8be0-87a385e512f0.png)

mariobehling:
Sorry, but I still see images are downloaded from Google after the merge.

aviaryan:
@mariobehling I just signed in using g+ and the photo is indeed being download to our server. Maybe in you case, it was an old profile. Here is the link I got for my profile image. 

https://opevbucketname.s3.amazonaws.com/users/7/avatar/anZYamExbX/1472539024.83.jpg

niranjan94:
@mariobehling could this issue be closed ?

mariobehling:
Sure.

PR #2375, opened 2016-08-24
Download files in profile page

rafalkowalski:
#2335 @niranja 94 please review


mariobehling:
Images are still loaded from Google though.

PR #2397, opened 2016-08-27
Update height on changing width with aspect ratio.

SaptakS:
#2019 
- [x] Aspect ratio ticked by default
- [x] Change height on changing width when aspect ratio ticked and vice versa.

Also made some changes in the logic of storing the image size settings.

@mariobehling @niranjan94 @rafalkowalski please review.


mariobehling:
Looks like quality settings and logo size settings disappeared after upgrade.

screenshot from 2016-08-27 17 44 27

SaptakS:
@mariobehling I think this happens because I updated the way things are handled in DB but the old DB values are still there in the live system. So I will fix that..

SaptakS:
@mariobehling fixed. http://open-event-dev.herokuapp.com/admin/settings/ Please check and confirm.

Issue #2436, opened 2016-09-25
Shift from Dockerfile to Docker Compose


DengYiping:
I have implement Dockerfile during GCI, and I believe it is time for us to switch to Docker Compose. Docker Compose will separate the database and the container, and this application will be more maintainable. 

I certainly can do that, but as a mentor of Codeheat, I think it is better to leave this task to Codeheat participants. And if this task is pending for a long time, I will do it by myself.

It should be a straight forward task:
- Modify Dockerfile and remove the database setup
- Add docker-compose file, expose the port, link with database

aviaryan:
It has been removed. Please check. I migrated Dockerfile to Docker compose in this GSOC.

amitkumarj441:
@DengYiping I just exposed the port in docker compose file and added the Postgres volume.

DengYiping:
It’s fine, I didn’t see there is a docker-compose file @amitkumarj441

amitkumarj441:
@DengYiping Please have a look into the docker-compose file [77f7dd2](https://github.com/fossasia/open-event-orga-server/commit/77f7dd2380dec51d0b277b84bdd3bec51c3cb4f1)

PR #2437, opened 2016-09-25
Update docker-compose.yml

amitkumarj441:
added exposed port and postgres volume #2436

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,8 +13,8 @@ services:
     image: postgres:9.5
     environment:
       POSTGRES_PASSWORD: 'test'
-    volumes:
-      - 'pgdata:/var/lib/postgresql/data/'
+    volumes_from:
+      - data
     ports:
       - '5432:5432'
 
@@ -29,7 +29,7 @@ services:
   web:
     build: .
     expose:
-    - "8000"
+    - "5000"
     environment:
       DATABASE_URL: postgresql://postgres:test@postgres:5432/opev
       REDIS_URL: redis://redis:6379/0
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,15 @@ services:
     volumes:
       - /var/lib/postgresql
     command: echo true
+    
+  nginx:
+    build: .
+    ports:
+        - "80:80"
+    volumes_from:
+        - data
+    links:
+        - data:data
 
   postgres:
     image: postgres:9.5
@@ -38,6 +47,7 @@ services:
     links:
       - postgres:postgres
       - redis:redis
+      - nginx:nginx
     volumes:
       - 'storage:/open_event/static'
     ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,8 +13,8 @@ services:
     image: postgres:9.5
     environment:
       POSTGRES_PASSWORD: 'test'
-    volumes_from:
-      - data
+    volumes:
+      - 'pgdata:/var/lib/postgresql/data/'
     ports:
       - '5432:5432'
 
@@ -28,6 +28,8 @@ services:
 
   web:
     build: .
+    expose:
+    - "8000"
     environment:
       DATABASE_URL: postgresql://postgres:test@postgres:5432/opev
       REDIS_URL: redis://redis:6379/0


aviaryan:
@amitkumarj441 

Thank you for your contribution. Here are some points - 

1. It is recommended to have the postgres data container stored in the same image as that of postgres. I don't know the exact reason for this but I have read about this a lot. Your change will remove that.

2. The server runs on port 5000 by default and it is exposed as port 80 by the `80:5000` setting. So I don't understand what is the need of the addition you made. 

Please clarify these things. Also please see INSTALLATION_DOCKER.md , it contains some info on how to run Open Event using Docker.

amitkumarj441:
@aviaryan It was my mistake, I hadn't checked the port but now it's fixed because the application is fired on port 8000. Thanks for letting me know about the existing installation file of docker. Can I add a service to the docker compose file for reverse proxy to forward requests to the static file directory.

aviaryan:
Please see my comment above. I don't see a need for exposing port 5000. The docker version of server runs on port 80 by default and it is the default port of any system. So orga-server can be easily hosted on any cloud service online without having to deal with ports. 

> The server runs on port 5000 by default and it is exposed as port 80 by the 80:5000 setting. So I don't understand what is the need of the addition you made.

--- 

> Can I add a service to the docker compose file for reverse proxy to forward requests to the static file directory.

Can you please explain more about this. What will happen if you *"forward requests to the static file directory."*.

shivamMg:
@amitkumarj441 Do you mean using another server like nginx to serve static files and and forward other requests to the application server?

amitkumarj441:
@shivamMg Alright! That's what exactly I'm trying to put that nginx into docker-compose file!

shivamMg:
Nginx is in our [stack](https://github.com/fossasia/open-event-orga-server/blob/development/install.sh) but it is not configured to serve any files. It would need configuration to serve static files and to pass other requests to gunicorn. I have the config file if you need. But changes are required for both Vagrant and Docker.

amitkumarj441:
@shivamMg Thanks for letting me know about the existing stack! I'll do it right away for Docker and then Vagrant!

niranjan94:
@aviaryan @shivamMg could one of you review this please ? I don't have docker setup to test it.

docker-compose.yml
```diff
@@ -8,6 +8,15 @@ services:
     volumes:
       - /var/lib/postgresql
     command: echo true
+    
+  nginx:
+    build: .
+    ports:
+        - "80:80"
```
aviaryan: I don't understand the need of this. The docker-compose service "nginx" runs ./Dockerfile as the build command. 
Essentially this is the same as service "web". So why run 2 web services ? Also the ./Dockerfile runs nothing on port 80 so this should not do anything. 
**I think nginx has no role to play in our system.**
@shivamMg Can you please provide your views on this.

docker-compose.yml
```diff
@@ -28,6 +37,8 @@ services:
 
   web:
     build: .
+    expose:
+    - "5000"
```
aviaryan: @amitkumarj441 Can you please clarify why this is needed ?

docker-compose.yml
shivamMg: The web service has already taken port 80 in host. The correct way would be to assign Nginx the port 80 in host (like above). Nginx would directly serve the static assets and redirect other requests to the web (gunicorn server) service. These redirects would be inter-container requests. 

Another way of doing it could be just using the nginx server installed inside the web service. Run it on 5000 port (port on which gunicorn now runs on), assign gunicorn to some other port, and do the same redirection like above. In this nginx and gunicorn would run on the same container.

Gunicorn shouldn't be used to serve static assets in production. Nginx can be used as a proxy server to serve static content directly and pass other requests to gunicorn. I think we should discuss it in the chat first. @amitkumarj441 @aviaryan

amitkumarj441: @shivamMg Alright! I'll ping this issue on [Gitter](https://gitter.im/fossasia/open-event-orga-server).

docker-compose.yml
amitkumarj441: @aviaryan In the above web directory, the application is fired up on port 5000 and then the same port is then forwarded to port 80 on the host environment, but for this service we should add environment variables to persist our container running from the resulting image.

docker-compose.yml
aviaryan: @shivamMg I understand now. Thank you. 
@amitkumarj441 
But still, using '.' as the build file won't do anything. It will just re-run the web service (which is pointless). In this case, we will have to create a new Dockerfile for nginx and use it in the nginx service build. See [this article](https://realpython.com/blog/python/dockerizing-flask-with-compose-and-machine-from-localhost-to-the-cloud/) for an example.

amitkumarj441: @aviaryan Yes, alright! Where(folder) should I push the dockerfile for nginx?

niranjan94:
@aviaryan @shivamMg @amitkumarj441 what's the status of this ? This PR has been open for a long time :sweat_smile:

shivamMg:
I think a new PR would be better. This one would require a lot of changes.
@amitkumarj441 You can close this PR and make a new one.

amitkumarj441:
@niranjan94 Please do open the existing issue and do assign me the docker tasks so that I can make PR  in this issue.

PR #2443, opened 2016-09-29
librabbitmq version set to 1.5.1

SaptakS:
librabbitmq version 1.6.1 build-out fails in Fedora ( [issue](https://github.com/celery/librabbitmq/issues/61) ). So hardcoding the version of librabbitmq to 1.5.1 for the time in which it works fine.

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -43,7 +43,7 @@ stripe
 xhtml2pdf
 nameparser
 pillow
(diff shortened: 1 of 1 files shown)


aviaryan:
Wow. That seems to be a way old issue (from 2014) .. Fixing the version to 1.5.1 seems like the correct thing to do here.

PR #2448, opened 2016-10-01
Fix SyntaxError for porting to Python 3

fedexist:
Now *.py files should be compatible with Python 3. I think it's worth mentioning that not all the packages required are compatible with Python 3 and therefore should be changed accordingly. In this PR I merely made ported the files.


aviaryan:
> I think it's worth mentioning that not all the packages required are compatible with Python 3 and therefore should be changed accordingly.

That was what I was worried about the most. Finding replacements for those libraries can take a lot of time and can even return inconsistent results. Let's leave this issue now, as suggested by @mariobehling  we can upgrade to Python 3 once the application has reached production stage. Thanks for your work. :beers: 
Will close this PR now.

PR #2476, opened 2016-10-15
Allow organizer to enter discount code when creating an event

### LINKED ISSUE #2324: Implement discount code for events and marketers
As we need to get marketers support to get the event system used, we need a way to identify which event organizer comes through which marketer. A way to achieve this is through discount codes.

To do this we need a way to create discount codes and link them to specific marketers accounts. So, for example a marketer gives out a discount code "X20482TOP" and that discount code gets organizers off 10% for an event over three months. Then in the monthly report the event organizer needs to get 10% off from the money he needs to pay.

In the admin sales overview we should also have a column "Marketer" to show all events with discount codes of a specific marketer. Based on the sales we can then share the income with a marketer.

Features we need are:

* [x] Implement Role "Marketer" as System role https://github.com/fossasia/open-event-orga-server/issues/2453
* [x] Possibility to create event discounts with percentage, number of times the discount voucher is valid and user account the voucher is linked to
* [x] Show of discounts in column in sales and show of linked marketer
* [x] Calculate discount in sales monthly
* [x] Implement Sales overview for events of marketers (where and how would that be best?)
* [x] In admin tab "Sales" show subtab sales "By Marketer" https://github.com/fossasia/open-event-orga-server/issues/2452
* [x] In admin tab "Sales" show subtab sales "By Discount Code" https://github.com/fossasia/open-event-orga-server/issues/2452

niranjan94:
Related issue: #2324.


niranjan94:
@mariobehling I have fixed the tests. Please review and merge.

PR #2490, opened 2016-10-21
Autoselect timezone based on user location

### LINKED ISSUE #2472: Preselect timezone when creating new event
When I create a new event, the Timezone dropdown should have a default value based on my location. I am in Vietnam, for example, the dropdown should be "UTC+7" by default.

niranjan94:
Resolves #2472 

@mariobehling  please review and merge


app/views/admin/models_views/events.py
```diff
@@ -3,15 +3,18 @@
 import os
 import traceback
 import binascii
+from urllib2 import urlopen
```
SaptakS: @niranjan94 can you please check this?

niranjan94: Will do :)

SaptakS: Will merge after that then.

PR #2493, opened 2016-10-24
Added super admin email to settings

### LINKED ISSUE #2230: Add to super admin settings field with email
Email field is neccessary to define, because we don't won't to hardcode email in code

niranjan94:
Resolves #2230


mariobehling:
Very good.

Issue #2046, opened 2016-08-04
Propose Internationalization Flow


mariobehling:
How can we implement internationalization? Things to take into account:
- [x] Translate UI and make it country specific, e.g. offer Canada (EN) and Canada (FR)
- [x] Multilanguage Version of Pages https://open-event-dev.herokuapp.com/admin/content/
- [x] Names of locations e.g. cities have different names in different languages
- [x] Propose workflow for volunteers to translate system

![screenshot from 2016-08-04 09 32 59](https://cloud.githubusercontent.com/assets/1583873/17393911/b706070a-5a26-11e6-972e-40d3ab1d8d3d.png)

rafalkowalski:
I think that Flask Babel https://pythonhosted.org/Flask-Babel/ library will be enough, to translate UI. Then we need to define a page where volonteers can easily edit page content.

aditya1702:
@mariobehling Yes we can implement it using Flask Babel

mariobehling:
> Then we need to define a page where volonteers can easily edit page content.

How? Where? I do not see this kind of option in Babel. As far as I can see a solution to edit pot/po files is needed.

rafalkowalski:
So My idea is to create tab in admin panel where volunteers can choose language then have a possibility to edit phrases via simple html form.

mariobehling:
@rafalkowalski We do not really want to handle volunteers directly on the system. It would be better if we could upload po/pot files (whatever is needed) into the system directly. So it would be good to:
* [ ] a) to have the ability to download the original po/pot as a sample for translation (and that can be used in translation applications like rosetta
* [ ] b) have the ability to upload po/pot files for new/updated languages directly into the system
* [ ] c) have a folder where po/pot's can be added in the repo

mariobehling:
As we are getting closer to the release and still do not have an internationalization implementation this is a blocker now.

SaptakS:
@mariobehling I think we can use [this](http://blog.miguelgrinberg.com/post/the-flask-mega-tutorial-part-xiv-i18n-and-l10n)

mariobehling:
@SaptakS That is the same that was suggested in this thread already, isn't it? So, if everyone is suggesting this. Let's go with it.

sainiak009:
This is how I propose to implement it : 

- I will define a "Languages" dictionary in config.py where all the available languages will be listed.

[...]

PR #2506, opened 2016-10-27
Resolves #2467 and #2046 : Strings marked for I18N in templates

### LINKED ISSUE #2467: Mark strings in views/templates that would be locale-specific
Parent Issue: #2046

All strings in views/templates which will need translation needs to be marked.

sainiak009:
I have scanned through all the templates and tried my best to mark all strings. Please review and let me know if any error of missing strings. New .po/.pot files are generated. I have generated translations from Google Translator Toolkit, Most of the strings are converted but few are left and can be edited manually by downloading .po file by a volunteer.


sainiak009:
@niranjan94 Please have a look.

app/templates/404.html
```diff
@@ -5,7 +5,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>You got 404'd - HTTP 404 - Page Not Found - Open Event</title>
+    <title>{{_('You got')}} 404'd - HTTP 404 - {{_('Page Not Found')}} - {{_('Open Event')}}</title>
```
niranjan94: Don't mark **Open Event** for translation. That would be name of the service. EventYay.

mariobehling: Yes, the name of the service is generic. Someone else could set it up using a different name for example.

sainiak009: @mariobehling @niranjan94 "Open event" is removed from marked strings

sainiak009:
@niranjan94 @SaptakS Please review and merge

PR #2514, opened 2016-10-31
installation commands for postgreql,nodejs added

In-vincible:
Added them in the dependency menu only.


aviaryan:
I am not sure whether apt-get install nodejs is the way to go. It may install a very old nodejs version. 
http://askubuntu.com/questions/594656/how-to-install-the-latest-versions-of-nodejs-and-npm-for-ubuntu-14-04-lts
@In-vincible @SaptakS views ?

SaptakS:
@aviaryan valid point. @In-vincible can you please make the necessary changes?

In-vincible:
@SaptakS yeah I am into making the changes.

In-vincible:
@SaptakS should I add a curl installation or manual one?

In-vincible:
@aviaryan @SaptakS I have added curl installation for nodejs v6.x , since v6 is LTS

niranjan94:
@In-vincible @aviaryan @SaptakS I think [nvm](https://github.com/creationix/nvm) is a better option. Gives the user freedom to have multiple node versions on their dev machine.

In-vincible:
@niranjan94 nvm is indeed better option, which version of nodejs We should add for default installation

niranjan94:
@In-vincible the latest LTS `v6.9.1` should do

In-vincible:
@niranjan94 I am adding a command for nvm installation too..is it required?

niranjan94:
@In-vincible sure ... But be sure to specify that, the step can be skipped if they already have nvm or node installed

In-vincible:
@niranjan94  any specific reason why coveralls test failing ?

niranjan94:
@In-vincible it isn't supposed to decrease since only .md files have been modified ... But, no issue. We can ignore it for now,.

PR #2522, opened 2016-11-07
Continuous deployment to Google Container Engine

niranjan94:
Related issue: #2495

_This should work. But, I cannot be sure unless it's merged into development._

@mariobehling please review and merge

Issue #2495, opened 2016-10-25
Set up System on Google Cloud and configure automatic deployments


mariobehling:
Kubernetes enables applications to scale up or down based on the needs. Please set up eventyay on Google cloud. 

Discussed set up:
- deploy DB on one system
- deploy other parts on separate container(s)

Other:
- Document how to ensure updates (continues integration/continuous deployment)

niranjan94:
**TODO:**
- [ ] Persistent storage disk for PostgreSQL & Redis
- [ ] Fix `404` error on deployment by properly setting SERVER_NAME during deployment 
- [ ] Continuous deployment on each GitHub push (like heroku)

mariobehling:
>  Continuous deployment on each GitHub push (like heroku)

I guess, we should be adding another branch "production" after an initial test period.

niranjan94:
@mariobehling correct. And that branch should also have tagged releases.

PR #2523, opened 2016-11-07
Fix travis build fail due to gcloud client installation

niranjan94:
Follow up to #2522 

Related issue: #2495


mariobehling:
Still errored.

PR #2536, opened 2016-11-12
Updating INSTALLATION_LOCAL.md

mohitsaini12628:
There are various commands in the file that does not mention in which directory the user should be present before running the command. Also, there are various commands which are different for mac users. I updated some of them. Will update the rest as I find


mohitsaini12628:
You mean to squash all the commits into one single commit ?

PR #2558, opened 2016-11-16
Added app liveliness probe in kubernetes

niranjan94:
Added liveliness probes so that kubernetes can automatically restart containers if the app goes offline

@SaptakS @mariobehling please review and merge.


mariobehling:
Please check: branch is out-of-date with the base branch

Issue #2559, opened 2016-11-16
Session type duration picker not in HH/MM format 


niranjan94:
Session type duration picker not in HH/MM format when a new session type is added.

PR #2563, opened 2016-11-17
Use HH/MM for session length picker

### LINKED ISSUE #2561: Event delete button does not work
The delete button on the following page does not work
http://open-event-dev.herokuapp.com/events/

This is what I get after clicking on the delete button

screenshot from 2016-11-16 22-47-44

niranjan94:
Resolves #2561


SaptakS:
Resolves #2559

PR #2572, opened 2016-11-18
Added www to naked domain redirect via an nginx redirector

niranjan94:
**In this PR:**
- Change internal web deployment port to `8080`
- Redirect `www.eventyay.com` to `https://eventyay.com`

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -483,14 +483,6 @@ def after_request(response):
                                                                                                  query.context))
     return response
 
-# HTTPS Redirect
-@app.before_request
-def before_request():
-    if request.url.startswith('http://') and current_app.config.get('FORCE_SSL', False):
-        url = request.url.replace('http://', 'https://', 1)
-        code = 301
-        return redirect(url, code=code)
-
 # Flask-SocketIO integration
 socketio = None
 if current_app.config.get('INTEGRATE_SOCKETIO', False):
--- a/kubernetes/yamls/redirector/redirector-deployment.yml
+++ b/kubernetes/yamls/redirector/redirector-deployment.yml
@@ -1,0 +1,23 @@
+kind: Deployment
+apiVersion: extensions/v1beta1
+metadata:
+  name: redirector
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: redirector
+    spec:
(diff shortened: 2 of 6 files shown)


mariobehling:
@niranjan94 I am not clear, why the domain redirect has to be taken care of in the code here. Can we simply change the DNS entry and add a redirect there rather than in the code?

niranjan94:
@mariobehling 

A DNS change cannot do a redirect. A DNS can only _point_ a domain/subdomain to a server.

For example,
If we set a `CNAME` at `www.eventyay.com` to point to `eventyay.com`,

We'll not be getting a redirect. Instead, the same web app will be loaded at both domain which isn't right.

mariobehling:
@niranjan94 I might not be using the correct terminology, but a URL redirect outside of the code here would be preferable.

How can we keep our specific eventyay configuration outside of the code here?

niranjan94:
@mariobehling server configuration, most often than not, is usually deployment specific ... And since we have our server configuration here (kubernetes), I guess we may end up having server-related things like this redirect here ...

niranjan94:
But I understand what you're saying ...

niranjan94:
@mariobehling I'm not sure what can be done here. 

Apart from redirection, there are other parts of the kubernetes configuration, that's specific to the **Event Yay** GCE Deployment.
1. The eventyay.com domain names specified in `kubernetes/yamls/web/ingress-notls.yml` and `kubernetes/yamls/web/ingress-tls.yml`
2. The external IP address specified in `kubernetes/yamls/nginx/service.yml`

We don't seem to be having an alternative to doing it this way. As kubernetes expects some parameters in the configuration file itself. 

So, 

> How can we keep our specific eventyay configuration outside of the code here

The only way to completely do this, is if we can have two sets of kubernetes configuration. One eventyay specific and another one as generic as possible.

niranjan94:
If this

> The only way to completely do this, is if we can have two sets of kubernetes configuration. One eventyay specific and another one as generic as possible.

is alright, please merge this PR and I'll create two sets of configuration.

mariobehling:
> create two sets of configuration.

Good point. Please open an issue and do this later. We need to get eventyay running first.

Issue #1991, opened 2016-08-01
Documents required for running event system


mariobehling:
In order to make the system wholly functional we also need to get documents up including
- [x] Imprint
- [x] Terms and references that comply with laws
- [x] Copyright Information
- [ ] Documentation - How it works (https://github.com/fossasia/support.eventyay.com/issues)
- [ ] Documentation - FAQ (https://github.com/fossasia/support.eventyay.com/issues)
- [x] Privacy info
- [x] Cookies Info
- [ ] Developers Page

bradstimpson:
Hi Folks, I can help out here.  What document has the highest priority?

mariobehling:
Hi, there is no special priority. Any help is welcome. Thank you!

SaptakS:
@bradstimpson are you still working on this issue?

niranjan94:
Documentation can be added at https://github.com/fossasia/support.eventyay.com/issues.

PR #2582, opened 2016-11-21
Added imprint, terms and privacy policy for eventyay

niranjan94:
Related issue: #1991

Documents can be previewed [here](https://github.com/niranjan94/open-event-orga-server/tree/documentation/eventyay-docs/docs/eventyay).


mariobehling:
@niranjan94 @SaptakS I think this PR is a misunderstanding and it should be reverted.

The page content should not be added into the code. There should be pages, that are filled from the admin UI of the system.

niranjan94:
@mariobehling it hasn't been added in code. Just as an MD file

mariobehling:
Yes, I understand. It should not be in the repo then. Please revert.

PR #2591, opened 2016-11-24
Fix file upload issue in kubernetes

niranjan94:


mariobehling:
Let's make a merge into Master as well to check that this is all working.

niranjan94:
@mariobehling not necessary ... eventyay is still using development for now

PR #2597, opened 2016-11-24
Displayed most recent commit on kubernetes deployment

niranjan94:
Currently in Heroku in the super admin page we have information about the most recent commit that has been deployed. It is very useful when debugging possible regressions. This PR adds the same information for kubernetes deployments too.


mariobehling:
Thanks. Please check: branch is out-of-date.

PR #2600, opened 2016-11-26
Merge development into master

niranjan94:


SaptakS:
@niranjan I am sending the PR for the speaker image. It is an important fix as without it none of the public speaker forms image upload works. After that maybe we can merge.

niranjan94:
@SaptakS cool :)

Issue #2605, opened 2016-11-28
Clean up python dependencies


niranjan94:
We have loads of python dependencies of which a few are totally unused and can be removed or replaced with a few lines of python code instead of an external dep. 

**Tasks:**
- [x] Find the dependencies that are unused and remove them (#2689)
- [ ] If any dependency is just used for a very small function or purpose which could be replaced with vanilla python, replace the dependency

nikhilrayaprolu:
@niranjan94 I would like to work on this issue . can you please point the file  where the dependencies are mentioned.

niranjan94:
@nikhilrayaprolu inside the `requirements` folder (https://github.com/fossasia/open-event-orga-server/tree/development/requirements)

nikhilrayaprolu:
@niranjan94 out of all those requirements can you sort out those which can be removed. I will work on those requirements and find alternative python codes to replace them

niranjan94:
@nikhilrayaprolu well ... that what this issue is about... I have updated the issue description.

mamakancha:
I would like to get this issue assigned

niranjan94:
@mamakancha  ... You can proceed :)

mamakancha:
@niranjan94 Hi. Can I get a starting point? As in I am not able to understand how to go about solving this issue.

niranjan94:
We have all the dependencies for this project listed inside the `requirements` folder (https://github.com/fossasia/open-event-orga-server/tree/development/requirements). 

But there maybe some dependencies that are unused. Or some dependencies that can be easily replaced by just a few lines of python code instead of big external dependency. 

You'll have to go through all the dependencies and figure out which are used and which aren't and remove the ones that are not used.

If any dependencies can be replaced by a few lines, replace them and remove the dependency.

PR #2627, opened 2016-12-02
Re-structure views/routes & cleanup

niranjan94:
- Restructured views and routes
- Re-arranged all view files
- Removed `Flask-Admin` dependency. (we have no use of more than 80% of its features) (#2605)
- Removed `WTF-Forms`,`Flask-WTF` dependencies. (we don't use it at all. It was used in v1) (#2605)
- Remove a few unused methods from helpers

~_The tests would probably fail and need some fixing_~


niranjan94:
The travis build has been fixed. 
@SaptakS @mariobehling please review and merge

Issue #2649, opened 2016-12-07
Start page: Longer addresses change height of box and break styling


mariobehling:
Some addresses of event locations are longer requiring two lines on the overview page. This results in a different height of the box on the start page and breaks the style.

Please change the styling to ensure all boxes are displayed in rows.

![screenshot from 2016-12-07 07-46-10](https://cloud.githubusercontent.com/assets/1583873/20950288/7780420c-bc51-11e6-8d2a-13a3269bdf78.png)

SaptakS:
@magdalenesuo thanks for the pull request.

PR #2665, opened 2016-12-11
address length problem (issue #2649)  fixed

magdalenesuo:
Issue #2649 fixed- added tooltip

(https://cloud.githubusercontent.com/assets/20799954/21082256/ddc17b62-bffd-11e6-9e70-ca8efdc9c7a1.png)

### DIFF
--- a/app/templates/gentelella/index.html
+++ b/app/templates/gentelella/index.html
@@ -24,7 +24,7 @@
         }
 
         .event-holder .thumbnail {
-            padding: 0px;
+            padding: 0;
         }
 
         .event-holder .time {
@@ -48,7 +48,7 @@
         .event-holder .caption .locate{
             visibility: hidden;
             overflow: visible;
-            width: 150px;
+            width: 180px;
             border-radius: 6px;
             padding: 5px 0;
             background-color: #444444;
@@ -164,7 +164,7 @@
                     <small class="time">{{ event.start_time.strftime('%a, %B %d %I:%M %p') }}</small>
                     <h4 class="name">{{ event.name }}</h4>
                     <div class="location_name">{{ event.location_name.split(",")[0] }}</div>
-                    <span class="locate">{{ event.location_name.split(",")[0] }}</span>
+                    <span class="locate">{{ event.location_name }}</span>
                     <hr>
                     <p class="tags">
                         {% if event.type %}
--- a/app/templates/gentelella/index.html
+++ b/app/templates/gentelella/index.html
@@ -24,7 +24,7 @@
         }
 
         .event-holder .thumbnail {
-            padding: 0;
+            padding: 0px;
         }
 
         .event-holder .time {
@@ -36,8 +36,30 @@
             text-overflow: ellipsis;
         }
 
-        .event-holder .location_name {
+        .event-holder .location_name { 
+            white-space: nowrap; 
+            width: 12em; 
+            overflow: hidden;
+            text-overflow: ellipsis; 
+            position: relative;
+            display: inline-block;
+        }
 
+        .event-holder .caption .locate{
+            visibility: hidden;
+            overflow: visible;
+            width: 150px;
+            border-radius: 6px;
+            padding: 5px 0;
+            background-color: #444444;
+            color: white;
+            text-align: center;
+            position: absolute;
+            z-index: 1;           
+        }
+
+        .event-holder:hover .caption .locate {
+            visibility: visible;
         }
 
         .event-holder .tags {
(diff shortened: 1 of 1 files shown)


app/templates/gentelella/index.html
```diff
@@ -41,12 +43,18 @@
...
+            transition: background-color 0.5s;
+            height: 18px; 
+            padding-right: 10px;
         }
 
         .event-holder .caption {
             background-color: #FFFFFF;
             transition: background-color 0.5s;
+            height:140px;
+
```
SaptakS: Please remove the unwanted spaces.

niranjan94:
@magdalenesuo multi-line location doesn't look good. The event box looks tall and many times the location name is only a single line. 

Instead of increasing the height to accommodate multiple lines, keep the height as it is, and if the location name goes beyond a length, maybe ellipsize  it and show it as a tooltip or something ? 

Most of the use cases, the user would be filtering for events close by to his location and anyway only first few words of the location name is gonna matter to the user 

@SaptakS what do you feel ?

niranjan94:
> Most of the use cases, the user would be filtering for events close by to his location and anyway only first few words of the location name is gonna matter to the user 

For example,

If the location name is _Science Centre, 15 Science Centre Rd, Singapore_ , 
for a user who is searching for events in singapore, only __Science Centre_ is gonna matter.

app/templates/gentelella/index.html
magdalenesuo: removes

app/templates/gentelella/index.html
```diff
@@ -142,7 +150,9 @@
                     <small class="time">{{ event.start_time.strftime('%a, %B %d %I:%M %p') }}</small>
                     <h4 class="name">{{ event.name }}</h4>
                     <small class="location_name">{{ event.location_name.split(",")[0] }}</small>
-                    <hr>
+                    
+                </div>
+                <hr>
                     <p class="tags">
```
magdalenesuo: yes, fixed

SaptakS:
@niranjan I agree with you but the problem is sometimes the first few part of the location itself goes beyond length. In that case problem arises. If that doesn't happen there is no problem anyways. So not really sure what should be the thing. Tooltip doesn't see good. We may just truncate in that case?

SaptakS:
> If the location name is Science Centre, 15 Science Centre Rd, Singapore,
for a user who is searching for events in singapore, only Science Centre is gonna matter.

Exactly. In fact if it is comma separated it doesn't create an issue at all. The issue is only if someone write like Science Centre 15 Science Centre Rd Singapore or if the name itself is that big. Then we have a problem.

I think we should be giving a fixed size and not allow the box to have different sizes. The only question is whether to make it shorter or longer?

SaptakS:
@niranjan what do you think?

niranjan94:
@SaptakS I'm not talking about a split by comma ... Just ellipsize it when it goes beyond the width of the box (I have used a js function to do that in the scheduler). Then add a title attribute to the span which has the location name with the full location. So, when a user hovers on the ellipized location name, he/she will be able to see the full name as a native browser tooltip

SaptakS:
> So, when a user hovers on the ellipized location name, he/she will be able to see the full name as a native browser tooltip

That's a good idea. Only I am not very sure whether a tooltip is necessary.

magdalenesuo:
@niranjan94  Oh I didn't know about it. Thanks , will look into it :)

app/templates/gentelella/index.html
```diff
@@ -141,7 +163,8 @@
                 <div class="caption">
                     <small class="time">{{ event.start_time.strftime('%a, %B %d %I:%M %p') }}</small>
                     <h4 class="name">{{ event.name }}</h4>
-                    <small class="location_name">{{ event.location_name.split(",")[0] }}</small>
+                    <div class="location_name">{{ event.location_name.split(",")[0] }}</div>
+                    <span class="locate">{{ event.location_name.split(",")[0] }}</span>
```
SaptakS: Inside locate, show the entire event.location.

app/templates/gentelella/index.html
```diff
@@ -24,7 +24,7 @@
         }
 
         .event-holder .thumbnail {
-            padding: 0;
+            padding: 0px;
```
SaptakS: Please change 0px to 0... 0 has no unit...

magdalenesuo:
@SaptakS @niranjan94 I have made the changes

app/templates/gentelella/index.html
magdalenesuo: changes made

magdalenesuo: (https://cloud.githubusercontent.com/assets/20799954/21082256/ddc17b62-bffd-11e6-9e70-ca8efdc9c7a1.png)

Issue #2677, opened 2016-12-20
Error in placeholder upload 


shubham-padia:
I was getting an error when uploading an image in placeholders `OSError: [Errno 2] No such file or directory: '/home/shubham/github/orga2/static/uploads/temp/temp/' ` or `OSError: [Errno 17] File exists: '/home/shubham/github/orga2/static/uploads/temp/' `.
I've fixed the problem in the code, but can someone please double check if the error exists or not ?

niranjan94:
Could you make a PR with your fix ?

shubham-padia:
closed via #2678

PR #2678, opened 2016-12-20
fix image upload errors in placeholders section

shubham-padia:
@niranjan94 Fixes errors in #2677


shubham-padia:
@niranjan94 The update button works good,just had to refresh the page.

app/views/super_admin/content.py
```diff
@@ -71,15 +72,20 @@ def placeholder_upload():
             wpercent = (basewidth / float(img.size[0]))
             hsize = int((float(img.size[1]) * float(wpercent)))
             img = img.resize((basewidth, hsize), PIL.Image.ANTIALIAS)
-            os.mkdir(app.config['TEMP_UPLOADS_FOLDER'])
+            try:
+                os.mkdir(app.config['TEMP_UPLOADS_FOLDER'])
```
SaptakS: This puts in try catch but the error still occurs, right? This time it will just pass to except block. So the upload error isn't fixed I guess? Am I missing something?

shubham-padia: @SaptakS  the error was due to the extra `  + '/temp/') ` and without putting that try catch block, i was not able to discover `OSError: [Errno 2] No such file or directory: '/home/shubham/github/orga2/static/uploads/temp/temp/' `

SaptakS: Okay.

SaptakS: I will check the error in my local system once and then approve it. @niranjan94 feel free to approve if you have checked.

shubham-padia: @SaptakS should i remove the try..catch block if you want, it makes no difference anyways.

niranjan94: @shubham-padia hold on ... lemme check it locally now

PR #2681, opened 2016-12-21
Periodic update from weblate

niranjan94:
@SaptakS please review and merge

Note: do not squash. merge as is.


SaptakS:
@niranjan94 thanks for making the note. :smile:

Issue #2308, opened 2016-08-20
Add all system images to system image settings


mariobehling:
Currently a number of images do not show up in the settings at http://open-event-dev.herokuapp.com/admin/content/

Please add images like:
- [x] Top left logo
- [x] Speaker placeholder image
- [x] Profile placeholder image
- [x] Speaker placeholder icon image
- [x] Profile placeholder icon image
- [x] Any other images

Also add:
- [x] A helper text detailing the required standard pixel width/height

The tasks to be completed in this issue are:
- [x] Implement upload/view functionality for all the above mentioned images at `/admin/content/`
- [x] Make sure the application references to the upload images throughout.

amitkumarj441:
@mariobehling I'm willing to take up this issue. Can I add these images?

niranjan94:
@amitkumarj441 you can go ahead :)

amitkumarj441:
@niranjan94 Thanks for your prompt response! Where I can initiate PRs for these images?

niranjan94:
@amitkumarj441 over here https://github.com/fossasia/open-event-orga-server/pulls just like any other PR for this project.

niranjan94:
@amitkumarj441 also, please see `/admin/content/` to know what is expected

amitkumarj441:
Thanks @niranjan94 ! I'll soon initiate some PRs in content!

niranjan94:
@amitkumarj441 what is the progress on this issue ?

amitkumarj441:
Sorry @niranjan94 for late reply!

You haven't updated this issue, we already discussed about this issue.
Please update this issue, so that I can able to make some PRs!

niranjan94:
@amitkumarj441 done. Please proceed.

shubham-padia:
@niranjan94 @mariobehling Can i work on this issue ?
 
I've roughly implemented the points 

> Implement upload/view functionality for all the above mentioned images at /admin/content/

and

> Please add images like

I can start working on 

> Make sure the application references to the upload images throughout.

if i get a go ahead.
Thank You.

niranjan94:
@shubham-padia yes. You can proceed.

niranjan94:
@shubham-padia what's the progress on this issue ?

shubham-padia:
@niranjan94 #2516 has been merged, only testing on the point 

> Make sure the application references to the upload images throughout.

Remains to make sure there aren't any discrepancies .
[...]

PR #2685, opened 2016-12-21
add support for custom logo placeholder

shubham-padia:
Regarding #2308 .
Show custom logo if the user has uploaded the logo image, otherwise app name will be printed out in the top-left side as usual.

selection_024


app/templates/gentelella/admin/menu.html
```diff
@@ -4,9 +4,25 @@
             <nav>
                 <ul class="nav navbar-nav">
                     <li style="margin-left: 10px;">
-                        <a href="/" style="font-size: 26px; font-weight: 300;">
-                            {{ app_name }}
-                        </a>
+
+                            {% set vars = {'foo': False,img: 'dummy'} %}
+                                {% for custom_image in custom_placeholder %}
+                                    {% if custom_image.name == "Logo" %}
```
niranjan94: Could you think of a more efficient way to determine if a custom logo exists or not instead of looping through everything like this ?

niranjan94: It would be awesome if you can figure something out ... We could use the same for all other placeholder images too. I feel the current implementation is very hacky.

shubham-padia: :+1:

shubham-padia:
@niranjan94 please review again.

shubham-padia:
@niranjan94 should I apply this approach for all placeholders and make another PR ?

niranjan94:
@shubham-padia sure :) that would be great

PR #2686, opened 2016-12-21
Translated some words ( using weblate )

adarsh-ideal:
Translated some words (English to Hindi) using weblate service.


niranjan94:
Periodic merges from weblate will be done by the project maintainers ... You only have to do the translation on weblate.

Issue #2644, opened 2016-12-06
Change open event sample session length format


mariobehling:
In the scheduler and publicly available schedule short sessions are not showing up properly. Compare: https://eventyay.com/events/5/scheduler/

![screenshot from 2016-12-06 16-00-01](https://cloud.githubusercontent.com/assets/1583873/20919202/35f6dee2-bbcd-11e6-9bee-11cd521447e1.png)

niranjan94:
@mariobehling it's because the session timings seem to be overlapping. Which is not supposed to happen.

Marauderer97:
Hello, I would like to work on this.. only I cannot see what exactly needs to be done.. could some one please elaborate? Where is the length attribute being displayed?

niranjan94:
> Would it be better to implement this in a time format? Like: 00:30

Sorry @mariobehling I noticed this comment only now. Currently session length is stored as a `float` in the database. That's the reason it's being displayed as `0.3` in the samples.

I think it would be better if we change it to like `00:30` (just like how it is displayed on the event wizard - step 3)

niranjan94:
@Marauderer97 

- The session type model has to be changed such that length is considered as a string instead of a float. (be sure to generate the required migrations too).
- The API definition for session type has to be changed such that length is considered as a string instead of a float. (event better would be a custom type with validation).
- All samples at `fossasia/open-event` have to be modified accordingly.

Marauderer97:
@niranjan94  Hello, sorry for the delay, the installations was giving me problems, fixed it finally, I have done the first two steps, but what does all samples at fossasia/open-event refer to exactly?

niranjan94:
@Marauderer97 these are the samples https://github.com/fossasia/open-event/

Marauderer97:
@niranjan94 Understood, thanks for patiently elaborating, so does that mean, i make two PRs, one here and one to fossasia/open-event?

PR #2705, opened 2016-12-23
Changed the length from float to string

Marauderer97:
Changed the length field's data type in model and api to string from float. @niranjan94 Hope this is what you wanted with reference to #2644


niranjan94:
@Marauderer97 you're missing the migration files for the model change

Marauderer97:
@niranjan94 I updated the branch

Marauderer97:
@niranjan94 I updated the branch

Vagrantfile
```diff
@@ -71,5 +71,7 @@ Vagrant.configure(2) do |config|
   # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
   # documentation for more information about their specific syntax and use.
   config.vm.provision :shell, :path => "scripts/install.sh"
-
+  config.proxy.http     = "http://proxy.iiit.ac.in:8080"
```
niranjan94: Please remove these lines since they're specific to your network

Marauderer97:
@niranjan94 Yes, I had added them by mistake, I reverted my change in my final commit

Vagrantfile
Marauderer97: Already fixed in the next commit. Sorry about that.

Issue #2692, opened 2016-12-22
Js error in search page when the background image of any event in the result does not exist


niranjan94:
Eg: https://eventyay.com/explore/world/events/

```
vue.js:523 [Vue warn]: Error when rendering component <event-box>: 
vue.js:2963 Uncaught TypeError: Cannot read property 'split' of null
    at VueComponent.locationName (event-box-component.js:25)
    at Watcher.get (vue.js:1656)
    at Watcher.evaluate (vue.js:1764)
    at Proxy.computedGetter (vue.js:1955)
    at Proxy.eval (eval at makeFunction (vue.js:8238), <anonymous>:2:548)
    at VueComponent.Vue._render (vue.js:2954)
    at VueComponent.<anonymous> (vue.js:2191)
    at Watcher.get (vue.js:1656)
    at new Watcher (vue.js:1648)
    at VueComponent.Vue._mount (vue.js:2190)
```

nikhilrayaprolu:
@niranjan94 can I work on this issue?
and is there any sample event that I could import for reproducing this error?

niranjan94:
No import necessary. Just create a new event yourself.

nikhilrayaprolu:
looks like this is an error when the event has no location name. but I could see that for publishing an event a location name is made  mandatory, hence we wont break this issue. if its really required a condition at https://github.com/fossasia/open-event-orga-server/blob/development/app/static/js/admin/guest/explore/event-box-component.js#L25 may work.

niranjan94:
But that doesn't mean we should be fixing this issue ... Now because of one flawed event, the entire search page doesn't work. We need to ensure that irrespective of what data is coming in (empty/filled) our functionality doesn't break like this ...

PR #2710, opened 2016-12-23
changes for event box component

nikhilrayaprolu:
In reference to #2692 @niranjan94


app/static/js/admin/guest/explore/event-box-component.js
```diff
@@ -22,7 +22,12 @@ Vue.component('event-box', {
             return moment(this.event.start_time).format('ddd, MMM DD HH:mm A');
         },
         locationName: function () {
-            return this.event.location_name.split(",")[0];
+            if(this.event.location_name){
```
SaptakS: I think you should also check for non-empty string. This checks only for the None. Non-empty and undefined should also be checked I think.

app/static/js/admin/guest/explore/event-box-component.js
```diff
@@ -22,7 +22,12 @@ Vue.component('event-box', {
...
+                return this.event.location_name.split(",")[0];
+            }
+            else{
```
SaptakS: Please follow the indentation rule of jslint to put `else` on the same line as closing bracket of `if`. Also provide spaces before opening brackets.

app/static/js/admin/guest/explore/event-box-component.js
```diff
@@ -22,7 +22,12 @@ Vue.component('event-box', {
...
+                return "";
```
SaptakS: I don't think returning a blank string necessary either. Please check.

app/static/js/admin/guest/explore/event-box-component.js
nikhilrayaprolu: split works well with non-empty string,and since we are retrieving from model and we defined the attribute location_name as a String and initialising it to None , there is no choice for it to be returned as undefined. @SaptakS 
screenshot from 2016-12-24 12-22-18

nikhilrayaprolu:
@SaptakS made a new commit

app/static/js/admin/guest/explore/event-box-component.js
```diff
@@ -22,7 +22,9 @@ Vue.component('event-box', {
             return moment(this.event.start_time).format('ddd, MMM DD HH:mm A');
         },
         locationName: function () {
-            return this.event.location_name.split(",")[0];
+            if (this.event.location_name) {
```
SaptakS: Does this check for both `null` and `undefined`. Did you try that?

niranjan94: @nikhilrayaprolu also, we need to return a default value ... Since this is a computed value

PR #2727, opened 2016-12-26
Fix social media section: Save social links to the db

### LINKED ISSUE #2713: Links and Social Media Section Does Not Work
Tried to add External Link/Twitter/Facebook links to the event page, but links do not show up at the right hand side of the event page as before.

shubham-padia:
resolves #2713

### DIFF
--- a/app/helpers/wizard/event.py
+++ b/app/helpers/wizard/event.py
@@ -23,6 +23,7 @@ from app.models.tax import Tax
 from app.models.ticket import Ticket, ticket_tags_table, TicketTag
 from app.models.user import ORGANIZER
 from app.models.users_events_roles import UsersEventsRoles
+from app.models.social_link import SocialLink
 
 
 def get_event_json(event_id):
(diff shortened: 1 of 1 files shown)


app/helpers/wizard/event.py
```diff
@@ -402,3 +405,12 @@ def save_resized_background(background_image_file, event_id, size, image_sizes):
     )
 
     return save_resized_image(background_image_file, width_, height_, basewidth, aspect, height_size, upload_path)
+
+
+def save_social_links(social_links, event):
+
+    for social_link in social_links:
+        social_exists = db.session.query(SocialLink).filter_by(name=social_link['name'], link=social_link['link'], event_id=event.id).scalar()
```
niranjan94: Use the SQLAlachemy ORM syntax that we are using throughout the application like `SocialLink.query.filter_by` .. What you have done is also correct but let's follow the same syntax throughout ...

niranjan94: Do the search only by name. Also make sure the search is case-insensitive. The user may want to change the URL of the social link named 'facebook'. 

Update the existing record else create a new record.

shubham-padia: @niranjan94 Okay :smiley: , made the required changes

Issue #2741, opened 2016-12-27
'External Event URL' field  not shown on event create in public page


magdalenesuo:
If I fill in directly the external event URL it is not shown but if I add a social link and fill it, then it is shown..

filled data:
https://cloud.githubusercontent.com/assets/20799954/21501567/e84a3056-cc6e-11e6-910c-514e6768bb46.png

output:
https://cloud.githubusercontent.com/assets/20799954/21501565/e74a55c8-cc6e-11e6-902b-dc99a58379d1.png

SaptakS:
I think event_url is not getting saved in the Database at all since there is no code in the backend. @magdalenesuo can you add that?

PR #2743, opened 2016-12-27
external event url problem fixed

magdalenesuo:
Issue #2741 fixed
fixed: https://cloud.githubusercontent.com/assets/20799954/21504319/d47d6204-cc83-11e6-875c-2581c3cc124e.png

### DIFF
--- a/app/templates/gentelella/admin/event/wizard/basic-details.html
+++ b/app/templates/gentelella/admin/event/wizard/basic-details.html
@@ -204,7 +204,7 @@
                        value="External Event URL" readonly style="background-color: white;">
             </div>
             <div class="col-sm-5 input-group" v-bind:class="{ 'has-error': !isValidLinkEntry(event.event_url) }">
-                <input type="url" class="form-control" v-model="event.event_url" attribute name="event_url" placeholder="Link">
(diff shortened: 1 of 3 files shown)


app/templates/gentelella/admin/event/wizard/basic-details.html
```diff
@@ -204,7 +204,7 @@
                        value="External Event URL" readonly style="background-color: white;">
             </div>
             <div class="col-sm-5 input-group" v-bind:class="{ 'has-error': !isValidLinkEntry(event.event_url) }">
-                <input type="url" class="form-control" v-model="event.event_url" placeholder="Link">
+                <input type="url" class="form-control" v-model="event.event_url" attribute name="event_url" placeholder="Link">
```
niranjan94: @magdalenesuo could you explain what this `attribute name="event_url"` is for ?

magdalenesuo: @niranjan94 I added the name attribute since I found that all input files had name to it. But it was my mistake to write attribute aswell. Should I remove it?

niranjan94: You don't actually need the name in the wizard. But no problem if it's there either.

PR #2745, opened 2016-12-28
Make sessions page responsive 

shubham-padia:
and also show expanding and contracting navbar only if tracks are greater than max tracks.
You can check it live on https://orga2.herokuapp.com/e/72a1b259/sessions/ 
For comparison on eventyay: https://eventyay.com/e/18252ab6/sessions/


shubham-padia:
@niranjan94  As the navbar was pre-populated with the help of javascript, the above tracks could be seen when the page was loaded ( along with all the images) completely .
Fixed that issue. Please review again.
Also reverted other changes in favour of #2746.

niranjan94:
@shubham-padia the track names still seem to be missing at https://orga2.herokuapp.com/e/72a1b259/sessions/ when viewed from mobile ?

shubham-padia:
@niranjan94 They seem to be missing after #2746,sorry.. did not check the deployed version after updating the branch, working on it.
@SaptakS That is the expected behaviour, the max no. of tracks is fixed, right now = 7. Should we reduce the max no. of tracks to fit them in 1 line?

shubham-padia:
@SaptakS @niranjan94 Can you please review again . I've also removed the extra space on the right of track title. Thanks !!

mariobehling:
Hm. I missed that PR: There should not even be an expand button. It should always show all tracks.

Issue #2763, opened 2016-12-29
Custom Roles: Show only those tabs to which user has access to


shubham-padia:
Parent issue: #2756.
e.g If the user has access to `reports`,`modules` and `messages`, he sees all the other tabs as an option. Show only those tabs to which he has access to.
Thus, in below screenshot only 3 tabs should be displayed i.e `reports`,`modules` and `messages`

![selection_040](https://cloud.githubusercontent.com/assets/13910561/21541264/fc278002-cdda-11e6-9a77-0c1502e0edbf.png)

shubham-padia:
Closed via #2765

PR #2765, opened 2016-12-29
Generate navigation_bar list in the views itself

### LINKED ISSUE #2673: Track number of people going to a particular event.
@the-dagger assign me this. This feature will help users know about the popularity of the session through the number of people going to the event.

shubham-padia:
Fixes #2673 .
Generate `navigation_bar` list in the views itself instead of the template, according to accessible tabs. Previously the list was hardcoded causing #2763.

In this case, the user has access to `reports`,`modules` and `messages`, so only those tabs are being displayed.

After:
selection_041

Before:
selection_040

### DIFF
--- a/app/templates/gentelella/admin/super_admin/dashboard.html
+++ b/app/templates/gentelella/admin/super_admin/dashboard.html
@@ -18,19 +18,6 @@
         }
     </style>
 {% endblock %}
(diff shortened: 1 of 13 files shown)


niranjan94:
@shubham-padia wow ! This is great :)

PR #2776, opened 2016-12-30
Forbid trashing/deleting super_admin from users panel

shubham-padia:
abort with error 403 if someone tries to delete super_admin from the users panel.

### DIFF
--- a/app/views/super_admin/users.py
+++ b/app/views/super_admin/users.py
@@ -88,6 +88,9 @@ def details_view(user_id):
 
 @sadmin_users.route('/<user_id>/trash/', methods=('GET',))
 def trash_view(user_id):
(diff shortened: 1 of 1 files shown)


shubham-padia:
@niranjan94 @SaptakS Before merging, I also noticed that any other admin can edit details of super admin by viewing his/her details in the `details_view` . Should this be allowed, as editing details will only make changes to the data being displayed of the super admin and is not a security bug?

shubham-padia:
@niranjan94 @SaptakS Nevermind the above comment. 
Please review this. Opened a seperate issue for the above comment at #2777 .

niranjan94:
@shubham-padia please include the relevant issue in the PR description

PR #2788, opened 2016-12-30
Wizard session type time: use ":" instead of "."

### LINKED ISSUE #2781: Standard time for session
Standard time displayed for sessions is currently 03.00. It should rather be:
"00:30"

Please use ":" instead of "."

shubham-padia:
fixes #2781 .

### DIFF
--- a/app/static/js/admin/event/wizard/sessions-speakers/vue-components.js
+++ b/app/static/js/admin/event/wizard/sessions-speakers/vue-components.js
@@ -47,7 +47,7 @@ Vue.component('session-type', {
         this.$nextTick(function () {
             var $div = $(this.$el);
             $div.find(".session-length").timepicker({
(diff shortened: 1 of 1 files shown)


mariobehling:
Thanks. In order to fully solve the issue, please help to fix the time usage of 03.00. 

Instead it should be 00:30.

As in my comment, the time changes should also be saved properly. Cheers.

shubham-padia:
@mariobehling Sorry, didn't include the screenshot
screenshot from 2016-12-31 04-14-18

>  the time changes should also be saved properly

The time changes are being saved properly right now

> help to fix the time usage of 03.00.Instead it should be 00:30.

This has also been fixed.

You can even check it on https://orga2.herokuapp.com ( P.S after registering if you get an internal server error, just ignore it and login with the credentials you provided in the registration form )

Issue #2806, opened 2016-12-31
Errors showing ticketing info 


mariobehling:
The ticketing box has the following issues:

* [ ] It shows "the ticket sales will start.." even though the ticket sales start time is in the past
* [ ] It does not show a box to choose multiples tickets

![screenshot from 2016-12-31 22-23-32](https://cloud.githubusercontent.com/assets/1583873/21579296/f338725c-cfa7-11e6-9f75-2af8330e07a2.png)

shubham-padia:
I would like to work on this.

shubham-padia:
Solved via #2822

PR #2818, opened 2017-01-01
Ticket start time according to event timezone

### LINKED ISSUE #2805: Event Ticket time start on local PC time instead of Time of Event location
After the user entered the ticket, the ticket time starting the sales is automatically chosen - as the current time.

The time displayed is the current ticket time of the time zone. If the timezone of the user/PC is different to the timezone of the event, the ticket time still shows the PC timezone.

Expected behavior: Show the timezone of the event and make all ticket sales according to the event timezone.

screenshot from 2016-12-31 22-19-44

shubham-padia:
Fixes #2805 .
Visit http://orga3.herokuapp.com/ for a demo.
Please review.

### DIFF
--- a/app/static/js/admin/event/wizard/vue-app.js
+++ b/app/static/js/admin/event/wizard/vue-app.js
@@ -155,6 +155,9 @@ var app = new Vue({
         },
         'step': function (step) {
             history.replaceState(null, '', "./" + step);
+        },
+        'event.timezone': function (value) {
+            this.event.timezone = value;
         }
     },
     methods: {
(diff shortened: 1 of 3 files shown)


mariobehling:
The commit also seems to change the event start time, which results in "invalid time range".

screenshot from 2017-01-01 23-52-08

mariobehling:
Oh, it seems this bug also exists now on eventyay.com.

mariobehling:
Does not solve the issue.

Also I see that the pre-filled ticket sales time starts at 11pm and ends at 10pm.

The preview states ticket sales "will" start, but the time has passed already.

screenshot from 2017-01-01 23-57-32

shubham-padia:
The issue that you mentioned in the comment is a different issue #2806 which is yet to be solved. 
This pr solves #2805 according to which default start time for the event should be according to the event timezone

shubham-padia:
You can change the event timezone -> add a new ticket -> view start time of the ticket sales which will be according to the event timezone

mariobehling:
Event timezone is Berlin UTC +1. Sales shows "This ticket's sale will start on January 01, 2017 11:50 PM". Time is: January 2, 00:08. Ticket sales should be available, but they are not. Which time is the ticket sales using? The issue is that ticket sales are not available even though they should in the timezone defined.

shubham-padia:
The issue that you are mentioning is #2806 . This PR does not fix that.
This PR only fixes #2805 .

shubham-padia:
I think i'll close this for now and provide changes for both #2806 and #2805

mariobehling:
Ok, I think both need to be fixed together. Thank you for taking care!

shubham-padia:
No problem 😄

PR #2823, opened 2017-01-02
Language fixed

### LINKED ISSUE #2817: Improve language listings
Please change as follows:
* [x] reduce space between languages to show more languages in one space
* [x] For Chinese only show a) Chinese (Simplified) and b) Chinese (Traditional)

magdalenesuo:
Fixed #2817 
space decreased and chinese language updated
screenshot from 2017-01-02 21-53-41


mariobehling:
@SaptakS @niranjan94 Is this inline with the language setup on weblate? It seems weblate cannot be merged right now. Is this related?

SaptakS:
@mariobehling yes. We need to update and merge the weblate repo. Since the 2 chinese languages have been removed, we have to merge it manually fixing the merge conflicts in weblate.

mariobehling:
@SaptakS We should have discussed this before. I would like to avoid that we miss existing translations.

Chinese (Taiwan) = Traditional Chinese
Chinese (China) = Simplified Chinese

The idea of this issue was not to get rid of any versions, but to unify versions.

mariobehling:
It looks like we are indeed loosing translations. Maybe this PR should be reverted and only the footer.css should be committed in a first step.

niranjan94:
@mariobehling ... let this PR stay. Reverting would mess up our files even more. I'll manually fix the weblate repo so that we don't loose translations.

niranjan94:
@mariobehling the weblate repo has been fixed. Lost translations to be restored are here #2828

mariobehling:
Thank you.

PR #2838, opened 2017-01-04
Resize with aspect ratio on fixed

### LINKED ISSUE #2812: Image proportions incorrect on front page
The uploaded images are not scaled and cropped correctly. Instead it appears that they are stretched.

screenshot from 2017-01-01 14-04-38

SaptakS:
#2812

@mariobehling @niranjan this fixes the aspect ratio problem.


mariobehling:
Good job!

Issue #2815, opened 2017-01-01
Discount Codes not working


mariobehling:
* [x] Applying a discount code results in a message "The promotional code entered is valid. No offer has been applied to this order."
* [x] Trying to edit a discount codes shows that parameters have not been saved including
   * [x] Percent
   * [x] All other options
* [x] For "Amount" it is only possible to enter full numbers. Please ensure e.g. amounts like 5.95 USD can be entered.

![screenshot from 2017-01-01 14-30-43](https://cloud.githubusercontent.com/assets/1583873/21581347/42173b56-d02f-11e6-876f-0210ab2b6309.png)

nikhilrayaprolu:
@mariobehling for the third task. I think adding a drop down to select currency type would work?

CosmicCoder96:
@mariobehling I would like to work on this.

mariobehling:
@CosmicCoder96 Let's do this task as first come, first claim, ok? But there are many other tasks. Please feel free to claim any other issue. Thank you.

CosmicCoder96:
@mariobehling Of course.Thank you.

nikhilrayaprolu:
@mariobehling it looks like nothing has been implemented for the discount functionality.
https://github.com/fossasia/open-event-orga-server/blob/development/app/helpers/ticketing.py#L209
I am working on this issue by creating functionality for giving discounts

mariobehling:
@niranjan94 Please provide input, if any. Thank you.

niranjan94:
@mariobehling the PR #2840 looks good to me. But needs fixing since travis is failing

niranjan94:
And yes. Looks like discounts were never added to the code that calculates the order amount. :sweat:

mariobehling:
Thanks. Please test this more in detail. Found some things on a quick look:

* [ ] deleting items in "amount" shows error on "Amount" and "Percentage"
* [ ] ticks or crosses overlap on form field
* [ ] If one field has an error all fields show an error

![screenshot from 2017-01-07 05-21-54](https://cloud.githubusercontent.com/assets/1583873/21739364/7ec773b8-d499-11e6-91d4-748fb20fcaee.png)

* [ ] Cannot see my "max" or "min" number - display is hidden

![screenshot from 2017-01-07 05-22-32](https://cloud.githubusercontent.com/assets/1583873/21739368/8f40292e-d499-11e6-8343-6def970fdbb7.png)

[...]

PR #2840, opened 2017-01-04
discount coupon changes

nikhilrayaprolu:
@mariobehling in reference to fix issue #2815 . fixed all the tasks mentioned over there


nikhilrayaprolu:
@niranjan94 travis passed

PR #2844, opened 2017-01-06
Add kubernetes config for android generator

niranjan94:
Resolves https://github.com/fossasia/open-event-android/issues/945

### DIFF
--- a/kubernetes/images/generators/android/Dockerfile
+++ b/kubernetes/images/generators/android/Dockerfile
@@ -1,0 +1,30 @@
+FROM python:2-slim
+MAINTAINER Niranjan Rajendran <niranjan94@yahoo.com>
+
+ARG COMMIT_HASH
+ARG BRANCH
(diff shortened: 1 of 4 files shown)


mariobehling:
Is it a good idea to include this in this repo? Would it be possible to add another key and do this from the Android repo itself?

niranjan94:
@mariobehling since we had all the kubernetes config here, Thought I'd put it here ... But yes, it is possible to move to the android app repo. I'll do it.

mariobehling:
ok, thanks. As these repos are separated the deployment should be separated for now. (I hope it is the right strategy.)

Please get Harshit on board and give him access to the cloud deployment of the Android app. Thank you.

niranjan94:
Agreed. Let's keep the deployments separate. Since and android app deployments is not gonna affect eventyay deployments in any way.

@mariobehling yes. harshit and myself have been discussing about this

PR #2865, opened 2017-01-07
Store ticket price as float and display accordingly

### LINKED ISSUE #2852: Ticketing: Show currency including Cents on Wizard Step 1
The wizard in step currently shows the ticket rate without cents. There might be events that charge a rate like 5.50 USD. 

Expected: Field also shows Cents.

shubham-padia:
fixes #2852 .

selection_055


mariobehling:
Hi, I think this is a misunderstanding. The ticket display on the user end is good.

The issue is about when the organizer enters prices. There should be a price including cents, e.g. $10.50.

screenshot from 2017-01-07 20-21-32

shubham-padia:
@mariobehling The organiser can enter decimal values as per his/her choice.
selection_056

Or did the issue mean there should be a difference of 50 cents when you increment or decrement the value with the up/down arrow?
Also the ticket price was stored in integer previously, so this fixes that.

mariobehling:
> The organiser can enter decimal values as per his/her choice.

Does not work for me on eventyay using Chrome.

Your last column shows number like $16.5 instead of $16.50. Please check. The current display in the participant front end is fine.

shubham-padia:
Fixed the mentioned issues. Travis also passing. Please review again.

> Does not work for me on eventyay using Chrome.

This should fix that too, also increment/decrement will be in the values of 0.01.

selection_058

selection_057

mariobehling:
Thank you!

Issue #2867, opened 2017-01-08
Invalid argument error in `getpass` method.


gaeun:
<img width="922" alt="screen shot 2017-01-08 at 12 03 18 pm" src="https://cloud.githubusercontent.com/assets/13972732/21746963/8cbde34c-d59a-11e6-9b4a-5b912aa059c7.png">

After running `python create_db.py` there is an error in the getpass method that prevents login as super_admin. I can work on this issue.

gaeun:
Closing via #2868

PR #2868, opened 2017-01-08
Issue #2867: Fixed invalid argument error in getpass method.

gaeun:
The create_db.py command works well now.

screen shot 2017-01-08 at 12 21 56 pm

### DIFF
--- a/create_db.py
+++ b/create_db.py
@@ -1,6 +1,7 @@
 import re
 import sys
-from getpass import getpass
+import getpass
+import eventlet
 
 from app import current_app
 from flask.ext.migrate import stamp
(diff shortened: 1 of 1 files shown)


gaeun:
@niranjan94 Previously, as shown in #2867, the create_db.py script didn't work (an invalid argument error occurred after entering the login email). I used the eventlet patcher (which requirements.txt already imports) to fix this issue. As shown in the screenshot, the email and password arguments are now accepted properly.

niranjan94:
@gaeun yes. I got that. I meant, could you explain what the actual issue was, why it occurs and how your code fixes it ?

gaeun:
I think an error occurs when osx calls on the getpass method, so I used the unpatched os module from eventlet

PR #2875, opened 2017-01-08
Add type as password in smpt password input field

### LINKED ISSUE #2862: hide password while setting the smtp settings
screenshot from 2017-01-07 23-55-02

hiding the password field would be better at /admin/settings/ ,since many users prefer having the password field hided
we could implement a small (show password) button beside the input field for user convenience.

this feature could be extended to other modules for client secrets too

PARAG00991:
Resolves #2862


PARAG00991:
I can also add **Show Password** kind of option if required. Do tell me if required.

PR #2884, opened 2017-01-09
Centered the options in footer menu vertically

### LINKED ISSUE #2877: Footer bar not vertically centered in area
After recent changes the footer bar is not anymore vertically centered. Expected: Imprint and other info is centered in the middle between the two lines above/below.

screenshot from 2017-01-08 23-09-15

PARAG00991:
Fixes #2877 
Screenshot:
footer_correction

### DIFF
--- a/app/static/css/custom/footer.css
+++ b/app/static/css/custom/footer.css
@@ -52,7 +52,7 @@ footer .copyright{
 
 #footer-menu select {background: #282C35; border-color:#AAAAAA; color: #AAAAAA  }
 
-#footer-menu-options{
+#footer-menu > div > div:first-child{
     padding: 10px 10px;
 }
 
--- a/app/static/css/custom/footer.css
+++ b/app/static/css/custom/footer.css
@@ -38,7 +38,6 @@ footer .row {
     height: auto;
     border-top: 1px solid #AAAAAA;
     border-bottom: 1px solid #AAAAAA;
-
 }
 
 #footer-menu select {margin-top: 10px}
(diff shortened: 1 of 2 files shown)


app/static/css/custom/footer.css
```diff
@@ -53,6 +52,10 @@ footer .copyright{
 
 #footer-menu select {background: #282C35; border-color:#AAAAAA; color: #AAAAAA  }
 
+#footer-menu-options{
```
niranjan94: @PARAG00991 Don't create a new ID for just this ... Refer the div from the already existing  `#footer` ID

PARAG00991: @niranjan94 May I refer it using `#footer-menu .row .col-sm-12` ?

niranjan94: Use `#footer-menu > div > div` ... This is more direct. `#footer-menu .row .col-sm-12` would refer to any row inside footer-menu

PARAG00991: @niranjan94 Done as requested!

app/static/css/custom/footer.css
```diff
@@ -52,7 +52,7 @@ footer .copyright{
 
 #footer-menu select {background: #282C35; border-color:#AAAAAA; color: #AAAAAA  }
 
-#footer-menu-options{
+#footer-menu > div > div:first-child{
```
SaptakS: Can you please explain the code? What you did seems to be doing the same thing as before.

SaptakS:
@PARAG00991 please fix the username by which the commit is made. I think it is wrong. Please check and ensure.

PARAG00991:
@SaptakS I guess, I'm using different email-id for making commit to what I use for github account but username is same I think. Would it going to make any difference? If yes, then I'll change it and ensure this thing from next time.

Issue #2850, opened 2017-01-07
Not being redirected to "404 not found" page when an arbitrary URL is entered


arpitn30:
While any arbitrary url is entered after https://eventyay.com/* or http://open-event-dev.herokuapp.com/*, it just shows a blank page instead of being redirected to the 404 error page.

Abhi2424shek:
Can you add some description including the current results, the expected ones and hints on how to reproduce the issue and relevant screen shots or links pls

SaptakS:
@arpitn30 you can work on this if you want.

arpitn30:
Yeah Okay, I'll work on it.

PR #2892, opened 2017-01-09
Fixed redirection to 404 page when arbitrary URL is entered

arpitn30:
Issue #2850
app/views/public/pages.py updated

Added an 'if' condition that returns 404 error message instead of displaying a blank page if the the query filter function cannot find the specified URL in the database.

### DIFF
--- a/app/views/public/pages.py
+++ b/app/views/public/pages.py
@@ -1,5 +1,6 @@
 from flask import Blueprint
 from flask import render_template
+from flask import abort
 
 from app.helpers.data_getter import DataGetter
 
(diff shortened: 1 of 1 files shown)


app/views/public/pages.py
```diff
@@ -10,6 +10,8 @@
 def url_view(url):
     from app import get_locale
     page = DataGetter.get_page_by_url('/' + url, get_locale())
+    if page == None:
+		return render_template('gentelella/errors/404.html'), 404
```
niranjan94: No. Don't directly handle the 404 page like this. Just call the `abort` method with the status code. let flask handle it.

Eg.
`abort(404)`

arpitn30: Done!

Issue #2790, opened 2016-12-31
Sponsorship details not saved or displayed properly


mariobehling:
There are some issues with the sponsorship wizard:

* [x] Going back an forth between wizard steps results in sponsors not being saved
* [x] The order of sponsors sometimes changes after saving them
* [x] The sponsor levels are not made use of - sponsors are not displayed according to their level
* [x] The type information e.g. Gold is not saved in the form
* [x] The description is not displayed in the alt tag and therefore does not show up on scroll over the images

![screenshot from 2016-12-31 13-12-17](https://cloud.githubusercontent.com/assets/1583873/21577448/2516ad54-cf5b-11e6-8498-865da06dd6bc.png)

rohithasrk:
@mariobehling : I'd like to work on this. Can you please help me with the files I'd need to look into?

niranjan94:
@rohithasrk You can work on this :) try browsing through the code base. Start at the event creation wizard.

SaptakS:
> The description is not displayed in the alt tag and therefore does not show up on scroll over the images

@mariobehling do you want me to show the description in form of tooltip? But then description having all the formatting doesn't make sense.

mariobehling:
@SaptakS We are collecting the description with formatting. So, we can later add a "sponsors page", but this is lower priority. Still, the way it is, is good then. Tooltip sounds good for now, but the priority is to get the sorting right for now.

I will open another issue as I have some ideas how to improve this whole section. The first step before we make more substantial changes - is to make the current system work. Right now it does not. So, order and levels...

SaptakS:
@mariobehling have implemented the sorting based on the level. Please check #2872

SaptakS:
> The order of sponsors sometimes changes after saving them

The last edited sponsor gets shown at the last. Though this is a weird phenomenon but now since sorting is done based on level, the ordering inside a single level may change after editing but it will still be sorted based on levels.

PR #2903, opened 2017-01-10
Create Sponsor Fixes

### LINKED ISSUE #2778: Images not displayed from sponsors
screenshot from 2016-12-30 17-04-21

Please compare: https://eventyay.com/e/ecc2001a

SaptakS:
#2778 #2790 

Things done in this PR:

- [x] Fixed Image saving problem
- [x] Fixed ambiguity in order of showing sponsors
- [x] Fixed error in saving data on changing steps in wizard.

@mariobehling this should fix both the issues. I have solved all the issues that I was able to recreate. Please check and let me know if I missed some issues. @niranjan94 review.

### DIFF
--- a/app/api/helpers/custom_fields.py
+++ b/app/api/helpers/custom_fields.py
@@ -26,12 +26,6 @@ class CustomField(Raw):
         format the text in database for output
         works only for GET requests
         """
-        if not self.validate(value):
-            #print 'Validation of field with value \"%s\" (%s) failed' % (
-            #    value, str(self.__class__.__name__))
-            # raise MarshallingError
-            # disabling for development purposes as the server crashes when
-            # exception is raised. can be enabled when the project is mature
         if self.__schema_type__ == 'string':
             return unicode(value)
         else:
--- a/app/api/helpers/custom_fields.py
+++ b/app/api/helpers/custom_fields.py
@@ -27,8 +27,8 @@ class CustomField(Raw):
         works only for GET requests
         """
         if not self.validate(value):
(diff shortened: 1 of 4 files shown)


app/api/helpers/custom_fields.py
```diff
@@ -27,8 +27,8 @@ def format(self, value):
         works only for GET requests
         """
         if not self.validate(value):
-            print 'Validation of field with value \"%s\" (%s) failed' % (
-                value, str(self.__class__.__name__))
+            #print 'Validation of field with value \"%s\" (%s) failed' % (
```
niranjan94: Thanks for this :smile:

app/static/js/admin/event/wizard/vue-app.js
```diff
@@ -390,10 +390,11 @@ function save(stepToSave, state, callback) {
         state: state
     };
 
+    /* Commented out for now until better condition is found
```
niranjan94: :+1:

app/api/helpers/custom_fields.py
niranjan94: Remove this entire if statement

SaptakS: I commented this out because in the comment says "disabling for development purposes as the server crashes when exception is raised. can be enabled when the project is mature". So I thought maybe we could look into this later. If you say I will remove completely.

niranjan94: Travis is failing because of an empty if block. Remove it altogether. We can always add back when needed.

SaptakS:
@niranjan94 check now.

PR #2920, opened 2017-01-11
Use Postgres database for unit testing

### LINKED ISSUE #2918: Use Postgres database for unit testing
Related: #2881 

Currently open event uses SQLite database for testing (both locally and on Travis). This approach has a number of problems. 

1. DateTime field is not natively supported in SQLite. It is just a string in SQLite so testing date-time database fields becomes difficult. (I was writing tests for pentabarf export/import and there I noticed this issue. I remember facing the same issue a lot of times in the past summer but I always used workarounds to get through.)
2. Other advanced datatypes like Color field are not supported in SQLite. 
3. A production server will most probably run on Postgres. So it is in our advantage to test using the same database engine. 

To do this, minor changes will be required in the code and documentation. The changes are working fine so far in my local system.

aviaryan:
Fixes #2918


aviaryan:
@niranjan94 Thanks.

PR #2921, opened 2017-01-11
Flash messages added to admin's content and modules sections

### LINKED ISSUE #2908: No Flash message showing on clicking save button
When admin clicks on the save button after making changes in admin `modules` and **social links** section in admin `content`, flash message doesn't display showing `Changes Saved Successfully` or something like that.

PARAG00991:
Resolves #2908

### DIFF
--- a/app/views/super_admin/content.py
+++ b/app/views/super_admin/content.py
@@ -48,7 +48,7 @@ def index_view():
         for key, value in dic.items():
             settings[key] = value[0]
             set_settings(**settings)
-        flash("Changes Saved Successfully!!!")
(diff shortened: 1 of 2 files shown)


app/views/super_admin/content.py
```diff
@@ -47,6 +48,7 @@ def index_view():
         for key, value in dic.items():
             settings[key] = value[0]
             set_settings(**settings)
+        flash("Changes Saved Successfully!!!")
```
niranjan94: Please drop the exclamation marks. A simple `Changes have been saved.` would do.

PR #2922, opened 2017-01-11
Added tests for pentabarf, fixed issues in pentabarf import/export

aviaryan:
https://github.com/fossasia/open-event-orga-server/issues/2881#issuecomment-271487021

* BUG: No city or venue was being exported in pentabarf. So import of pentabarf always failed as both city and venue were NoneType. (see the diff for fix)
* Fixed more bugs (see importers/pentabarf.xml diff)


aviaryan:
The build fails because of issues in python-pentabarf package. It has been fixed but has not been pushed to pip yet. https://github.com/niranjan94/python-pentabarf-xml/issues/1
@niranjan94 Can you please update the package?

niranjan94:
@aviaryan will update

PR #2958, opened 2017-01-15
BOWER installation

Himanshi-Khandelwal:
some command changes in bower installation.


SaptakS:
@Himanshi-Khandelwal the documentation you have added is purely system dependent and has nothing to do with the documentation of this project. Those steps will differ based on your permission level, which linux distribution you are using and the various system settings. So these changes are in no way needed in our project documentation. Thanks for the contribution though. 

I would request you to look into other issues. There are many issues left to be solved.

Closing this.

PR #2976, opened 2017-01-19
Internal Server should be fixed in admin/sales/events

SaptakS:
Changed from `str` to `unicode`


mariobehling:
@SaptakS Thank you! This PR should also be merged to master quickly. So, we can use the sales area again.

PR #2983, opened 2017-01-19
Discount gets applied to selected tickets only in Public Page

### LINKED ISSUE #2975: Discount in the public event page gets applied to all tickets.
Discount in the public event page gets applied to all tickets instead of getting applied to only those tickets which are applicable for the discount.

magdalenesuo:
Fixed #2975
10% Discount on Ticket2 and vip_hidden:
screenshot from 2017-01-19 20-19-31


mariobehling:
@SaptakS Please also merge this to master as it fixes another important bug.

Hong Phuc would like to send out a link to speakers ticket, but we don't have that feature yet. @magdalenesuo If you are interested, maybe check if you would like to implement that: https://github.com/fossasia/open-event-orga-server/issues/2982

magdalenesuo:
@mariobehling Yes, I'll check it out :)

niranjan94:
@mariobehling master has been updated

PR #3001, opened 2017-01-20
Add Scheduled backups of the database

### LINKED ISSUE #2899: Setup automatic scheduled backups for the eventyay.com database

niranjan94:
Resolves #2899

@mariobehling please review and merge. ~Backups are initiated every 5 days~ Backups are initiated every hour.


mariobehling:
@niranjan94 I would have thought backups should happen several times an hour rather or not? It is possible that people sell a lot of tickets within 5 days and if information gets lost, there would be a big problem. What do you think?

PR #3002, opened 2017-01-20
Enable Proxy protocol in nginx ingress controller and gunicorn

### LINKED ISSUE #2722: Search Function: Location incorrect based on user's IP address
Following up from issue #2714 - the search function does not accurately reflect the country a user is in based on his or her IP address. For instance, I was shown search results for the United States instead of Singapore - although I live in Singapore.

2a5eb670-ca39-11e6-862d-432549c1184a

niranjan94:
Should resolve #2722

@mariobehling

Will require a server downtime of about 20 mins for this to be activated.

### DIFF
--- a/kubernetes/run.sh
+++ b/kubernetes/run.sh
@@ -6,7 +6,7 @@ python manage.py initialize_db -c open_event_test_user@fossasia.org:fossasia
 python manage.py db upgrade > /dev/null 2>&1
 if [ "$DEPLOYMENT" == "web" ]
 then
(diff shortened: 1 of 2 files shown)


mariobehling:
@niranjan94 Thank you.

> Will require a server downtime of about 20 mins for this to be activated.

Then best would be this happens a soon as possible. Could you do it, please?

niranjan94:
@mariobehling will do it now then.

niranjan94:
This PR was not able to resolve #2722. Exploring other options.

PR #3026, opened 2017-01-24
Implement Cancel and Delete order

### LINKED ISSUE #2986: Implement Cancel and Delete order
Currently orders cannot be cancelled or deleted by the organizer. Please implement a cancel and delete action from the ticket dashboard.

* [ ] Cancel means the ticket stays in the dashboard but shows up as canceled and in this case all generated ticket QR codes should become invalid.
* [ ] Delete means all tickets of the order are canceled and the tickets and orders is deleted from the organizer dashboard, but stays in a status "deleted" on the admin dashboard
* [ ] Auto-delete "Deleted Tickets" after 30 days from the admin dashboard.

screenshot from 2017-01-19 17-14-55

shubham-padia:
Resolves #2986 .
selection_096


shubham-padia:
@niranjan94  

>  if an already paid for order is cancelled/deleted, how are the refunds handled ?

AFAIK we do not have the payment details of the user stored, so nothing for that for now. Should removing/cancelling be disabled for completed orders?

niranjan94:
@shubham-padia actually, we do store the unique order reference ID from paypal/stripe for each payment. We can initiate refunds via Stripe/PayPal API. 

@mariobehling do we need refund functionality or can cancel/delete option be disabled for paid orders. ?

shubham-padia:
>  do we need refund functionality or can cancel/delete option be disabled for paid orders. ?

@mariobehling suggestions on what should be done?

mariobehling:
> @mariobehling do we need refund functionality or can cancel/delete option be disabled for paid orders.

For now, disable the delete functionality for these in the system.

> @shubham-padia actually, we do store the unique order reference ID from paypal/stripe for each payment. We can initiate refunds via Stripe/PayPal API.

Yes, this is the way to go. Please open an issue, but tag it "later" until we have a more stable release. For now we can simply "cancel" paid tickets if required and tell the user to do the refund through paypal. We have so many areas to take care of and better follow up on this later, I would say.

shubham-padia:
- > @mariobehling do we need refund functionality or can cancel/delete option be disabled for paid orders.
For now, disable the delete functionality for these in the system.

- > For now we can simply "cancel" paid tickets if required and tell the user to do the refund through paypal

@mariobehling The above two statements seem conflicting, first one says to disable cancelling for paid orders while the second says otherwise. Please suggest further

shubham-padia:
@niranjan94 @mariobehling Please review. following are some screenshots:
If the ticket is paid and status is completed it cannot be deleted
1.)
selection_104

2.)
selection_101

3.)
selection_103

4.)
selection_099

5.)
selection_098

PR #3035, opened 2017-01-27
Orders: pending->expired (3 days) expired->deleted (6 days)

### LINKED ISSUE #2985: Auto delete pending orders after 72 hours
Currently pending orders seem to be kept in the system forever. Please delete pending orders and pending attendees automatically after 72 hours.

screenshot from 2017-01-19 17-08-56

shubham-padia:
resolves #2985 .
- Pending orders are changed to expired after 3 days from the date of creation.
- All expired orders are deleted from admin dashboard after 6 days from the date of creation.
- And the above deleted orders will be permanently deleted after 30 days from the date of deletion (as implemented previously)

Also, for maintainers: the cron job associated to the the task runs from monday to friday, so you may not be able to see the changes in the system over the next weekend :wink:

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -535,8 +535,8 @@ def integrate_socketio():
 
 scheduler = BackgroundScheduler(timezone=utc)
 scheduler.add_job(send_mail_to_expired_orders, 'interval', hours=5)
-scheduler.add_job(empty_trash, 'cron', day_of_week='mon-fri', hour=5, minute=30)
-scheduler.add_job(send_after_event_mail, 'cron', day_of_week='mon-fri', hour=5, minute=30)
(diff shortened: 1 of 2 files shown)


mariobehling:
Thank you!

> Also, for maintainers: the cron job associated to the the task runs from monday to friday, so you may not be able to see the changes in the system over the next weekend

Why not each day? I think, no need to limit it from Monday to Friday.

shubham-padia:
> Why not each day? I think, no need to limit it from Monday to Friday.

It was previously set this way by the devs, so did not change it, must be some reason :sweat_smile: . maybe @niranjan94 @SaptakS can elaborate on it more.

niranjan94:
@shubham-padia make it such that it is not limited to weekdays

shubham-padia:
@niranjan94 @mariobehling done .

PR #3068, opened 2017-01-29
display only published session

### LINKED ISSUE #3054: Only show sessions when schedule has been published
There are issues with the public sessions page:
* [ ] Sessions that have not a time show up at the beginning of the program
* [ ] The session page shows up even though the schedule has not been published

screenshot from 2017-01-28 19-13-05

magdalenesuo:
Fixed #3054
No published session case:
screenshot from 2017-01-29 03-29-57

When Published and is given time:
screenshot from 2017-01-29 23-26-24
screenshot from 2017-01-29 23-26-39

Speaker:
screenshot from 2017-01-30 00-00-23


mariobehling:
> please apply the same rule for the speakers as well. Thanks.

@magdalenesuo @SaptakS This was not part of the issue and should have been discussed in the issue itself first. Actually this is not the desired outcome at the moment. We need the speakers back on the page.

PR #3088, opened 2017-01-30
Show original uploaded photo in case small image not found

SaptakS:


mariobehling:
@SaptakS Which issue is this related to? I am unclear about this.

SaptakS:
@mariobehling this is just a backup plan for showing the speakers. So @niranjan94  has made the changes for speaker photos. But in case speakers have already made changes and was still facing error, this will act as a backup plan to show the original image in case resized image not found.

mariobehling:
Yes, I can see that this is working for the small thumbnails, but they are loading very slowly now.

SaptakS:
@mariobehling this won't happen from next events or anything. There was some bug due to some code commit yesterday due to which some images weren't resized and saved. As a result they weren't shown. So since we cannot or shouldn't ask the speakers to make an entry again or so on, I made this backup. I will check this out and try to manually make the small entries for the speakers who don't have the resized images.

PR #3100, opened 2017-01-31
Image filenames changed and speaker images corrected

### LINKED ISSUE #3099: Speakers images on public event page larger than required and saved non-standard way
The speaker images on the public event page are loading very slowly. It seems the systems calls the original - very large - image sizes. The images also have non-standard names that includes a dot in the file name e.g. 1485768628.36.png. 

* [x] Please avoid file names with a dot
* [x] Implement the automatic generation of the correct image size. In https://eventyay.com/admin/settings/#!images it is marked that the standard should be 150 x 150 px
* [x] Image is not rendered correctly on upload fix this bug and re-generate the relevant images.

screenshot from 2017-01-31 18-09-52

niranjan94:
Resolves #3099

### DIFF
--- a/app/helpers/assets/images.py
+++ b/app/helpers/assets/images.py
@@ -1,10 +1,14 @@
 import os
-import time
+import uuid
 
 import PIL
 from PIL import Image
 from flask import current_app as app
 from app.helpers.storage import upload, UploadedFile
+
+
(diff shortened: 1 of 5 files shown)


hpdang:
Thank you!

PR #3109, opened 2017-02-01
implement ajax upload for session media

### LINKED ISSUE #2801: Add progress upload bar and list uploaded items in Call for Speakers form
The call for speakers form can have upload buttons for slides, video and audio. Currently it does not show that an item is uploaded and it does not show the upload progress. Please implement the following:

* [x] Show upload progress
* [ ] Add "Add More" - Enable upload of several items
* [x] Offer delete functionality
* [x] List uploaded files in the form (currently they are not shown)

screenshot from 2016-12-31 16-56-27

shubham-padia:
fixes point 1,3 and 4 from #2801 
Remove removes the file,e.g if the user accidentally uploaded an audio file, but doesn't want any audio to be there.
selection_111

### DIFF
--- a/app/helpers/data.py
+++ b/app/helpers/data.py
@@ -7,6 +7,7 @@ import traceback
 from datetime import datetime, timedelta
 from os import path
 from urllib2 import urlopen
+from uuid import uuid4
 
 import PIL
 import oauth2
@@ -17,7 +18,7 @@ from flask.ext.scrypt import generate_password_hash, generate_random_salt
 from flask_socketio import emit
 from requests_oauthlib import OAuth2Session
 
-from app.helpers.assets.images import get_image_file_name
+from app.helpers.assets.images import get_image_file_name, get_path_of_temp_url
 from app.helpers.cache import cache
 from app.helpers.helpers import string_empty, string_not_empty
 from app.helpers.notification_email_triggers import trigger_new_session_notifications, \
@@ -202,9 +203,21 @@ class DataManager(object):
         :param event_id: Session belongs to Event by event id
         """
         form = request.form
-        slide_file = DataManager.get_files_from_request(request, 'slides')
-        video_file = DataManager.get_files_from_request(request, 'video')
-        audio_file = DataManager.get_files_from_request(request, 'audio')
+        slide_temp_url = form.get('slides_url')
+        video_temp_url = form.get('video_url')
+        audio_temp_url = form.get('audio_url')
+        slide_file = ''
+        video_file = ''
+        audio_file = ''
+
+        if slide_temp_url:
+            slide_file = UploadedFile(get_path_of_temp_url(slide_temp_url), slide_temp_url.rsplit('/', 1)[1])
+
+        if video_temp_url:
+            video_file = UploadedFile(get_path_of_temp_url(video_temp_url), video_temp_url.rsplit('/', 1)[1])
+
+        if audio_temp_url:
+            audio_file = UploadedFile(get_path_of_temp_url(audio_temp_url), audio_temp_url.rsplit('/', 1)[1])
 
         if not state:
             state = form.get('state', 'draft')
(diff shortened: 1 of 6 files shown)


mariobehling:
Thanks. Please check travis.

app/helpers/data.py
```diff
@@ -287,6 +300,12 @@ def add_session_to_event(request, event_id, state=None, use_current_user=True):
                     Helper.send_notif_invite_papers(user, event.name, cfs_link, link)
 
     @staticmethod
+    def add_session_media(request, media):
+        file = DataManager.get_files_from_request(request, media)
```
SaptakS: Please change the variable name to avoid this error

mariobehling:
@SaptakS How does it look like once the file has been uploaded? Can we follow the way gmail does it?

Show file name and progress bar and once it is uploaded a "X" to delete it? This would also give the possibility to upload several files.

What do you think?

SaptakS:
> Show file name and progress bar and once it is uploaded a "X" to delete it? 

@mariobehling that sounds great. 

> This would also give the possibility to upload several files.

I am not sure this is required because after all for a particular session I would think there will be only one slide. However this can be a good enhancement. Will require some modification in backend to allow multiple files. Right now it allows a single file.

mariobehling:
> Right now it allows a single file.

Ok, then let's do several files in another issue. It would be useful, as some sessions e.g. panels could have 5 people where each of the panelist has own slides.

shubham-padia:
@SaptakS @mariobehling 
I've changed it to something like this, is this what you meant?
selection_112

> Ok, then let's do several files in another issue. It would be useful, as some sessions e.g. panels could have 5 people where each of the panelist has own slides.

 alternatively zip files can be uploaded for now :)

SaptakS:
@shubham-padia put the cross button at the end of the blue progress bar (I am guessing that is a progress bar) and I don't think download button is needed.

shubham-padia:
>  I don't think download button is needed.

This is to view the current files uploaded when editing the session, so i think it should be there

mariobehling:
* I think no need to make it full width.
*  Also align text left.
* Yes, download button is not necessary. Just follow gmail and let the user download by clicking on the file name.
* Move "Add File" below item.

mariobehling:
Compare to Gmail:
screenshot from 2017-02-01 12-47-00

app/static/js/events/session/new.js
```diff
@@ -18,3 +18,55 @@ $(document).ready(function () {
...
+    $('#progress-' + name).hide();
+};
+
+var fileUpload = function(name, media){
+    var fd = new FormData();
+    fd.append(name, media.files[0]);
+    filename = media.files[0].name;
+    $.ajax({
+        xhr: function(){
+            var xhr = new window.XMLHttpRequest();
```
sudheesh001: I don't understand why there is a mixed usage of `jquery` and pure `javascript` over here. Why has this approach been taken to do this ?

niranjan94: @sudheesh001 because afaik, in jQuery's ajax implementation, the only way to get the progress is to access the underlying XHR instance.

shubham-padia: agree with @niranjan94 :+1:

shubham-padia:
@mariobehling @SaptakS 
selection_114
Is this the desired output, if its ok, ill make the required changes and push it.

mariobehling:
> Is this the desired output, if its ok, ill make the required changes and push it.

Yes, if the file is clickable and downloadable when it is finished.

shubham-padia:
Please review at http://orga3.herokuapp.com/e/64a65ca2/ once. Ready to merge

app/helpers/data.py
```diff
@@ -202,9 +203,24 @@ def add_session_to_event(request, event_id, state=None, use_current_user=True):
...
-        audio_file = DataManager.get_files_from_request(request, 'audio')
+        slide_temp_url = form.get('slides_url')
+        video_temp_url = form.get('video_url')
+        audio_temp_url = form.get('audio_url')
+        slide_file = ''
+        video_file = ''
+        audio_file = ''
+
+        if slide_temp_url:
+            slide_file = UploadedFile('{}/static/{}'.format(path.realpath('.'), slide_temp_url[len('/serve_static/'):]),
```
niranjan94: @shubham-padia instead of this, could you use the `get_path_of_temp_url` method from `app/helpers/assets/images.py` ?

shubham-padia: @niranjan94 Thanks :)

shubham-padia: Resolved

PR #3129, opened 2017-02-02
Preserve session state on edit

### LINKED ISSUE #3059: Session Overview with scrollbars, long text, draft issue
The session overview has a few issues:
* [ ] shows scroll bars after saving sessions - Expected: No scrollbars
* [x] After user adds more information to an accepted session, the session shows up again in draft mode - Expected: Session should stay in same mode as before
* [ ] The abstract text in overview is very long - Expected: Only show the first 200 characters of the abstract

screenshot from 2017-01-29 04-36-14

### LINKED ISSUE #3094: Proposal Stage changing from pending to accepted after editing content
Editing session proposals as an organizer results in sessions showing up in the "Accepted Filter" on the speakers dashboard, but at the same time they are not promoted to the public event page.

Expected: Editing speakers proposals should not change their state.

screenshot from 2017-01-30 22-30-23

### LINKED ISSUE #3124: If speaker or organizer edits a session it goes back to Draft Mode, but it should not
If speaker or organizer edits a session it goes back to Draft Mode, but it should not. Expected: It should stay in the mode that it was already. Probably the save button is using the same function as "save draft" instead of "submit".

The result of the draft mode is, that accepted talks do not show up anymore on the live page.

E.g. check at https://eventyay.com/events/6/sessions/

niranjan94:
Resolves #3124
Resolves #3094
Partially resolves #3059 (Point 2)

### DIFF
--- a/app/helpers/data.py
+++ b/app/helpers/data.py
@@ -351,29 +351,29 @@ class DataManager(object):
             video_file = ''
             audio_file = ''
 
-            if slide_temp_url and slide_temp_url!=session.slides:
+            if slide_temp_url and slide_temp_url != session.slides:
                 slide_file = UploadedFile(get_path_of_temp_url(slide_temp_url), slide_temp_url.rsplit('/', 1)[1])
 
-            if video_temp_url and video_temp_url!=session.video:
+            if video_temp_url and video_temp_url != session.video:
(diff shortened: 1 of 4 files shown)


SaptakS:
@niranjan94 can you check whether this also fixes #3094 and second point of #3059 ?

niranjan94:
@SaptakS yes. Description has been updated

PR #3135, opened 2017-02-02
issue #3046 Enable applying of access code and discount code on event sa…

### LINKED ISSUE #3046: Enable to use access code and discount code together
If the user generates an access code, it is not possible to apply a discount code as well. Please implement as follows:
* [ ] Allow to enter a discount code on a page that has an access code applied
* [ ] Allow the combination of access + discount code links (manually for now) e.g. `/e/47582707/?acode=VIP&dcode=888`

Jatinbalodhi96:
Resolves #3046

now user can apply discount code on the page where access code is already applied,
`/e/47582707/?access_code=VIP&discount_code=888`
In this commit access code is separated as `access_code` for Access Code and Discount Code as `discount_code`,
while generating a new discount code and access code user will get a different links
`localhost:5000/e/47582707/?access_code=VIP` and
`localhost:5000/e/47582707/?discount_code=888`

**with this commit user can only be able to generate new discount and access code link but not the combination of both access and discount code, (manually entering link for now)**


app/templates/gentelella/guest/event/_ticketing_box.html
```diff
@@ -270,9 +270,9 @@ <h3 class="panel-title">{{ _("Ticket Information") }}</h3>
         var discount_amount = -1;
         var promo_code_status = 0;
         var discount_code = "";
-        {% if code %}
+        {% if discount_code %}
```
SaptakS: Tell me this. Where are you checking for access_code in this page? I don't see any access_code related code here. So how is access_code working. Please explain.

Jatinbalodhi96: no I'm not, the issue was that both access code and discount code use the same query parameter `code` so when user uses access code link he cannot use discount code, The issue is to separate them and we can use access code in future commits.

niranjan94: Please do both in this same PR

Jatinbalodhi96: @niranjan94 okay, but i need some help about how it's going to work? so i need to discuss with someone, if you can help?

niranjan94: @Jatinbalodhi96 I think it's pretty clear in #3046

PR #3146, opened 2017-02-04
Fixes #3092: Adds a confirm dialog before deleting event

### LINKED ISSUE #3092: Event Dashboard: Add security step before deleting event
The event dashboard offers the functionality to delete an event in the last row of the table. As some events hold a lot of data this step is relatively "dangerous" for the user. Therefore please implement the deletion process as follows.

When the user clicks the delection button please open a box with the following
   * [ ] Add the question: "Are you sure you want to delete this event?"
   * [ ] "Please type the name of the event into the below box" (Offer a text box here)
   * [ ] Show two buttons a) "Cancel" and b) "Delete Event (this cannot be undone)"
   * [ ] Before user has typed in the event name show the deletion button somehow grayed and inactive
   * [ ] After the user has typed in the event name correctly show the deletion button in red and active

This flow is similar to when a user deletes a repository in Github. Please compare the flow there.

CosmicCoder96:
Fixes #3092 
Here is a screenshot: 
image

### DIFF
--- a/app/static/js/events/event-base.js
+++ b/app/static/js/events/event-base.js
@@ -1,0 +1,15 @@
+var deleteButton = $('#delete-event-button');
+var deleteCheck = $('#delete-check');
+var eventName = deleteCheck.attr('data-event-name');
+
+deleteButton.addClass('disabled');
+deleteCheck.keyup(function() {
+     if(deleteCheck.val() === eventName) {
+            deleteButton.removeClass('disabled');
+     }
+     else {
+        if(!deleteButton.hasClass('disabled')) {
+        	deleteButton.addClass('disabled');
+        }
+     }
+});
--- a/app/templates/gentelella/users/events/event_base.html
+++ b/app/templates/gentelella/users/events/event_base.html
@@ -107,11 +107,13 @@
                     <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span
                             aria-hidden="true">&times;</span></button>
                     <h4 class="modal-title">{{ _("Are you sure you would like to delete this event") }} ?</h4>
+                    <span>{{ _("Deleting the event will delete all the data associated with it.") }} <br><br>{{ _("Please enter the full name of the event to continue") }}</span><br>
+                    <input id='delete-check' type="text" data-event-name="{{ event.name }}">
                 </div>
                 <div class="modal-footer">
                     <a type="button" class="btn btn-default" data-dismiss="modal">{{ _("Cancel") }}</a>
                     <a href="/events/{{ event.id }}/trash/" type="button"
-                       class="btn btn-danger">{{ _("Confirm Delete") }}</a>
+                       class="btn btn-danger" id = "delete-event-button">{{ _("Confirm Delete") }}</a>
                 </div>
             </div>
         </div>
@@ -119,4 +121,5 @@
 {% endblock %}
 {% block tail_js %}
     {{ super() }}
+    <script type = "text/javascript" src = "{{ url_for('static', filename='js/events/event-base.js') }}"></script>
 {% endblock %}


app/templates/gentelella/users/events/event_base.html
```diff
@@ -107,16 +107,34 @@
...
+        $('#delete-button').addClass('disabled');
+        $('#delete-check').keyup(function(){
+            var x = $('#delete-check').val();
+            if(x == '{{ event.name }}')
+            {
+                $('#delete-button').removeClass('disabled');
+            }
+            else
+            {
+                if(!$('#delete-button').hasClass('disabled'))
```
niranjan94: You're using `$('#delete-button')` multiple times. Which is inefficient. Store it in a variable and refer from that.

app/templates/gentelella/users/events/event_base.html
```diff
@@ -107,16 +107,34 @@
...
         </div>
     </div>
 {% endblock %}
 {% block tail_js %}
     {{ super() }}
+    <script type = "text/javascript">
+        $('#delete-button').addClass('disabled');
+        $('#delete-check').keyup(function(){
+            var x = $('#delete-check').val();
+            if(x == '{{ event.name }}')
```
niranjan94: Use `===`

app/templates/gentelella/users/events/event_base.html
niranjan94: It's a button .. So use the disabled prop ... Don't set the class

CosmicCoder96: @niranjan94 The delete button is actually an anchor tag with the type button, and hence it doesn't support the disabled property like an original button. That's why I went with using the default bootstrap disabled class.

CosmicCoder96:
@niranjan94 Please review again.

CosmicCoder96:
@niranjan94 I have created a separate js file.

app/templates/gentelella/users/events/event_base.html
```diff
@@ -107,16 +107,19 @@
                     <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span
                             aria-hidden="true">&times;</span></button>
                     <h4 class="modal-title">{{ _("Are you sure you would like to delete this event") }} ?</h4>
+                    <span>{{ _("Deleting the event will delete all the data associated with it.") }} <br><br>{{ _("Please enter the full name of the event to continue") }}</span><br>
+                    <input id = 'delete-check' type = "text"/ data-eventname = "{{ event.name }}">
```
niranjan94: Let's give proper names. For attributes separate multiple words with a hyphen. So `data-eventname` would become `data-event-name`

CosmicCoder96:
@niranjan94 Thanks for the review. I have corrected the name.

app/templates/gentelella/users/events/event_base.html
```diff
@@ -107,16 +107,19 @@
                     <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span
                             aria-hidden="true">&times;</span></button>
                     <h4 class="modal-title">{{ _("Are you sure you would like to delete this event") }} ?</h4>
+                    <span>{{ _("Deleting the event will delete all the data associated with it.") }} <br><br>{{ _("Please enter the full name of the event to continue") }}</span><br>
+                    <input id = 'delete-check' type = "text"/ data-event-name = "{{ event.name }}">
```
niranjan94: There seems to be an extra slash here. 
And for html attributes, no need spacing around equals symbol.

niranjan94: There seems to be an extra slash here.
And for html attributes, no need spacing around equals symbol.

app/static/js/events/event-base.js
```diff
@@ -0,0 +1,16 @@
+var deleteButton = $('#delete-event-button');
+var deleteCheck = $('#delete-check');
+var eventName = deleteCheck.attr('data-event-name');
+
+deleteButton.addClass('disabled');
+deleteCheck.keyup(function(){
+     if(deleteCheck.val() === eventName)
```
niranjan94: Please follow a uniform code style. The bracket should start in the same line as the if. And make sure the ending bracket has the same indentation as if.

CosmicCoder96:
@niranjan94 I have made the required changes.

Issue #3110, opened 2017-02-01
Admin Panel: Show detailed user numbers


mariobehling:
In the admin panel there is a widget with a table "System Roles". Please improve as follows:

* [ ] Add Row "Unverified User"
* [ ] On the last row show "Total No. of Users"

![screenshot from 2017-02-01 11-25-18](https://cloud.githubusercontent.com/assets/1583873/22503343/99f8450e-e871-11e6-8f22-59ba0537ea2c.png)

Himanshi-Khandelwal:
I am on it :)

Himanshi-Khandelwal:
@mariobehling  @SaptakS  @niranjan94 who can be called as an unverified user? 
(out of number_live_events , number_draft_events , number_past_events , organizers ,  co_organizers ,  track_organizers ,   attendees , accepted_sessions ,  rejected_sessions.)

or some other field is to be created?

Himanshi-Khandelwal:
@mariobehling  @SaptakS  @niranjan94  difficulty in finding  unverified user count .Please help

Himanshi-Khandelwal:
@mariobehling @niranjan94 @SaptakS  issue solved !! 
![screenshot from 2017-02-03 12 24 25](https://cloud.githubusercontent.com/assets/17119462/22582289/f410c958-ea0b-11e6-9a4b-9cc51965a874.png)

Abhi2424shek:
@Himanshi-Khandelwal PR? #?

Himanshi-Khandelwal:
this issue is solved !! and the PR is merged ..

PR #3148, opened 2017-02-04
unverified users and total users added #3110

Himanshi-Khandelwal:
Rows of unverified users and total number of users added in the table.

### DIFF
--- a/app/views/super_admin/super_admin.py
+++ b/app/views/super_admin/super_admin.py
@@ -27,6 +27,7 @@ def index_view():
     super_admins = DataGetter.get_all_super_admins()
     admins = DataGetter.get_all_admins()
     registered_users = DataGetter.get_all_registered_users()
+    unverified_users = DataGetter.get_all_registered_users()
     organizers = get_count(DataGetter.get_all_user_roles(ORGANIZER))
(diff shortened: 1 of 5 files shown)


app/views/super_admin/super_admin.py
```diff
@@ -27,6 +27,7 @@ def index_view():
     super_admins = DataGetter.get_all_super_admins()
     admins = DataGetter.get_all_admins()
     registered_users = DataGetter.get_all_registered_users()
+    unverified_users = DataGetter.get_all_registered_users()
```
magdalenesuo: Shouldn't this be get_all_unverified_users()?

app/views/super_admin/super_admin.py
```diff
@@ -35,6 +36,7 @@ def index_view():
     rejected_sessions = get_count(DataGetter.get_all_rejected_sessions())
     draft_sessions = get_count(DataGetter.get_all_draft_sessions())
     email_times = DataGetter.get_email_by_times()
+    total_users= super_admins + admins + registered_users + unverified_users
```
niranjan94: This would give a wrong count ... An admin by default is a registered user ... So, the count would be more than actual

PR #3150, opened 2017-02-05
unverified and total number of users #3110

Himanshi-Khandelwal:
Rows of unverified and total number of users added.


app/views/super_admin/super_admin.py
```diff
@@ -35,6 +36,7 @@ def index_view():
     rejected_sessions = get_count(DataGetter.get_all_rejected_sessions())
     draft_sessions = get_count(DataGetter.get_all_draft_sessions())
     email_times = DataGetter.get_email_by_times()
+    total_users= super_admins + admins + registered_users + unverified_users
```
niranjan94: This will give a wrong count. Both super_admin and admins would come under either registered or unverified ...

Himanshi-Khandelwal: then total users= registered_users+ unverified_users ?

niranjan94: Yes.

Himanshi-Khandelwal: ok also while squashing the commits facing problem

PR #3160, opened 2017-02-07
unverified and total users#3110

Himanshi-Khandelwal:
Rows for unverified and total number of users added.

### DIFF
--- a/app/views/super_admin/super_admin.py
+++ b/app/views/super_admin/super_admin.py
@@ -27,6 +27,7 @@ def index_view():
     super_admins = DataGetter.get_all_super_admins()
     admins = DataGetter.get_all_admins()
     registered_users = DataGetter.get_all_registered_users()
+    unverified_users = DataGetter.get_all_unverified_users()
     organizers = get_count(DataGetter.get_all_user_roles(ORGANIZER))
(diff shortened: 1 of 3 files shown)


app/views/super_admin/super_admin.py
```diff
@@ -35,6 +36,7 @@ def index_view():
     rejected_sessions = get_count(DataGetter.get_all_rejected_sessions())
     draft_sessions = get_count(DataGetter.get_all_draft_sessions())
     email_times = DataGetter.get_email_by_times()
+    total_users=registered_users + unverified_users
```
niranjan94: Instead of doing the total here, storing it in a variable and then passing to the template and referring it there, it will be easier/simpler to do the total directly in the template

niranjan94: @Himanshi-Khandelwal

Himanshi-Khandelwal: @niranjan94  for this I need to make a function for doing total right ?  and is this a wrong way to do ?

niranjan94: Why do you need a function ? Just do it directly in the template 

Instead of 

`<td>{{ total_users }}</td>`

Put 

`<td>{{ registered_users + unverified_users }}</td>`

Issue #3161, opened 2017-02-07
Simplify "Add Microlocations" in Scheduler


mariobehling:
There are a number of fields in the "Add Microlocations.
![screenshot from 2017-02-07 10-53-00](https://cloud.githubusercontent.com/assets/1583873/22689324/b161e808-ed30-11e6-96b5-56e5b8350859.png)

Please simplify this as follows:

* [ ] Change "Name" to "Room Name"
* [ ] Move "Floor" to second position and make it optional
* [ ] Make Longitude/Latitude optional
* [ ] The bottom option for "Room" does not appear anywhere and therefore can be taken out apparently

CosmicCoder96:
@mariobehling I would like to work on this.

niranjan94:
You can work on this once you're done with #3117

PR #3162, opened 2017-02-07
simplified the form as per #3161

dristybutola:
Made changes in scheduler.html .Would be checking for other changes further #3161


niranjan94:
@dristybutola please always add a comment on the issue you'd like to work on. This issue has already been solved in another PR.

PR #3164, opened 2017-02-07
name variable to room_name

dristybutola:
#3161 changed name variable to room_variable, added optional requirement .


niranjan94:
No database change is required. All the issue needs is for the label to change from `Name` to `Room name`. Leave the database untouched

Issue #3078, opened 2017-01-30
Autofill option for ticket buyer 


hpdang:
if I want to buy a ticket for myself only, I still have to fill in my name and email twice. Is it possible to implement an autofill mode? 

![autorefill](https://cloud.githubusercontent.com/assets/1658742/22419249/dd24d41e-e6dc-11e6-8a10-4ddfcdc6e851.png)

CosmicCoder96:
@hpdang I would like to work on this.

PR #3165, opened 2017-02-07
Issue #3078 :  Adds option of auto filling

CosmicCoder96:
The original IDs for various elements were repeated, hence they have been changed to distinguish individual text fields.
There may be multiple ticket types and the buyer might want to autofill anyone of them.
If the buyer choses any particular ticket type, the rest of the checkboxes automatically get disabled
The autofill box appears only once for a particular ticket type.

#3078

Preview: 
image

@niranjan94 
Please review.

### DIFF
--- a/app/static/js/guest/ticketing.js
+++ b/app/static/js/guest/ticketing.js
@@ -173,3 +173,25 @@ $payViaStripe.on('click', function (e) {
 $(window).on('popstate', function () {
     handler.close();
 });
+
+var checked = false;
+var checkboxes =  $('.autofill-checkbox');
+checkboxes.click(function(){
+        var firstName = $('#firstname').val();
+        var lastname =  $('#lastname').val();
+        var email =     $('#email').val();
+        var requiredId = this.id;
+        checkboxes.prop('disabled', true);
+        $(this).prop('disabled', false);
+        if($(this).is(':checked')) {
+            $('#' + requiredId + '-holders\\[firstname\\]').val(firstName);
+            $('#' + requiredId + '-holders\\[lastname\\]').val(lastname);
+            $('#' + requiredId + '-holders\\[email\\]').val(email);
+        }
+        else {
+            $('#' + requiredId + '-holders\\[firstname\\]').val("");
+            $('#' + requiredId + '-holders\\[lastname\\]').val("");
+            $('#' + requiredId + '-holders\\[email\\]').val("");
+            checkboxes.prop('disabled', false );
+        }
+});
--- a/app/templates/gentelella/guest/ticketing/order_pre_payment.html
+++ b/app/templates/gentelella/guest/ticketing/order_pre_payment.html
@@ -1,4 +1,4 @@
-{% extends 'gentelella/guest/ticketing/order_base.html' %}
+        {% extends 'gentelella/guest/ticketing/order_base.html' %}
 
 {% block title %}{{ _("Your Order") }}{% endblock %}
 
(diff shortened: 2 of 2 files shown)


niranjan94:
@CosmicCoder96 which issue is this related to ?

CosmicCoder96:
@niranjan94 issue#3078

app/static/js/vendor/jquery/jquery.checkboxes-1.2.0.js
```diff
@@ -0,0 +1,212 @@
+'use strict';
```
niranjan94: This library is completely necessary. Please remove it. Stick to simple checkboxes.

app/static/js/tickets/prepay-tickets.js
```diff
@@ -0,0 +1,19 @@
+$('.autofill-checkbox').click(function(){
```
niranjan94: There is already a js file for this page `js/guest/ticketing.js`. put the code in that.

app/static/js/tickets/prepay-tickets.js
```diff
@@ -0,0 +1,19 @@
...
+        var firstName = $('#firstname').val();
+        var lastname =  $('#lastname').val();
+        var email =     $('#email').val();
+        var requiredId = $(this).data('ticket-type') + '-' + this.id ;
+        if($(this).is(':checked')) {
+            $(document.getElementById(requiredId + '-holders[firstname]')).val(firstName);
```
niranjan94: We use jQuery just we don't have to do `document.getElementById` 

You can just get the element by doing `$('#' + requiredId + '-holders[firstname]')`. As simple as that.

app/static/js/vendor/jquery/jquery.checkboxes-1.2.0.js
CosmicCoder96: @niranjan94 That library takes care of disabling users from selecting multiple checkboxes pretty neatly. Hence was added,  is it strictly necessary to exclude it?

app/static/js/tickets/prepay-tickets.js
CosmicCoder96: @niranjan94 document.getElementById was deliberately used instead of jquery selector because of the square brackets in the id of the input fields. Since they have a special meaning to jquery selectors, it can't be used. I tried escaping them with \\ but it didn't work either. Hence I resorted to DOM selector. It's a bad practice to use square braces in the ID of the elements if we are using jquery, but I didn't change the id as it might have had implications in other files as well. What course do you recommend , to continue with it or change the ID?

app/static/js/vendor/jquery/jquery.checkboxes-1.2.0.js
niranjan94: Yes. Exclude it. We already have too many libraries for even simple things ... Disabled cursor over hovering can even be done with simple jQuery/css

app/static/js/tickets/prepay-tickets.js
niranjan94: I didn't notice the brackets before. But even that can be easily escaped.

CosmicCoder96:
@niranjan94 Please review again , I have made the required changes :)

app/static/js/guest/ticketing.js
```diff
@@ -173,3 +173,25 @@ $payViaStripe.on('click', function (e) {
 $(window).on('popstate', function () {
     handler.close();
 });
+
+var checked = false;
+var checkboxes =  $('.autofill-checkbox')
```
niranjan94: @CosmicCoder96 please fix this

CosmicCoder96:
@SaptakS I have made the changes.

PR #3169, opened 2017-02-09
Fixes #3117 : timezone issue fixed

### LINKED ISSUE #3117: Call for Speakers not open even though time to open is specified
The call for papers does not open according to the time specified. Possibly it uses the server timezone instead of the event timezone.

screenshot from 2017-02-01 15-45-26

CosmicCoder96:
The call for speaker now opens at the correct time , after fixing the timezone difference.


SaptakS:
@CosmicCoder96 please always mention the issue link which is being solved by the PR. Thanks.

PR #3170, opened 2017-02-09
Scheduler search fixed

### LINKED ISSUE #3158: Search not working for Scheduler
Search in scheduler does not work.

screenshot from 2017-02-07 10-40-01

magdalenesuo:
Fixed #3158
screenshot from 2017-02-09 21-23-33
screenshot from 2017-02-09 21-23-40

### DIFF
--- a/app/static/js/vendor/jquery/jquery.codezero.js
+++ b/app/static/js/vendor/jquery/jquery.codezero.js
@@ -150,7 +150,7 @@ var cachedFuzzyMatch;
  * Cache the results of a RegExp match.
  * @type {Function}
  */
-if ('undefined' === !typeof _) {
+if ('undefined' !== typeof _) {
     cachedFuzzyMatch = _.memoize(function (pattern) {
         return new RegExp(pattern.split("").reduce(function (a, b) {
             return a + '[^' + b + ']*' + b;


app/static/js/vendor/jquery/jquery.codezero.js
```diff
@@ -150,13 +150,11 @@ var cachedFuzzyMatch;
...
  * @type {Function}
  */
-if ('undefined' === !typeof _) {
-    cachedFuzzyMatch = _.memoize(function (pattern) {
-        return new RegExp(pattern.split("").reduce(function (a, b) {
-            return a + '[^' + b + ']*' + b;
-        }), 'i');
-    });
-}
+cachedFuzzyMatch = _.memoize(function (pattern) {
```
niranjan94: @magdalenesuo the if statement was intentional ... `jquery.codezero.js` is included in some pages where lodash is not included. So in those pages, an error will occur if we try to use a lodash function without the lodash library.

magdalenesuo: @niranjan94 That IF function was always returning false though.

niranjan94: Huh .. that's weird ... I was the one who had implemented this long ago ... Could you check why ?

niranjan94: @magdalenesuo change it to 
`'undefined' !== typeof _`

PR #3174, opened 2017-02-10
Fixes #3113: Adds from name to email

### LINKED ISSUE #3113: Specify from "Name" for emails in admin dashboard
The system emails that are sent out from eventyay currently show "info" as the name. Please implement a field in the email settings to specify a "Name" so when people get an email from the system this name shows up in their inbox.

screenshot from 2017-02-01 12-49-24

CosmicCoder96:
Fixes #3113 

Specifies from "Name" for emails in admin dashboard

image

### DIFF
--- a/app/helpers/helpers.py
+++ b/app/helpers/helpers.py
@@ -391,7 +391,11 @@ def send_email(to, action, subject, html):
     """
     if not string_empty(to):
         email_service = get_settings()['email_service']
-        email_from = get_settings()['email_from']
+        email_from_name = get_settings()['email_from_name']
+        if email_service == 'smtp':
+            email_from = email_from_name + '<' + get_settings()['email_from'] + '>'
+        else:
+            email_from = get_settings()['email_from']
         payload = {
             'to': to,
             'from': email_from,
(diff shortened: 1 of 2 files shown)


niranjan94:
And @CosmicCoder96 ... please comment on the issue _before_ you start working on it so that we can ensure multiple people are not working on the same issue

CosmicCoder96:
@niranjan94
Thanks for the review. I will be sure to keep that in mind. 
I have now placed the from name text field on the upper part of the form to make it common to both. And since the payload in helpers.py is common to both smtp and sendgrid, I believe this should suffice. Please review again.

niranjan94:
@CosmicCoder96 no. this would fail for sendgrid. Sendgrid expects only an email ID in the `from` field. For sendgrid you'd need to put the name in the `fromname` field. 

Please see https://sendgrid.com/docs/API_Reference/Web_API/mail.html

niranjan94:
When using a third-party API/sdk, never assume a behavior. Always check the appropriate documentation and confirm.

CosmicCoder96:
@niranjan94
Thanks for the advice. I have made the necessary changes.

PR #3176, opened 2017-02-10
Shows all dates in scheduler

### LINKED ISSUE #3159: Scheduler does not show option for different event dates
When event has several dates only the first day is displayed in the scheduler.

Expected: There should be buttons on top for the different event days

screenshot from 2017-02-07 10-45-58

SaptakS:
Fixes #3159 @niranjan94 review

screenshot from 2017-02-12 03-00-50
screenshot from 2017-02-12 03-00-54
screenshot from 2017-02-12 03-00-56

### DIFF
--- a/app/static/js/events/scheduler.js
+++ b/app/static/js/events/scheduler.js
@@ -784,11 +784,11 @@ function processMicrolocationSession(microlocations, sessions, callback) {
                 sessionsStore[dayIndex] = [session];
             }
 
-            for (var dayIndex in days) {
-                if (_.isArray(unscheduledStore[dayIndex])) {
-                    unscheduledStore[dayIndex].push(session);
+            for (var index in days) {
+                if (_.isArray(unscheduledStore[index])) {
+                    unscheduledStore[index].push(session);
                 } else {
-                    unscheduledStore[dayIndex] = [session];
+                    unscheduledStore[index] = [session];
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


niranjan94:
@SaptakS we could make it such that all sessions have `null`/`None` as their times when they're created. I guess that's how it's supposed to be.

SaptakS:
@niranjan94 that can be done. But all in all we need to modify the logic of pushing session only to the set of one date... All sessions must be pushed to all date sets at the beginning I think... What do you say?? null/None won't affect much in the logic at present.

mariobehling:
Setting a session to all dates could be a temporary solution, if changing the logic takes a longer time. But setting it to "None" would actually be the logical implementation.

SaptakS:
@mariobehling @niranjan94 updated code and provided screenshots. Please check.

app/static/js/events/scheduler.js
```diff
@@ -763,12 +778,19 @@ function processMicrolocationSession(microlocations, sessions, callback) {
...
 
             var dayIndex = _.indexOf(days, dayString);
-
             if (_.isArray(sessionsStore[dayIndex])) {
                 sessionsStore[dayIndex].push(session);
             } else {
                 sessionsStore[dayIndex] = [session];
             }
+
+            for (var dayIndex in days) {
```
niranjan94: @SaptakS please fix this issue. Looks good otherwise.

mariobehling:
Good work. Please also see this follow up issue here: https://github.com/fossasia/open-event-orga-server/issues/3192

I am unclear about the start and end time. I do not see that it is defined for all days, but available times are limited on the calendar.

PR #3189, opened 2017-02-12
Fixed: Tickets were not deleting from database

### LINKED ISSUE #3188: Deleting ticket not executed
After the user deletes a ticket, it is not shown anymore and when the user continues and saves it it shows as successfully done. However going back to the editing of the event shows: the ticket is still there. It also shows up on "Access Code" creation if it is a hidden ticket.

Expected: If no purchase has been made, tickets it should be possible to delete a ticket successfully.

screenshot from 2017-02-11 23-23-33

poush:
Fixes #3188 

To reviewer: Please let me know if short note required for the changes made and what was causing the problem.

Migration were created to modify column property in "order_tickets" table. Because of CASCADE nature and non nullable primary key set on "ticket_id" column. We were not able to set NULL to this column.


app/models/ticket.py
```diff
@@ -70,8 +70,12 @@ def has_order_tickets(self):
         Else False.
         """
         from app.helpers.helpers import get_count
-        count = get_count(OrderTicket.query.filter_by(ticket_id=self.id))
-        return bool(count > 0)
+        count =  Order.query.filter( Order.id.in_( OrderTicket.query.with_entities(OrderTicket.order_id).filter_by(ticket_id=self.id).all() ) ).filter_by(status='completed').all()
```
poush: @niranjan94 Please let me know if I can improve the writing style of this query

app/models/order.py
```diff
@@ -7,7 +7,7 @@
 class OrderTicket(db.Model):
     __tablename__ = 'orders_tickets'
     order_id = db.Column(db.Integer, db.ForeignKey('orders.id', ondelete='CASCADE'), primary_key=True)
-    ticket_id = db.Column(db.Integer, db.ForeignKey('ticket.id', ondelete='CASCADE'), primary_key=True)
+    ticket_id = db.Column(db.Integer, db.ForeignKey('ticket.id', ondelete='CASCADE'), nullable=True)
```
niranjan94: What is the logic behind this change ?

app/models/ticket.py
```diff
@@ -70,8 +70,12 @@ def has_order_tickets(self):
...
```
niranjan94: Make this statement more clear to understand. Take `OrderTicket.query.with_entities(OrderTicket.order_id).filter_by(ticket_id=self.id).all()` out. 

Use the `get_count` method. We don't want a full select query to run. All we need is the count.

app/models/order.py
niranjan94: An `ondelete='SET NULL'` would have required a `nullable` column. Whereas `CASCADE` will anyway just delete the record. Which is what we want to happen. Please revert back to the previous state and remove the migrations.

poush: That was the issue,```CASCADE``` was trying to set ```None``` to ```ticket_id``` instead of deleting the row.

```Dependency rule tried to blank-out primary key column 'orders_tickets.ticket_id' on instance '<OrderTicket at 0x10c7debd0>'```

This was the error

File: app/models/ticket.py (hunk 3 above)
Comment (poush) [inline app/models/ticket.py]: Yes, I noticed the same thing and will update this.

File: app/models/order.py (hunk 2 above)
Comment (poush) [inline app/models/order.py]: @niranjan94 
```IntegrityError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely) (psycopg2.IntegrityError) null value in column "ticket_id" violates not-null constraint
DETAIL:  Failing row contains (1, null, 1).
 [SQL: 'UPDATE orders_tickets SET ticket_id=%(ticket_id)s WHERE orders_tickets.order_id = %(orders_tickets_order_id)s'] [parameters: ({'ticket_id': None, 'orders_tickets_order_id': 1}, {'ticket_id': None, 'orders_tickets_order_id': 2}, {'ticket_id': None, 'orders_tickets_order_id': 3}, {'ticket_id': None, 'orders_tickets_order_id': 4}, {'ticket_id': None, 'orders_tickets_order_id': 5}, {'ticket_id': None, 'orders_tickets_order_id': 6}, {'ticket_id': None, 'orders_tickets_order_id': 7})]
```

When I removed primary constraint but without nullable field

niranjan94: Then please check why `CASCADE` is failing to delete the rows and fix that. A `CASCADE` is supposed delete the row when the related parent row is deleted.

poush:
Will make another pull request.

niranjan94:
@poush please avoid closing and re-opening PRs for a single issue. Hard to keep track of it.

poush:
@niranjan94 okay then I am making changes to here again.

poush:
@niranjan94 Sure will take care from next time.

Issue #2803, opened 2016-12-31
Implement Apps similar to twitter applications page


mariobehling:
Connected apps are not currently implement in a clear way, e.g. Google does not show up anywhere as a connected app.

Please create a tab "Applications" in the settings list at https://eventyay.com/settings/ similar to twitter and move the connected accounts from "Profile" here.

![screenshot from 2016-12-31 21-18-38](https://cloud.githubusercontent.com/assets/1583873/21579106/bfa95572-cf9e-11e6-9d4b-2c56970b43e7.png)

Change the display as connected (instead of a link display like on the profile).

Ensure the connections for the following show up:
* [ ] Facebook
* [ ] Google
* [ ] Twitter
* [ ] Instagram

![screenshot from 2016-12-31 20-49-04](https://cloud.githubusercontent.com/assets/1583873/21579052/e9a39d14-cf9a-11e6-89c1-555add656ff9.png)

gaeun:
@mariobehling Can I work on this please?

Elgayar:
@mariobehling @gaeun Has this issue been solved?

gaeun:
@Elgayar Yeah, refer to PR #2803. I just have to fix Travis errors.

mariobehling:
Thanks for the last PR. The issue is not solved entirely yet. So help is welcome. Also see my comments here, please: https://github.com/fossasia/open-event-orga-server/pull/2839#issuecomment-271061446

arpitn30:
I fixed the "Facebook does not show up" bug, should a put up a pull request or solve other bugs too before doing so?

mariobehling:
@arpitn30 Please make a PR and mention that it "partly fixes" the issue.

poush:
Also, I think we need to add a check as well where fields of those platforms will not be shown for which app token and secret in admin setting isn't set

niranjan94:
@poush we already have an OAuth implementation in the server. (we use it for login now). Please reuse the existing implementation and not add more libraries.

poush:
@mariobehling @niranjan94 Are we looking to store the tokens as well to perform any actions like giving option user/organizer to post on their page automatically?

poush:
@mariobehling Are we expecting experience similar to this below
![screenflow](https://cloud.githubusercontent.com/assets/1812082/26278407/9d387716-3db7-11e7-8b39-46e026dd0926.gif)

niranjan94:
@poush

PR #3190, opened 2017-02-12
Partly Solves #2803: Show facebook on applications page and aligned application fields

arpitn30:
This partly fixes issue #2803.
The facebook field was only showing up if user details were available so I removed the conditional statement and added the css attribute to align all the application fields.
application

### DIFF
--- a/app/templates/gentelella/users/settings/pages/applications.html
+++ b/app/templates/gentelella/users/settings/pages/applications.html
@@ -14,13 +14,12 @@
 {% block content %}
 
     <div style="display: block;">
-        {% if user.user_detail.facebook %}
         <div class="row" style="margin-bottom:20px">
 
             <div class="col-md-6 col-xs-6">
                 <label>Facebook:</label>
                 <div class="input-group">
-                    <span class="input-group-addon">facebook.com/</span>
+                    <span class="input-group-addon social-addon">facebook.com/</span>
                     <input type="text" name="facebook" class="form-control"
                            value="{{ user.user_detail.facebook|default('', true) | replace('https://facebook.com/', '') }}"
                            style="min-width: 400px"
@@ -37,13 +36,12 @@
             </div>
 
         </div>
-        {% endif %}
 
         <div class="row" style="margin-bottom:20px">
             <div class="col-md-6 col-xs-6">
                 <label>Twitter:</label>
                 <div class="input-group">
-                    <span class="input-group-addon" id="basic-addon3">twitter.com/</span>
+                    <span class="input-group-addon social-addon" id="basic-addon3">twitter.com/</span>
 
                     <input type="text" name="twitter" class="form-control"
                            value="{{ user.user_detail.twitter|default('', true) | replace('https://twitter.com/', '') }}"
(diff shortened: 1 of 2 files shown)


app/static/css/vendor/gentelella.css
```diff
@@ -5185,3 +5185,6 @@ ul.notifications {
   width: 0;
   }
 }
+#basic-addon3{
```
niranjan94: Do not make any changes in this file or any other file inside the vendor directory.

app/templates/gentelella/users/settings/pages/applications.html
```diff
@@ -14,13 +14,12 @@
...
 
     <div style="display: block;">
-        {% if user.user_detail.facebook %}
         <div class="row" style="margin-bottom:20px">
 
             <div class="col-md-6 col-xs-6">
                 <label>Facebook:</label>
                 <div class="input-group">
-                    <span class="input-group-addon">facebook.com/</span>
+                    <span class="input-group-addon" id="basic-addon3">facebook.com/</span>
```
niranjan94: Don't use ids like this for reference. Instead see if you can use a shared class and share the styling across similar components.

arpitn30: The class is also being used in the search button and I checked the ids are only used in the span tags. Should I just change the property of each span tag?

app/templates/gentelella/users/settings/pages/applications.html
```diff
@@ -14,13 +14,12 @@
...
 
     <div style="display: block;">
-        {% if user.user_detail.facebook %}
         <div class="row" style="margin-bottom:20px">
 
             <div class="col-md-6 col-xs-6">
                 <label>Facebook:</label>
                 <div class="input-group">
-                    <span class="input-group-addon">facebook.com/</span>
+                    <span class="input-group-addon" id="basic-addon3" style="min-width: 135px;">facebook.com/</span>
```
niranjan94: Please remove the ID. 
And instead of putting the `min-width` in each element, please move it to stylesheet

arpitn30:
Created class to implement css and added property to css file. Removed id attribute from facebook span tag

app/templates/gentelella/users/settings/pages/applications.html
arpitn30: The id attribute has been included in other blocks so I though to add it to the facebook block as well. Anyway removed it now as requested.

app/templates/gentelella/users/settings/pages/applications.html
```diff
@@ -14,13 +14,12 @@
...
 
     <div style="display: block;">
-        {% if user.user_detail.facebook %}
         <div class="row" style="margin-bottom:20px">
 
             <div class="col-md-6 col-xs-6">
                 <label>Facebook:</label>
                 <div class="input-group">
-                    <span class="input-group-addon">facebook.com/</span>
+                    <span class="input-group-addon minwidth">facebook.com/</span>
```
niranjan94: Don't give class names like this. Give a class name that is understandable and reusable. Like maybe `social-addon` or something like that.

arpitn30: Done!

PR #3191, opened 2017-02-12
Fixed: Tickets were not deleting from database

### LINKED ISSUE #3188: Deleting ticket not executed

poush:
Fixes #3188

### DIFF
--- a/app/models/ticket.py
+++ b/app/models/ticket.py
@@ -31,6 +31,7 @@ class Ticket(db.Model):
     event = db.relationship('Event', backref='tickets')
 
     tags = db.relationship('TicketTag', secondary=ticket_tags_table, backref='tickets')
+    order_ticket = db.relationship('OrderTicket', backref="ticket", passive_deletes=True)
 
     def __init__(self,
                  name=None,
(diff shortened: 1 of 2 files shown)


app/models/ticket.py
```diff
@@ -31,6 +31,7 @@ class Ticket(db.Model):
     event = db.relationship('Event', backref='tickets')
 
     tags = db.relationship('TicketTag', secondary=ticket_tags_table, backref='tickets')
+    OrderTicket = db.relationship('OrderTicket', backref="ticket", passive_deletes=True)
```
niranjan94: Please use `snake_case` naming schema

poush:
@niranjan94 Done with that change

app/models/ticket.py
```diff
@@ -1,4 +1,4 @@
-from app.models.order import OrderTicket, Order
+    from app.models.order import OrderTicket, Order
```
niranjan94: Please fix the codacy issue.

Issue #3196, opened 2017-02-13
Permissions are not editable in admin tab


mariobehling:
On https://eventyay.com/admin/permissions/#!eventroles the eventroles are not editable by the "admin" role.

![screenshot from 2017-02-13 12-17-48](https://cloud.githubusercontent.com/assets/1583873/22884809/08e1edb0-f1f7-11e6-8467-711c0a83a367.png)

Related to: https://github.com/fossasia/open-event-orga-server/issues/3195#issuecomment-279359247

Princu7:
@mariobehling I would like to work on this. I will send a fix soon.

mariobehling:
@Princu7 Can you confirm that the functions work? As a co-organizer I still receive the info "error saving" even though the co-organizers received all permission. Maybe there are permissions that we do not cover yet?

![screenshot from 2017-02-13 23-16-33](https://cloud.githubusercontent.com/assets/1583873/22905973/be569fba-f242-11e6-8975-f5dc080329cf.png)

Princu7:
@mariobehling Yeah, sure. I will have a look at the cause of the error :)

Princu7:
@mariobehling @niranjan94 The permissions are saved correctly. The problem occurs when the co-organizer saves the changes he made. Currently, we send an AJAX request to save the modifications. Inside the main event route to which the request is sent, we only check whether the current user is staff or organizer of that event or not. We don't do any additional checks. This is the reason why the request is aborted and we receive the info "error saving".

Here is the little code snippet responsible for the error

```
if not current_user.is_staff and not current_user.is_organizer(event_id):
    abort(403)
```

To do it correctly, we have to check the other roles of the current user in that event. And check whether he has the permissions to make the changes requested or not. 
I am trying to fix this issue. If I understood something wrong, please correct me :) Thanks!!

niranjan94:
@Princu7 you're right. Could you make a PR with the fix. ?

Princu7:
@niranjan94 Sure. I am working on it. Will send a patch soon :)

Princu7:
@niranjan94 @mariobehling I have done some work on this issue but there are many areas which I want to discuss.
Here is the test server: http://ancient-thicket-44821.herokuapp.com
Admin Credentials: 
```
user-id: princu7@gmail.com
password: fossasia
```

Co-Organizer Credentials(of the event _Hurry_)
```
Event link: http://ancient-thicket-44821.herokuapp.com/events/8/
[...]

PR #3205, opened 2017-02-17
Wizard permissions

Princu7:
**Work in Progress**
Issue #3196 

Please see the issue for details. Thanks!


app/static/js/events/wizard/vue-app.js
```diff
@@ -1,4 +1,6 @@
 var steps = ['', 'sponsors', 'sessions-tracks-rooms'];
+var modifications = {'sessions-tracks-rooms': {sessionCreate:0, sessionRemove:0, trackCreate:0, trackRemove:0, roomCreate:0, roomRemove:0},
```
niranjan94: Instead of doing all this, let the wizard itself be dynamic depending the permission a user has. For example, If a user cannot edit tracks, tracks section without any option to edit can be shown.

Princu7: @niranjan94 That's a good idea. I am making a pull request giving the co-organizer the same permissions as the organizer for the wizard permissions. Closing this request for a while.

PR #3218, opened 2017-02-18
Fixes - Scheduler: Search results showing up multiple times

### LINKED ISSUE #3214: Scheduler: Search results showing up twice when changing the date
When the user does a search e.g. "Python" on one date and then clicks another date and performs the same search, the search results are showing the same sessions twice. If the search is performed again, the search is showing up three times and more times for more searches.

screenshot from 2017-02-18 18-42-46

poush:
Fixes #3214

### DIFF
--- a/app/static/js/events/scheduler.js
+++ b/app/static/js/events/scheduler.js
@@ -969,6 +969,7 @@ $("#sessions-search").valueChange(function (value) {
     }
 
     filtered = _.sortBy(filtered, "title");
+    filtered = _.uniqBy(filtered, "id");
 
     $unscheduledSessionsHolder.html("");
 


mariobehling:
@poush Do you have some info on the cause of the error?

poush:
Yes, I checked code for the cause of this error. Looks like when we select the another date for timelime ```addSessionToUnscheduled()``` called from ```scheduler.js``` Line no nearby to ```880```

The issue is we are passing ```false``` to ```addSessionToUnscheduled``` casuing it to add dublicate object of same session in ```unscheduledStore```

poush:
If we try to log ```unscheduledStore``` every time when we select another date, we can see that redundant objects are being added to it. This is because of the limitation of ```addSessionToUnscheduled()``` and one possible solution could be introducing one parameter in function which can state whether it is called for due to timeline change or we can use ```_.uniqBy``` inside this function since primarily responsibility of setting up this object is of function ```addSessionToUnscheduled```

PR #3231, opened 2017-02-22
Speaker and Session form separated for 'call for speakers'

### LINKED ISSUE #3069: Enhance Form for Call of Speakers
The form to answer the call for speakers needs to be improved to be fully usable, e.g. if a speaker wants to submit two proposals it is currently not possible. Therefore, please implement the following:

* [ ] When the user clicks "Submit Proposal" on Call for Speakers page open a pop up with the following options a) Register as a Speaker and b) Propose Session
* [ ] Split forms for Speakers and Registration into separate forms
* [ ] If speaker has already filled in a session form offer option: "Add Another Session"

magdalenesuo:
fixes #3069 
No color check- https://github.com/fossasia/open-event-orga-server/issues/3069#issuecomment-279318801 not done (will create another issue and solve)

**What Has been Done:**
**_1)_** **If User Not logged in:** _Pop up to log in_ (already done in another PR)

**_2)_** **If User logged in:** Give option for _(a) Registering as speaker_ or _(b) Proposing Session._
        In case of **Register as Speaker** , he goes to only speaker form.
        In case of **Proposing Session**, The user gets to both Session and Speaker form (Since session is meaningless without speaker)
screenshot from 2017-02-22 18-53-16

**_3)_** **If User is registered as speaker**: _He/She can either edit his already existing session or add more sessions_
       In **Add more session** case, the user goes to session form only, not speaker form.
screenshot from 2017-02-22 20-27-39

### DIFF
--- a/app/helpers/sessions_speakers/speakers.py
+++ b/app/helpers/sessions_speakers/speakers.py
@@ -47,7 +47,8 @@ def save_speaker(request, event_id=None, speaker=None, user=None):
         save_to_db(speaker)
 
     speaker.email = trim_get_form(request.form, 'email', None)
-    speaker.name = trim_get_form(request.form, 'name', None)
+    if trim_get_form(request.form, 'name', None):
+        speaker.name = trim_get_form(request.form, 'name', None)
 
     if not speaker.user:
         if user:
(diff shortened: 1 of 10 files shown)


mariobehling:
Thank you! How does this change influence the functioning if the organizer adds a speaker in the sessions tab? Does this still work?

screenshot from 2017-02-22 23-50-34

magdalenesuo:
@mariobehling yes, they still work the normal way :)

PR #3234, opened 2017-02-23
Fixes faulty tooltip

### LINKED ISSUE #3222: Faulty tool tips on unscheduled sessions in  scheduler
This is how the tool tip looks on a scheduled session. 
image

However once a scheduled session is unscheduled , it appears in this form, where it is not supposed to appear. Only the sessions which have been unscheduled after being scheduled seem to behave like this. 
image

I would like to work on this.

CosmicCoder96:
fixes #3222 
giphy

### DIFF
--- a/app/static/js/events/scheduler.js
+++ b/app/static/js/events/scheduler.js
@@ -342,9 +342,11 @@ function addSessionToUnscheduled(sessionRef, isFiltering, shouldBroadcast) {
     }
 
     try {
-        sessionRefObject.$sessionElement.popover('destroy');
-    } catch(ignored) { }
-
+        setTimeout( function() {
+            $('.session.unscheduled').popover('hide');
+        }, 100);
+    }
+    catch(ignored) { }
 }
 
 /**


app/static/js/events/scheduler.js
```diff
@@ -342,9 +342,10 @@ function addSessionToUnscheduled(sessionRef, isFiltering, shouldBroadcast) {
     }
 
     try {
-        sessionRefObject.$sessionElement.popover('destroy');
-    } catch(ignored) { }
-
+        setTimeout( function() {
+            $('.session.unscheduled').popover('hide');}, 100);
```
niranjan94: Align and indent the code properly. Make it more readable. Otherwise it's okay.

CosmicCoder96: The delay is added because the issue was caused in the first place  due to asynchronous code.
And the popover('destroy'), did not delete the content inside the popover.

CosmicCoder96:
@niranjan94 Thanks for the review, Done.

PR #3280, opened 2017-03-01
Scheduler scroll fix

### LINKED ISSUE #3257: Public Schedule: Only show rooms with sessions and only show time from first session and until end of last session
The public schedule currently shows all available rooms and times for sessions. In the scheduler used by the organizer to create the events this is what is needed. 

In the public event schedule page only the used rooms and the times from the first session until the last session should be displayed.

Empty rooms are not important information for users.

screenshot from 2017-02-26 02-30-19

SaptakS:
Fixes #3257 

Works Done:

- [x] Show only rooms with sessions
- [x] Show time from first session to end of last session

See the screenshots below

screenshot from 2017-03-01 21-23-03
screenshot from 2017-03-01 21-23-08

@mariobehling @niranjan94 pease review.

### DIFF
--- a/app/static/js/events/scheduler.js
+++ b/app/static/js/events/scheduler.js
@@ -121,6 +121,8 @@ var $unscheduledSessionsHolder = $unscheduledSessionsList;
 var $noSessionsInfoBox = $("#no-sessions-info");
 var $dayButtonsHolder = $("#date-change-btn-holder");
 var $addMicrolocationForm = $('#add-microlocation-form');
+var $timelineTable = $('table.timeline-table');
+var $noSessionMessage = $('#no-session-message');
 
 var $mobileTimeline = $("#mobile-timeline");
 var $tracksTimeline = $("#tracks-timeline");
@@ -229,7 +231,6 @@ function addSessionToTimeline(sessionRef, position, shouldBroadcast) {
             sessionRefObject.session = updateSessionTime(sessionRefObject.$sessionElement);
         }
     }
-
 
     sessionRefObject.$sessionElement.css({
         "-webkit-transform": "",
@@ -849,8 +850,9 @@ function loadDateButtons() {
  */
 function loadMicrolocationsToTimeline(day) {
 
-    $('table.timeline-table').show();
-    $('#no-session-message').hide();
+    $timelineTable.show();
+    $noSessionMessage.hide();
+    $microlocationsHolder.find(".microlocation").show();
 
     var parsedDay = moment.utc(day, "Do MMMM YYYY");
     if (parsedDay.isSame(mainEvent.start_time, "day")) {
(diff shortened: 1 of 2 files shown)


app/static/js/events/scheduler.js
```diff
@@ -859,10 +862,45 @@ function loadMicrolocationsToTimeline(day) {
...
 
-    generateTimeUnits();
-
+    var least_hours = dayLevelTime.start.hours;
+    var least_minutes = dayLevelTime.start.minutes;
+    var max_hours = 0;
+    var max_minutes = 0;
     var dayIndex = _.indexOf(days, day);
 
+    if (/.+\/e\/.+\/schedule\//i.test(document.URL)) {
```
niranjan94: What is this test for ?

SaptakS: This test is so that this changes take place only in the public page and not in the scheduler in event dashboard. Because in event dashboard we need the entire time rather than from start session to end session. Same for the room.

niranjan94: @SaptakS

niranjan94: @SaptakS use the `isReadOnly()` function. `isReadOnly()` returns a `true` if the scheduler is running the public page.

SaptakS: ok cool.

SaptakS:
@niranjan94 made the changes. Please review.

app/static/js/events/scheduler.js
```diff
@@ -919,6 +957,12 @@ function loadMicrolocationsToTimeline(day) {
     $("[data-toggle=tooltip]").tooltip("hide");
 
     if (isReadOnly()) {
+        _.each($microlocations, function ($microlocation) {
+            $microlocation = $($microlocation);
+            if ($microlocation.find('.microlocation-inner').html().trim() === "") {
```
niranjan94: Instead of this, a more straightforward way would be to count the children.
`$microlocation.find('.microlocation-inner').children().length;`

app/static/js/events/scheduler.js
```diff
@@ -849,6 +848,10 @@ function loadDateButtons() {
  */
 function loadMicrolocationsToTimeline(day) {
 
+    $('table.timeline-table').show();
```
niranjan94: Move all selectors to the above (https://github.com/SaptakS/open-event-orga-server/blob/9c1f7b9f57ee26220ecd27825f818c7269e0e8bc/app/static/js/events/scheduler.js#L123) for consistency

app/static/js/events/scheduler.js
```diff
@@ -849,6 +848,10 @@ function loadDateButtons() {
...
+    $('#no-session-message').hide();
```
niranjan94: This too

app/static/js/events/scheduler.js
```diff
@@ -859,10 +862,45 @@ function loadMicrolocationsToTimeline(day) {
...
+                    max_hours= session.end_time.hours();
+                    if (session.end_time.minutes() > max_minutes) {
+                        max_minutes = session.end_time.minutes();
+                    }
+                }
+            }
+        });
+
+        if (max_hours === 0) {
+            $('table.timeline-table').hide();
```
niranjan94: These too

app/static/js/events/scheduler.js
SaptakS: This would give the same result, right? Can you explain why this is a better approach?

niranjan94: The result would be the same ... In most cases ... But by chance (an edge case) due to some error in the template or js code, a non-whitespace character ends up inside the `div`. Then just trimming and equating it to empty space would fail. 

And it's always better to look at the DOM as an object instead of a string.

PR #3281, opened 2017-03-01
start_time of scroll fixed

### LINKED ISSUE #3257: Public Schedule: Only show rooms with sessions and only show time from first session and until end of last session

SaptakS:
Fixes #3257


mariobehling:
screenshot from 2017-03-02 09-23-39

Sessions do not show up on FOSSASIA Summit. Second day times are not correct.

Issue #3085, opened 2017-01-30
Speakers Proposal Acceptance: Add Status "Confirmed"


mariobehling:
After a speaker submits a proposal currently there are the following states: 1. Pending, 2. Accepted, 3. Rejected.

When the session is accepted the speaker receives the info that he is accepted. The organizer still does not have any info if the speaker is really confirmed and has booked the travel. Therefore we need an additional state "Confirmed"

Please implement an additional button "Confirmed" and a filter on the status filter of the page.

![screenshot from 2017-01-30 14-23-38](https://cloud.githubusercontent.com/assets/1583873/22424738/2a5a79bc-e6f8-11e6-9472-51e89c30993b.png)

nishapoyarekar:
I would like to work on this. @SaptakS can you please assign this issue to me?

niranjan94:
@nishapoyarekar you can proceed

nishapoyarekar:
My understanding of the issue is that once Accepted, a mail needs to be sent to the speaker asking them to confirm their availability for the session.
I have attached a screenshot with the change-
<img width="1265" alt="confirmationemailiconandfilter" src="https://cloud.githubusercontent.com/assets/20510268/23354551/b81fc940-fcf7-11e6-98bd-a56408fbbd7d.png">
<img width="1087" alt="confirmationemail" src="https://cloud.githubusercontent.com/assets/20510268/23354550/b81fa1c2-fcf7-11e6-9ab2-00176f7cd983.png">


Is this what is expected?

mariobehling:
Yes, the right direction. Please implement as follows:
* [ ] Change color of green accepted buttons to yellow
* [ ] After session is accepted fffer two buttons in green with
   * [ ] a) Confirm without email
   * [ ] b) Confirm with email
* [ ] Keep the red buttons to reset the status to "rejected" if organization of sessions requires this at a later stage.

mariobehling:
@nishapoyarekar Thanks for the commit. Please change the color in accepted to "yellow" (not orange).

Currently it is not possible to send an email to "sessions pending".

![screenshot from 2017-03-07 09-38-24](https://cloud.githubusercontent.com/assets/1583873/23638022/19743f6e-031a-11e7-8888-f60a0e79f5bf.png)

Himanshi-Khandelwal:
@nishapoyarekar are you working on this?

Himanshi-Khandelwal:
If no one then I'll like to do the change.

[...]

PR #3290, opened 2017-03-03
Add the status 'Confirmed' in sessions view

nishapoyarekar:
#3085 need an additional state "Confirmed"
When the session is accepted the speaker receives the info that he is accepted.
The organizer still does not have any info if the speaker is really confirmed and has booked the travel. Therefore we need an additional state "Confirmed"

- Added a new value 'confirmed' to the choice_list in app/api/helpers/special_fields.py
- Trigger session state change in app/api/session.py for confirmed state -- line no. 166
- Added list element 'Confirmed' and maintain count confirmed.count in app/templates/gentelella/super_admin/events/events.html
- Added filter label in top row for confirmed, Add confirmed green button with and without email on click of both change status to confirmed, change accepted to yellow button in app/templates/gentlella/users/events/sessions/base_session_table.html
- Added green label for status 'confirmed' in details table app/templates/gentelella/users/mysessions/mysession_detail.html

### DIFF
--- a/app/api/helpers/special_fields.py
+++ b/app/api/helpers/special_fields.py
@@ -73,5 +73,5 @@ class SessionStateField(fields.ChoiceString):
 
     def __init__(self, **kwargs):
         super(SessionStateField, self).__init__(
-            choice_list=['pending', 'accepted', 'rejected'],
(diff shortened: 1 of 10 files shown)


nishapoyarekar:
@niranjan94 I have resolved the conflict and added issue number to the description #3085

PR #3303, opened 2017-03-05
Added extra 2 hours of time below to solve bottom info showing problem

### LINKED ISSUE #3287: Session information on bottom entry hidden in schedule calendar view
In the calendar the bottom entries are hidden within the frame. Expected: Show session info on top.

screenshot from 2017-03-03 09-21-09

SaptakS:
Fixes #3287

Solved by increasing the end time boundary in schedule by 2 hours. This allows for space to show the extra info which otherwise wasn't being shown. Other way of doing it is making the dropdown for bottom elements to appear on top but that would require too much computation. This is computationally simple and solves the purpose.

screenshot from 2017-03-05 18-04-08

@mariobehling @niranjan94 please review.

### DIFF
--- a/app/static/js/events/scheduler.js
+++ b/app/static/js/events/scheduler.js
@@ -901,7 +901,7 @@ function loadMicrolocationsToTimeline(day) {
         window.dayLevelTime.start.hours = least_hours;
         window.dayLevelTime.start.minutes = 0;
 
-        window.dayLevelTime.end.hours = max_hours;
+        window.dayLevelTime.end.hours = max_hours + 2;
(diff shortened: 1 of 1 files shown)


niranjan94:
Why would that require too much computation ? I think it can be solved by adjusting the `z-index` of the popover window. Did you try that instead of this ?

SaptakS:
Yes tried. Z index doesn't work because that div is inside the timeline div and the overflow-y is hidden. So increasing the height is the only solution to show it. It is not that the thing is behind so I can increase zindex. It is just it crosses the height assigned to parent div.

PR #3311, opened 2017-03-06
Added png export in public event page

poush:
Please review this png export process.

Sample export
timeline 88

Export Process
screenflow

The concept is simple, our timeline gets full screen for a moment to get full export including date published and date selected.


mariobehling:
@niranjan94 Let's get this to master to try it out.

PR #3336, opened 2017-03-08
Making tracks view on schedule page of an event similar to web-app

### LINKED ISSUE #3259: Tracks Page does not show additional info
Sessions on the tracks page are supposed to show additional info on clicking on the specific session (similar to web app and sessions view). Currently the area does not expand on clicking it.

screenshot from 2017-02-26 15-00-30

Princu7:
Fixes #3259 
* Tries to implement the tracks page of the schedule of an event closer to web app
* Test Server http://ancient-thicket-44821.herokuapp.com/e/624953ec/schedule/

### DIFF
--- a/app/static/js/events/scheduler.js
+++ b/app/static/js/events/scheduler.js
@@ -140,6 +140,7 @@ var dayButtonTemplate = $("#date-change-button-template").html();
 
 var mobileMicrolocationTemplate = $("#mobile-microlocation-template").html();
 var mobileSessionTemplate = $("#mobile-session-template").html();
+var mobileSpeakerTemplate = $("#mobile-speaker-template").html();
 /**
  * Data Getters
  * ============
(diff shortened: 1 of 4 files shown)


app/static/js/events/scheduler.js
```diff
@@ -259,18 +260,31 @@ function addSessionToTimeline(sessionRef, position, shouldBroadcast) {
...
-        speakers.push(speaker.name);
-    });
-    content += speakers.join(', ');
     $mobileSessionElement.find('.time').text(sessionRefObject.session.start_time.format('hh:mm A'));
-    $mobileSessionElement.find('.event').text(content);
-    $mobileSessionElement.find('.event').attr("data-target", "#session-track-details"+sessionRefObject.session.id);
+    $mobileSessionElement.find('.title').text(sessionRefObject.session.title);
     $mobileSessionElement.find('.session-track-details').attr("id", "session-track-details"+sessionRefObject.session.id);
-    $mobileSessionElement.find('.session-speakers').text("Speakers: " + speakers.join(', '));
+    $mobileSessionElement.find('.event').attr("data-target", "#session-track-details"+sessionRefObject.session.id + ",#session-track-short-details"+sessionRefObject.session.id);
```
SaptakS: Why do we need two comma separated ids in data target?

Princu7: There are two collapsible elements here. So the two ids refer to them. One of them shows the shorter version of the details of the speakers(it is visible by default). The other one is the longer version of the speaker details which is hidden by default. When we click on the container, the former one becomes hidden and the latter one becomes visible.

SaptakS: Isn't there a way to avoid two data target? I think having two data targets reduces code readability. If there isn't then will go with this. If there is, I would prefer to see that solution

Princu7: We could give both the divs a common class, maybe? Testing it right now!!

Princu7: Okay, I tried doing this using the class method but there seems to be problem. The collapsing is not working well and the layout of the page gets disturbed a bit. Will continue trying but it looks less likely right now.

mariobehling:
@Princu7 
* Please also add end time of sessions to the view.
* Ensure there is a space after speaker/before company: Speaker Name (Company/Organization)

Princu7:
@saptaks I tried to target the collapsible data elements by having a shared class but it didn't seem to work properly. Both were collapsing and expanding at the same time and the layout of the page was disturbed. I added the code which sets the default avatar for an image in case it is not found.
@mariobehling Done the changes

Please review
Test Server: http://ancient-thicket-44821.herokuapp.com/e/624953ec/schedule/

mariobehling:
Good job.

PR #3342, opened 2017-03-09
Fix DataTables breaking on Events Dashboard

### LINKED ISSUE #3252: Events Dashboard: Live, Draft, Past events overview table not consistent anymore
All views of Live, Draft and Past events should be consistent. They are not anymore currently, e.g. Draft page is showing a table that does not occupy 100% of the space.

screenshot from 2017-02-25 23-59-59

cepradeep:
Apparently, initializing DataTables on tables inside elements with
`display: none;` breaks their sizing. This is because DataTable is
unable to compute the size of such elements.

This was fixed by initializing DataTables only after the tab has
been opened. Also fixed updating the URL hash according to which
tab is open.

Fixes #3252

### DIFF
--- a/app/static/js/events/event-dashboard.js
+++ b/app/static/js/events/event-dashboard.js
@@ -1,0 +1,64 @@
+// Function that initializes a DataTable on .tables inside $(element).
+function initDataTable(element) {
+    if (!$.fn.DataTable.isDataTable(element + ' .table')) {
+        // The table on 'Import' tab has a different number of columns.
+        if (element === '#import') {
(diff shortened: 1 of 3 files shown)


cepradeep:
@niranjan94 I'm still learning the ropes, so does this look right?

cepradeep:
@niranjan94 Done!

Himanshi-Khandelwal:
@cepradeep I think this javascript file is not working , when I tried to change the column width , it was not showing any change.

PR #3344, opened 2017-03-09
event sessions: fix auto scroll to top

### LINKED ISSUE #3254: Scrolling session very slow
On https://eventyay.com/e/45da88b7/sessions/ the scrolling down for session info is very slow.

shubham-padia:
Fixes #3254 .(On sessions page, page would scroll to top automatically)

For reviewer:
This was a problem caused in sticky kit due to an update in chrome (issue: https://github.com/leafo/sticky-kit/issues/219). This problem has been fixed by someone in https://github.com/leafo/sticky-kit/pull/218, but is not merged by the maintainer, so placed the corrected files in `static/js/vendors`. Cheers.


shubham-padia:
> This might work now but future updates for sticky kit won't be downloaded as it is no longer installed through bower. The entire purpose of bower is to avoid having such files in the repo. Even if this is the only fix, this doesn't seem to be the right way of doing

@SaptakS We're already doing this for many of the files present in `static/js/vendors`, like icheck, multi-select and many more. Navigating the sessions page on chrome isnt possible without this. The current version is pretty stable and IMO there's nothing wrong with using a stable version of the file in the repo. Lets wait for @mariobehling @niranjan94 's inputs for now.....

mariobehling:
@shubham-padia @SaptakS If there is no other simple solution for now, please proceed and open a follow up issue to solve this with bower or another way later.

shubham-padia:
Did the requested changes @niranjan94

Issue #3339, opened 2017-03-09
Public Schedule add event name and move some info to bottom


mariobehling:
On the top area please add the name of the event in front of "Schedule".

Also move "Published on..", e.g.  (Published on March 01, 2017) to the bottom without brackets.

![screenshot from 2017-03-09 14-18-01](https://cloud.githubusercontent.com/assets/1583873/23738451/f4477772-04d3-11e7-93c3-12e546405b03.png)

LuD1161:
@mariobehling 
Is this good ?
![scheduler](https://cloud.githubusercontent.com/assets/17861054/23781616/aa455b3e-0574-11e7-9beb-0f97fe2712e3.png)

mariobehling:
Yes, very good.

Please also move "Published on.." to the same line as "All timings...". Keep it on the left side and get rid of the "( )" brackets.

LuD1161:
Making a PR with the a new branch and a new forked repo ...
The previous one was all messed up
Hope this is what is needed..

![image](https://cloud.githubusercontent.com/assets/17861054/23787630/f27216c6-0598-11e7-9640-f62e6cae1df9.png)

PR #3347, opened 2017-03-10
Changed the 'Published on' position and added event name to scheduler

LuD1161:
Done changes.
Solved #3339 

image


LuD1161:
@SaptakS 
It's done
screenshot from 2017-03-10 16-22-12

SaptakS:
@LuD1161 it still doesn't seem fixed. Also, it is now a lot below... It should be in the same top level as timezone and same left level as schedule.

LuD1161:
@SaptakS 
It's Done , hopefully

niranjan94:
@LuD1161 along with @SaptakS 's comments, also squash your commits into one.

LuD1161:
@niranjan94 

It's done.

image

Issue #3295, opened 2017-03-04
Add png export to public schedule page


mariobehling:
We already have a png export of the schedule, but it is only implemented for the organizers. It also has issues as it only exports the screen view and not always the full calendar.

Please implement the following:
* [ ] Add png export to the public schedule page e.g. https://eventyay.com/e/45da88b7/schedule/
* [ ] Ensure that the whole schedule is exported including areas that might be off-screen

![screenshot from 2017-03-04 14-43-40](https://cloud.githubusercontent.com/assets/1583873/23576750/2f74ebb6-00e9-11e7-8137-c1db6919a12b.png)

poush:
I tried creating a solution. Here is the sample export I'm able to download 
![timeline 88](https://cloud.githubusercontent.com/assets/1812082/23589628/7804771c-01f6-11e7-9b7e-cf8850131c67.png)

poush:
The concept is that I give the timeline full width for a moment to generate export 

![screenflow](https://cloud.githubusercontent.com/assets/1812082/23589660/262eb276-01f7-11e7-98a2-48b5c6ba746f.gif)

poush:
We can fade the screen with some popup or modal saying your export is in the process for this moment of the export process. 
@mariobehling @niranjan94 @SaptakS Would like to know your suggestions on this.

mariobehling:
@poush Thanks, please make a PR and iterate from there to see if the fade solution is more elegant.

mariobehling:
@poush Thanks for implementing this. In order to be fully functional the complete width of all tracks needs to be exported. The currently implemented version only exports the visible area. Please see below.

![timeline](https://cloud.githubusercontent.com/assets/1583873/23697338/604d5a16-0422-11e7-9038-486a5bcd211b.png)

mariobehling:
The "Export PNG" button should be located behind the "Sessions View" button (on the right of it).

Testing the full schedule view also only some areas of the schedule are exported as png.

poush:
@mariobehling Can you tell me how many rooms were there in that schedule?

poush:
If this was the schedule,

![timeline](https://cloud.githubusercontent.com/assets/1812082/23735277/981e080e-04aa-11e7-8a8f-7d45e5a7a524.png)

mariobehling:
@poush It is working now. Awesome!

I still see open issues on your screenshot though still:
a) Background is partly transparent. It should be white
[...]

PR #3369, opened 2017-03-12
export button shifted, whitespace reduced from export

poush:
Improves: #3295


mariobehling:
Thanks. Please provide a screenshot. Did you substitute the "PNG Export" with a download button, e.g. something like that, but maybe gray or another color of the existing design:
unnamed

poush:
@mariobehling Yes it is something similar to your suggestion. 
screen shot 2017-03-12 at 1 49 20 pm

Issue #3364, opened 2017-03-11
Discount works on free tickets


SaptakS:
Discount works on free tickets making the amount negative.
![screenshot from 2017-03-11 23-12-29](https://cloud.githubusercontent.com/assets/9530293/23825475/5c8be042-06b0-11e7-8ea9-4e1729ca3524.png)

poush:
Would like to propose a fix.

PR #3370, opened 2017-03-12
prevent discounts to set price to negative

poush:
Fix: #3364


mariobehling:
@SaptakS Yes, that is correct.

I opened a follow up issue here: https://github.com/fossasia/open-event-orga-server/issues/3376

Let's iterate step by step.

PR #3372, opened 2017-03-12
Lightbox takes entire width and height

SaptakS:
Fixes https://github.com/fossasia/open-event-orga-server/issues/3273#issuecomment-284416515

screenshot from 2017-03-12 15-11-32


mariobehling:
Very good.

PR #3373, opened 2017-03-12
time iterations made to 15minutes

SaptakS:
Fixes https://github.com/fossasia/open-event-orga-server/issues/3318

screenshot from 2017-03-12 15-19-45

@mariobehling @niranjan94 please review. @mariobehling it shows two rows below because the session ends at 11.45 and 11.45 + 15minutes = 12.00. So, it shows rows belonging to both 12.00 and 11.45.


mariobehling:
Good job!

PR #3395, opened 2017-03-16
Fixes #3383 : microlocation name break added for long names

### LINKED ISSUE #3383: Public Calendar Schedule: Add line break if room name is longer than cell
On the table of the schedule, if the room name is longer than the width of the cell, parts of the name are hidden or overlapping. Please ensure there are automatic line breaks in this case, e.g. here: https://eventyay.com/e/45da88b7/schedule/

screenshot from 2017-03-13 11-03-11

SaptakS:
In header, the name for microlocation applied with line break.

### DIFF
--- a/app/static/css/events/scheduler.css
+++ b/app/static/css/events/scheduler.css
@@ -139,7 +139,9 @@
 
 .timeline .microlocations.x1 .microlocation-header {
     height: 47px;
-    line-height: 47px;
+    display: flex;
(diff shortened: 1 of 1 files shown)


mariobehling:
Looks like height is hard-coded. It is works ok. Let's do an implementation that adapts to the content later.

mariobehling:
Ah, I see you took it out. good.

Issue #3399, opened 2017-03-16
Downloading Tickets have name "test"


mariobehling:
Tickets that are download from the system have the name "test". Expected: A ticket should have the name "EVENTNAME-Ticket.pdf".

LuD1161:
@mariobehling 
Working on this

PR #3400, opened 2017-03-16
Fix : Issue #3399 Downloading Tickets have name "test"

LuD1161:
Changed the Downloaded Filename


rctay:
I have some reservations. As `name` field of Events may not map well to filename requirements nor HTTP headers.

rctay:
We will just relook all these callsites in the future.

PR #3403, opened 2017-03-17
Show running events on live page

### LINKED ISSUE #3402: FOSSASIA Summit no longer on start page event though event is still running
Please ensure events that are still running are listed on start page until end of event, e.g. check https://eventyay.com

Princu7:
Fixes #3402 
* Filter the events on basis of their end time rather than their start date

* Screenshot
screenshot from 2017-03-17 21-15-54

Here the event 'Princu' has start time of Mar 17, 8 AM and is showing on the front page.

@mariobehling @niranjan94 @SaptakS Please review. Thanks

### DIFF
--- a/app/helpers/data_getter.py
+++ b/app/helpers/data_getter.py
@@ -417,8 +417,7 @@ class DataGetter(object):
             events = Event.query.filter(Event.state == 'Published')
         else:
             events = Event.query.filter(Event.state == 'Published').filter(Event.privacy != 'private')
-        events = events.filter(Event.start_time >= datetime.datetime.now()).filter(
-            Event.end_time >= datetime.datetime.now()).filter(Event.in_trash == 'False')
+        events = events.filter(Event.end_time >= datetime.datetime.now()).filter(Event.in_trash == 'False')
         return events
 
     @staticmethod


rctay:
I can see why the old code was wrong. We were taking the intersection when it should be the union. So I think this fix is incorrect.

Princu7:
@rctay Thanks for the review. I am sorry but I didn't get what you were saying. The original query was

`events = events.filter(Event.start_time >= datetime.datetime.now()).filter(Event.end_time >= datetime.datetime.now()).filter(Event.in_trash == 'False')`

Here there is no need of the additional filter `event.end_time >= datetime.datetime.now()` since `event.start_time <= event.end_time` is always true. So the above query is just equivalent to

`events.filter(Event.start_time >= datetime.datetime.now())`

I just changed it to
`events.filter(Event.end_time >= datetime.datetime.now())`

Please clarify. Thanks!!

niranjan94:
@rctay I have to agree with @Princu7 on this one ...we need all current and future event... And IMO, the easiest way to do that is to get all are ending now or after now. No union necessary as there is only a single query needed.

PR #3410, opened 2017-03-19
quick_fixed_doc

raj-maurya:
@SaptakS 
Please review! Thanks!


niranjan94:
@raj-maurya thanks for the fix 😄

PR #3411, opened 2017-03-19
guest/event/cfs: only show sessions from current event 

rctay:
While we're at it, bump up templating stuff into view code, and improve `<hr>` printing, so we don't print `<hr>` when there's nothing to separate, eg: (notice the hr's but nothing follows the hr:

image

### DIFF
--- a/app/templates/gentelella/guest/event/cfs.html
+++ b/app/templates/gentelella/guest/event/cfs.html
@@ -53,7 +53,7 @@
                             </div>
                         </div>
                     </div>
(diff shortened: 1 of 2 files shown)


rctay:
Oops it seems I might have introduced a whitespace change (although it was the right one). Fixing.

rctay:
Merged in b71374c00eedc4049278f34667525069843c0e42.

niranjan94:
@rctay let's not merge things without atleast one code-review ... :)

PR #3421, opened 2017-03-23
Register as user and session form fixed

### LINKED ISSUE #3420: 'Register as a Speaker' takes back to the same pop up
Even on registering as speaker, user is not registered. After registration, it takes back to the same "Register as Speaker" and "Propose Session" modal. 
That is:
screenshot from 2017-03-24 03-52-33

And on clicking "Propose session", it takes to speaker_session form rather than only session form.

magdalenesuo:
fixes #3420

### DIFF
--- a/app/templates/gentelella/guest/event/cfs.html
+++ b/app/templates/gentelella/guest/event/cfs.html
@@ -53,7 +53,7 @@
                             </div>
                         </div>
                     </div>
-                    {% if not existing_sessions[0] %}
+                    {% if not user_speaker[0] %}
                         <div id="speaker-session-modal" class="modal fade" tabindex="-1" role="dialog">
                             <div class="modal-dialog">
                                 <div class="modal-content">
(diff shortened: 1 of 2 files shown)


app/templates/gentelella/guest/event/cfs.html
```diff
@@ -69,30 +69,52 @@ <h4 class="modal-title">{{ _("Submit Proposal") }}? </h4>
...
                         </div>
                     {% else %}
-                        <div id="speaker-session-modal" class="modal fade" tabindex="-1" role="dialog">
-                            <div class="modal-dialog">
-                                <div class="modal-content">
-                                    <div class="modal-header">
-                                        <button type="button" class="close" data-dismiss="modal" aria-label="Close"></button>
-                                        <h4 class="modal-title">{{ _("Add Speaker Information") }} </h4>
+                        {% if not existing_sessions[0] %}
+                                <div id="speaker-session-modal" class="modal fade" tabindex="-1" role="dialog">
```
SaptakS: Fix indentation.

Issue #3335, opened 2017-03-08
Schedule: How can we get a PDF print out version from the browser


mariobehling:
The browser does not generate a proper pdf print-out version of the calendar schedule. 

https://eventyay.com/e/45da88b7/schedule/

Please implement a pdf.

cepradeep:
One option is to add `@media print` CSS styles, but I'm not sure how much improvement this will fetch us. It's worth a try, though.

rctay:
See #3398.

mariobehling:
Still pending full implementation.

niranjan94:
@mariobehling then it can be taken forward in issue #3615

No need to have two open issues for the same bug.

mariobehling:
Ok, sorry. Let me be more clear. As css was mentioned in this issue, let's follow up on the pdf version that requires a specific css in this issue. Made the title clearer for this.

PR #3443, opened 2017-03-30
CSS styles for schedule printing

rctay:
vis-a-vis #3335. We generate  a bootstrap that does not clear print styles (print.less), as it was clearing things like horizontal rules in the schedule timeline.

Preview: [FOSSASIA Summit - eventyay.pdf](https://github.com/fossasia/open-event-orga-server/files/888452/FOSSASIA.Summit.-.eventyay.pdf)

### DIFF
--- a/app/static/css/events/scheduler.css
+++ b/app/static/css/events/scheduler.css
@@ -65,6 +65,10 @@
 
 #timeline {
     background-color: #FFFFFF;
+    margin-top: 10px;
+    margin-left: 10px;
+    width: 100%;
+    overflow-x: hidden;
 }
 
 .timeline-table td {
--- a/app/static/css/guest/schedule.css
+++ b/app/static/css/guest/schedule.css
@@ -110,12 +110,58 @@ div {
         display: none;
     }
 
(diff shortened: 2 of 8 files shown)


rctay:
Excluding generated bootstrap.

rctay:
Oops, forgot to amend commit message. Done.

SaptakS:
@rctay can you please explain what you are trying to do and how this is going to help in getting print out of schedule?

rctay:
I am trying to get print per #3335, but some styles which are useful in the printout like the horizontal bars are getting cleared out by bootstrap's "clear print" rules. I have experimented with adding them back by hand but there are far too many styles getting cleared so it becomes a hodgepodge of `!important` rules. You can have a look at the approach I am taking in https://github.com/rctay/open-event-orga-server/tree/guest-schedule-print .

Hence I am generating a bootstrap css without "clear print" rules, which bootstrap includes by default via the `@import "print.less"`.

mariobehling:
@rctay @niranjan94 Should we merge this and iterate from here? It is already an improvement.

rctay:
Updated to fix a merge conflict due to activity on `development` branch.

PR #3445, opened 2017-04-01
Add licence_details field in the api

### LINKED ISSUE #3238: Wrong Information in REST API for FOSSASIA 2017 Summit
API end point : `https://eventyay.com/api/v1/events/6`

 - [ ]  Correct the copyright licence from `Attribution` to `Creative Commons Attribution License.`
      Issue : https://github.com/fossasia/open-event-webapp/issues/1063
 - [x]  Correct the Ticket URL from `https://eventyay.com/e/45da88b7/tickets/` to `https://eventyay.com/e/45da88b7/` 
     Issue: https://github.com/fossasia/open-event-webapp/issues/1067
@niranjan94 @SaptakS  Please help regarding this.

shubham-padia:
fixes #3238.

### DIFF
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -23,6 +23,7 @@ from app.api.helpers.helpers import requires_auth, parse_args, \
 from app.api.helpers.utils import PAGINATED_MODEL, PaginatedResourceBase, \
     PAGE_PARAMS, POST_RESPONSES, PUT_RESPONSES, BaseDAO, ServiceDAO
 from app.api.helpers.utils import Resource, ETAG_HEADER_DEFN
+from app.api.helpers.custom_fields import Licence
 
 api = Namespace('events', description='Events')
 
@@ -58,6 +59,7 @@ SOCIAL_LINK = api.model('SocialLink', {
     'link': fields.String(required=True)
 })
 
+
 SOCIAL_LINK_POST = api.clone('SocialLinkPost', SOCIAL_LINK)
 del SOCIAL_LINK_POST['id']
 
@@ -86,6 +88,7 @@ EVENT = api.model('Event', {
     'privacy': EventPrivacyField(),
     'ticket_url': fields.Uri(),
     'copyright': fields.Nested(EVENT_COPYRIGHT, allow_null=True),
+    'licence_details': Licence(attribute='copyright.licence', allow_null=True),
     'schedule_published_on': fields.DateTime(),
     'code_of_conduct': fields.String(),
     'social_links': fields.List(fields.Nested(SOCIAL_LINK), attribute='social_link'),
@@ -113,7 +116,7 @@ EVENT_POST = api.clone('EventPost', EVENT)
 del EVENT_POST['id']
 del EVENT_POST['social_links']
 del EVENT_POST['version']
-
+del EVENT_POST['licence_details']
 
 # ###################
 # Data Access Objects
--- a/app/api/helpers/custom_fields.py
+++ b/app/api/helpers/custom_fields.py
@@ -3,6 +3,7 @@ import colour
 from datetime import datetime
 from flask import request
 from flask.ext.restplus.fields import Raw, Nested, List
+from app.helpers.data_getter import DataGetter
 
 EMAIL_REGEX = re.compile(r'\S+@\S+\.\S+')
 URI_REGEX = re.compile(r'(http|https|ftp)://\S*(\S+\.|localhost(\:\d+)?/)\S+')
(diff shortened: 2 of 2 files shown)


app/api/events.py
```diff
@@ -23,9 +24,24 @@
 from app.api.helpers.utils import PAGINATED_MODEL, PaginatedResourceBase, \
     PAGE_PARAMS, POST_RESPONSES, PUT_RESPONSES, BaseDAO, ServiceDAO
 from app.api.helpers.utils import Resource, ETAG_HEADER_DEFN
+from app.helpers.static import EVENT_LICENCES
 
 api = Namespace('events', description='Events')
 
+
+class LicenceClass(fields.Raw):
```
aviaryan: Please don't include this class here. Use custom_fields.py

shubham-padia: Thanks !!

app/api/helpers/custom_fields.py
```diff
@@ -245,3 +246,15 @@ def validate(self, value):
             self.validation_error = 'Value of %s is not in available choices'
             return False
         return True
+
+
+class LicenceClass(CustomField):
```
aviaryan: It would be better if naming convention is followed here. Eg > All other classes in that file have names like String, Color, Boolean .. so I think you can make the class name as just 'Licence'. What do you think?

shubham-padia: :+1:

app/api/helpers/custom_fields.py
```diff
@@ -245,3 +246,15 @@ def validate(self, value):
             self.validation_error = 'Value of %s is not in available choices'
             return False
         return True
+
+
+class Licence(CustomField):
+    """
+    Custom field that gets the licence details given the licence name
+    """
+    __schema_example__ = "{'name': 'Licence name' 'long_name': 'Long licence name', \
```
aviaryan: I am not sure about using a dictionary as a custom field. Since you are base-classing CustomField, the `__schema_type__` is set to string. So you won't be able to pass something like `"license": {"f": "abc", "z": "aa"}` in API. Only something like `"license": "string_on_the_right_side"` will work.

shubham-padia: Please review the new changes  :) .

aviaryan: @shubham-padia Can you please explain me about #3238 which this PR fixes.
The title is "Wrong Information in REST API for FOSSASIA 2017 Summit" so it can be fixed by just updating the copyright field of the event, why is there a need to add a new field called "license" here?

shubham-padia: The webapp requires the `long_name` field of the licence right now. But as we are using the db model of copyright in the api, on `long_name` field is present there. Also right now, the webapp requires the `long_name` of the license, in future other details about the license may also be required, better to provide all the details in the api :) .

shubham-padia:
Travis failing. Having a look at it . Do not merge please.

tests/unittests/api/utils_post_data.py
```diff
@@ -34,6 +34,7 @@
         "year": 2016,
         "logo": ""
     },
+    'licence_details': {},
```
aviaryan: I don't think this needs to be there in POST API schema as it is an auto-generated field.

shubham-padia:
Please review again. Travis passing now. Thanks for the inputs @aviaryan

app/api/helpers/custom_fields.py
```diff
@@ -245,3 +246,20 @@ def validate(self, value):
...
+    __schema_example__ = {
+        'name': 'Licence name',
+        'long_name': 'Long licence name',
+        'description': 'Description about the licence',
+        'licence_url': 'http://example.com',
+        'licence_logo': 'http://example.com/example.jpg',
+        'licence_compact_logo': 'http://example.com/example.jpg'
+    }
+
+    def format(self, value):
```
aviaryan: What happens if `value` comes in as `None` ?

aviaryan:
@shubham-padia My pleasure 😄. Btw, I left a small comment. Please see.

app/api/helpers/custom_fields.py
shubham-padia: The `get_licence_details` function takes care of that.

tests/unittests/api/utils_post_data.py
```diff
@@ -34,6 +34,14 @@
         "year": 2016,
         "logo": ""
     },
+    'licence_details': {
```
aviaryan: Just like `social_links` and `version`, this field is also not included in the POST API schema so I don't think it is needed to be tested here.

PR #3447, opened 2017-04-02
CSS styles for schedule printing

rctay:
vis-a-vis #3335.

Requires #3443.

Preview: [FOSSASIA Summit - eventyay.pdf](https://github.com/fossasia/open-event-orga-server/files/888452/FOSSASIA.Summit.-.eventyay.pdf)

### DIFF
--- a/app/static/css/events/scheduler.css
+++ b/app/static/css/events/scheduler.css
@@ -65,6 +65,10 @@
 
 #timeline {
     background-color: #FFFFFF;
+    margin-top: 10px;
+    margin-left: 10px;
+    width: 100%;
+    overflow-x: hidden;
 }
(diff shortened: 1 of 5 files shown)


mariobehling:
Great! Is there any way to define landscape mode and set the PDF to A3? The goal would be to get a print out on one page.

SaptakS:
@rctay what happens if the number of microlocation exceeds the width, i.e., there is a horizontal scroll. Can you provide an example pdf exported for that? Thanks.

niranjan94:
@rctay since all this comes under a single enhancement, let it all be under one PR please ... Please merge this one along with the other PR (#3443)

PR #3449, opened 2017-04-03
Separate session and speaker forms in mysession

### LINKED ISSUE #3408: Follow speakers proposal logic in "My Sessions" tab
When a speaker submits a proposal there are two parts: a) Speakers Form b) Sessions Form

Once the session is submitted and the speakers goes to my sessions (e.g. https://eventyay.com/events/mysessions/), s/he only see a link to both combined. 

Please implement the same logic: Show each event with speakers form, submitted sessions and their status. Each form should have a separated link.

screenshot from 2017-03-19 11-33-22

SaptakS:
Fixes #3408 

Follow speaker's proposal logic in MySession

screenshot from 2017-04-03 22-17-07

@mariobehling please check.


mariobehling:
I am unclear what happens if there are several proposals for the same event. Cannot see it in the screenshot.

SaptakS:
@mariobehling firstly we get the list of session like we get right now. After we go to the detailed display page, we get this. So it is specific to session only.

Issue #3454, opened 2017-04-06
Call for Speakers (and potentially other pages): Links on right side bar, that should not be there


mariobehling:
The right side bar on the Call for Speakers pages there are links, where there should not be any links.

Please ensure there is nothing on the right side bar on
* Call for Speakers Page
* Submission Pages (Speaker and Proposal Submission)
* All Schedule Pages

![screenshot from 2017-04-06 14-27-15](https://cloud.githubusercontent.com/assets/1583873/24754140/81c887dc-1ad5-11e7-9720-028b17d2d05e.png)

PR #3473, opened 2017-04-08
removed right side links on cfs and submission pages

magdalenesuo:
fixes: #3454

### DIFF
--- a/app/templates/gentelella/guest/event/base.html
+++ b/app/templates/gentelella/guest/event/base.html
@@ -182,61 +182,62 @@
             {% block content %}
             {% endblock %}
         </div>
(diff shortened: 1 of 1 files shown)


app/templates/gentelella/guest/event/base.html
```diff
@@ -183,60 +183,63 @@ <h4 style="font-weight: 300;"><i class="glyphicon glyphicon-map-marker"></i> {{
             {% endblock %}
         </div>
         {% if active_page != 'sessions' and active_page != 'schedule' %}
```
SaptakS: Why not just add `active_page != 'cfs'` ?

PR #3475, opened 2017-04-09
Removal of Withdraw Proposal option

### LINKED ISSUE #3471: My Sessions Page: Do not show "Withdraw Proposal" once event is over
Speaker should not be able to withdraw a proposal once the event is over.

screenshot from 2017-04-08 15-15-08

Himanshi-Khandelwal:
Comparison was done between the session end date and the current date so as to decide the whether it is a past/present/future session which resolves #3471

### DIFF
--- a/app/templates/gentelella/users/mysessions/mysession_detail.html
+++ b/app/templates/gentelella/users/mysessions/mysession_detail.html
@@ -32,10 +32,12 @@
                     {% endif %}
                     <div class="btn-group pull-right">
                         {% if session.state == 'pending' or session.state == 'accepted' or session.state == 'confirmed' or session.state == 'rejected' %}
+                            {% if datetime_now <= session.end_time %}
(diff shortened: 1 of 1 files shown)


app/views/users/my_sessions.py
```diff
@@ -61,6 +61,8 @@ def display_session_view(session_id):
     return render_template('gentelella/users/mysessions/mysession_detail.html',
                            session=session,
                            speaker_form=speaker_form,
+                           end_time=session.end_time,
```
niranjan94: Why do you need to pass these two unnecessarily ? Endtime can be accessed with `session.end_time` in the template and the current time can be accessed with `datetime_now`

Himanshi-Khandelwal: Ok , I 'll do that way..

Himanshi-Khandelwal:
Please review.

PR #3477, opened 2017-04-09
manage event: scale logo and column ratios correctly

### LINKED ISSUE #3091: Event Dashboard: Adjust column width, event info/functions and add event logo
The event dashboard is using a very basic style and should be enhanced. The main idea of the event rows is a) to provide a link to the event itself and b) to give an overview.

To make it easier for the user according to these goals, please implement the following:

* [x] Add Event logo in a large rectangle on the left of the event and the event name below
* [x] Adjust the column widths to use the space more efficiently
* [x] Reduce the width of the last column

screenshot from 2017-01-30 20-16-37

shubham-padia:
fixes #3091 
selection_152

### DIFF
--- a/app/static/js/events/event-dashboard.js
+++ b/app/static/js/events/event-dashboard.js
@@ -10,10 +10,11 @@ function initDataTable(element) {
             "columnDefs": [
                 { "width": "7%", "targets": 0 },
                 { "width": "15%", "targets": 1 },
-                { "width": "23%", "targets": 2 },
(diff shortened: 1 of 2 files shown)


mariobehling:
Thanks. Please also help to adjust the column width to match the content better, e.g. the last column has unused space, which we can get rid of while other columns are squeezed.

shubham-padia:
Thanks !! updated it .
selection_153

PR #3490, opened 2017-04-14
db naming schema refactor: access_codes and discount_codes

### LINKED ISSUE #3489: Refactor access_codes and discount_codes according to the new naming schema
Parent issue: #3480

shubham-padia:
resolves #3489 .


shubham-padia:
@SaptakS One downside with a big refactor is it will be hard to review and as we don't have extensive tests for every functionality, many things may break during the refactor :-) . We'd have to decide b/w no. of commits and ease of review.

aviaryan:
@shubham-padia @SaptakS @niranjan94 
When renaming database columns, test cases that make affected code run are sufficient to ensure that renaming was correct. Like if `discount_code.valid_from` was renamed to `discount_code.valid_from_date_time`, then if all the lines where this change has happened have been covered in one test case or the another, then everything is fine. 
Codecov reports can be used to know which line has been tested and which not.
I think what we can do is add tests for all the untested lines in the PR too. 
That way, we are doing 2 things; we are improving testing of our application + making sure that PR won't cause any additional bugs in the future. 
When testing has been done correctly, the Codecov report comment will show "The diff coverage is 100%." (instead of the 52% now). 

If we are not in a hurry to do this change, I think we should go with this. Views??

niranjan94:
Wait ... Looking at this ... I feel for datetime fields, `<name>_at`/`<name>_from`/`<name>_till` would be better and shorter than `<name>_date_time`/`<name>_time` and easier to understand too 

Example, `start_at`, `end_at`, `created_at`,`valid_from`, `valid_till`

@SaptakS @shubham-padia @aviaryan

shubham-padia:
@niranjan94 @SaptakS sounds good. Closing this for now. Discussion can be continued on the parent issue itself.

PR #3496, opened 2017-04-14
Change logic for soft deletes

### LINKED ISSUE #3494: Change logic for soft deletes
In the current implementation, for soft-deletes we have two fields `in_trash` and `trash_date` (Example: see `app/models/event.py`) 

Instead, have one nullable datetime field called `deleted_at`. That field should have `null` as the default. If that field is null, then the record is not deleted. If it isn't null, then it is deleted. 

_(This is similar to the soft-delete implementation done in [Laravel Framework](https://laravel.com/docs/5.4/eloquent#soft-deleting))_

aviaryan:
Fixes #3494 

#### Notes

* To check for None in SQLAlchemy's filter functions, we use is_(None) and isnot(None) respectively. ([Docs](http://docs.sqlalchemy.org/en/rel_1_0/core/sqlelement.html#sqlalchemy.sql.operators.ColumnOperators.is_))
* To check for None in Jinja, we have [none](http://jinja.pocoo.org/docs/2.9/templates/#none) (small caps).


aviaryan:
@niranjan94 Earlier we had the trash_date field which behaved exactly same as deleted_at.. This PR has just renamed that field. So the old trash information is contained in that trash_date field which has been renamed to deleted_at. Please see [trash_user](https://github.com/fossasia/open-event-orga-server/pull/3496/files#diff-aa9d7cf132920541a1c3195a7c7961fbR1034), trash_event and trash_session method in the diff.

PR #3513, opened 2017-04-18
Pdf download of ticket order fixed for large ticket names

### LINKED ISSUE #3503: Eventyay Order Details: Download of tickets don't work.
The ticket download doesn't work for OSCAL event in eventyay.

SaptakS:
Fixes #3503 

@mariobehling this will solve the issue that was being faced in OSCAL for the supporter ticket. The issue was occurring because the length of the name of the ticket was overflowing the width, and the width of the frame in the pdf generation code wasn't correctly mentioned.


SaptakS:
For later references, this is the link I refered from: https://github.com/xhtml2pdf/xhtml2pdf/blob/master/doc/source/format_html.rst

PR #3514, opened 2017-04-19
Add placeholder_url to events API, fixes #3500

### LINKED ISSUE #3500: API & Placeholder images
In our new API-Centric approach, how do we handle placeholder images (based on category/topic/sub-topic) ? 

1. Does the API response for an event itself returns a placeholder image URL in the place of the background image if the background image doesn't exist ?
2. Or, does the API response for an event return `null` for a non-existent image and we have a separate endpoint that provides all the placeholder images for each category and the frontend can use both to display the correct placeholder.

@mariobehling @aviaryan @SaptakS @shubham-padia @magdalenesuo your views ?

aviaryan:
Fixes #3500

### DIFF
--- a/app/api/helpers/custom_fields.py
+++ b/app/api/helpers/custom_fields.py
@@ -272,11 +272,4 @@ class PlaceHolder(CustomField):
     __schema_example__ = 'http://url.com/img.ext'
 
     def format(self, value):
-        custom_placeholder = DataGetter.get_custom_placeholder_by_name(value.sub_topic or value.topic or 'Other')
-        if custom_placeholder:
-            return custom_placeholder.url
-        else:
-            default_images = DataGetter.get_event_default_images()
-            for key in [value.sub_topic, value.topic, 'Other']:
-                if key in default_images:
-                    return request.url_root.strip('/') + '/static/placeholders/' + default_images[key]
+        return DataGetter.get_placeholder_url_by_event(value)
--- a/app/api/helpers/custom_fields.py
+++ b/app/api/helpers/custom_fields.py
@@ -263,3 +263,20 @@ class Licence(Raw):
 
     def format(self, value):
         return DataGetter.get_licence_details(value)
+
+
+class PlaceHolder(CustomField):
+    """
+    Background Image PlaceHolder field for Events
+    """
+    __schema_example__ = 'http://url.com/img.ext'
+
+    def format(self, value):
+        custom_placeholder = DataGetter.get_custom_placeholder_by_name(value.sub_topic or value.topic or 'Other')
+        if custom_placeholder:
+            return custom_placeholder.url
+        else:
+            default_images = DataGetter.get_event_default_images()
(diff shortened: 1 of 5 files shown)


niranjan94:
@aviaryan 

While we are at it .. Can we put these fields `thumbnail`, `large` and `background_image` into a nested object ? 

Something like,

```json
{
    ... other event object keys ...

    "background_image": {
             "thumbnail_image_url": "url for that",
             "large_image_url": "url for that",
             "original_image_url": "url for that"
    }
}
```

If there is no background image available `background_image` will be `null`

aviaryan:
@niranjan94 Ok, but lets do that as a separate PR.

niranjan94:
@aviaryan cool. Please fix travis fail

app/api/helpers/custom_fields.py
```diff
@@ -263,3 +263,20 @@ class Licence(Raw):
...
+
+class PlaceHolder(CustomField):
+    """
+    Background Image PlaceHolder field for Events
+    """
+    __schema_example__ = 'http://url.com/img.ext'
+
+    def format(self, value):
+        custom_placeholder = DataGetter.get_custom_placeholder_by_name(value.sub_topic or value.topic or 'Other')
+        if custom_placeholder:
```
shubham-padia: It'd great if you can move this logic to a method
```
    def get_placeholder_by_event(event):
        custom_placeholder = DataGetter.get_custom_placeholder_by_name(event.sub_topic or event.topic or 'Other')
        if custom_placeholder:
            return custom_placeholder.url
        else:
            default_images = DataGetter.get_event_default_images()
            return url_for('static', filename='placeholders' + default_images[event.sub_topic or event.topic or 'Other'])
```
in data_getter.py so this can be reused whenever needed at other places

aviaryan: Nice catch. Not sure why I didn't do it before 😅

aviaryan: Btw, the following won't work as you think it will. 

```
default_images[event.sub_topic or event.topic or 'Other']
```

aviaryan:
Please review again.

app/helpers/data_getter.py
```diff
@@ -592,6 +592,17 @@ def get_event_default_images():
...
     @staticmethod
+    def get_placeholder_url_by_event(event):
+        custom_placeholder = DataGetter.get_custom_placeholder_by_name(event.sub_topic or event.topic or 'Other')
+        if custom_placeholder:
+            return custom_placeholder.url
+        else:
+            default_images = DataGetter.get_event_default_images()
+            for key in [event.sub_topic, event.topic, 'Other']:
+                if key in default_images:
+                    return request.url_root.strip('/') + url_for('static', filename='placeholders/' + default_images[key])
```
shubham-padia: please use `url_for` here, rest seems good to go

shubham-padia:
Thanks !!

PR #3527, opened 2017-04-23
Added backend and vue model

### LINKED ISSUE #2853: Let users control order of tickets
The tickets currently show up in the order entered by the user. The organizer might want another order though.

Please implement a way to change the order of the tickets in the ticket column, e.g. implement up and down arrows.

screenshot from 2017-01-07 11-54-05

### LINKED ISSUE #3532: Missing migration for removing `creator_id` field

SaptakS:
Fixes #2853 
Fixes #3532 

screenshot from 2017-04-25 16-41-05
screenshot from 2017-04-25 16-41-09

After reordering:

screenshot from 2017-04-25 16-41-16
screenshot from 2017-04-25 16-41-24


migrations/versions/d763cf2e7dc4_.py
```diff
@@ -0,0 +1,31 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(u'events_creator_id_fkey', 'events', type_='foreignkey')
+    op.drop_column('events', 'creator_id')
```
shubham-padia: @SaptakS Looks like this column doesn't exist atm, can you please recheck?

aviaryan: I have faced a similar issue as well when creating migrations for one of my PRs. Seems like `creator_id` field was deleted from the model but no migration was created to remove it.

shubham-padia: Thanks !! Opened and linked the issue to this.

SaptakS: @shubham-padia if the `creator_id` isn't there, what is the issue about? I don't understand what you are saying. The mistake is in this PR you are saying, right?

shubham-padia: The issue initially was that if the `creator_id` column was not there at the first place, why was it being dropped in the migration. But it seems like `creator_id` field was deleted from the model but no migration file was generated for the same. This issue is resolved.

shubham-padia: Resolved :white_check_mark:

mariobehling:
@SaptakS It seems you are mixing two different issues into one PR. This is not according to our best practices and might make it difficult for reviewers. Unfortunately the ticket sort order does not work on eventyay on the public page.

mariobehling:
My problem in this PR is that it is hard to find. The commit message does not state that this PR is about the ticket ordering.

SaptakS:
@mariobehling two different issues weren't mixed into one PR. It is just that one of the issues created got automatically solved by this PR. So I marked that here as well. @niranjan94 and @shubham-padia already knows that. You can check the discussion in the issues.

mariobehling:
Ok, thanks you.

PR #3529, opened 2017-04-23
Remove robot tests

niranjan94:
Since we'll be moving to a new frontend, robot tests here are actually pointless. We'll be having acceptance, integration and unit tests in the frontend.


aviaryan:
Don't we need to remove *.robot files?

SaptakS:
@aviaryan we can keep them for now, just in case we decide to stick with this instead of the new front-end. So I think keeping the *.robot files for now and just removing the test is a better option. What do you think?

niranjan94:
Actually ... I forgot about those files ... Let's remove them too ... they don't actually do anything useful ... We can always get them back from the history

SaptakS:
@niranjan94 well that's true too.. :sweat_smile:

Issue #3528, opened 2017-04-23
Remove eventlet dependency


SaptakS:
We are using eventlet only in notifications to make it real time which isn't essential right now. Eventlet breaks the docker integration and hence the kubernetes deployment. https://github.com/eventlet/eventlet/issues/401.

Since it is not essential to our system, we can remove the dependency on eventlet. Otherwise this breaks the eventyay deployment.

niranjan94:
We can re-implement web-sockets later when we have more use for them.

temoto:
Please try new eventlet version. Hopefully it fixes this problem

`pip install https://github.com/eventlet/eventlet/archive/b756447bab51046dfc6f1e0e299cc997ab343701.zip`

SaptakS:
Thanks @temoto for the fix. We have changed the architecture of the project some what so I am not sure we will be needing eventlet anymore but will surely check out. 🍪

PR #3530, opened 2017-04-23
Eventlet disabled by default in production server

SaptakS:
#3528 

For the time it disables socket io integration by default instead of turning it on by default. Disabled eventlet as of now.


niranjan94:
Let's test this on staging. Then if it fixes our issue, we'll move this.

PR #3568, opened 2017-05-05
Implemented font color support in front-end and back-end

### LINKED ISSUE #3554: Wizard Step 3 - Session Colors: Show preview and automatically choose black or white font color
Currently we only have standard font color white or black in the list and calendar views of the schedule.

The font color should be chosen automatically depending on the contrast. Please implement the following:
* behind each line implement a preview area with the background and automatically chosen font color a) black or b) white
* implement the font color in the schedule view
* add the font color choice to the API and JSON

screenshot from 2017-04-28 11-19-12
Show a small preview on the right here.

There is a slack discussion on this feature here http://stackoverflow.com/questions/1855884/determine-font-color-based-on-background-color/1855903

mejariamol:
fixes #3554 

screenshot from 2017-05-06 01-42-06
screenshot from 2017-05-06 01-44-58

### DIFF
--- a/app/api/sessions.py
+++ b/app/api/sessions.py
@@ -31,6 +31,7 @@ SESSION_TRACK = api.model('SessionTrack', {
     'id': fields.Integer(required=True),
     'name': fields.String(),
     'color': fields.Color(),
+    'font_color': fields.Color(),
 })
 
 SESSION_SPEAKER = api.model('SessionSpeaker', {
--- a/app/api/tracks.py
+++ b/app/api/tracks.py
@@ -24,6 +24,7 @@ TRACK = api.model('Track', {
     'name': fields.String(required=True),
     'description': fields.String(),
     'color': fields.Color(required=True),
+    'font_color': fields.Color(),
     'track_image_url': fields.Upload(),
     'location': fields.String(),
     'sessions': fields.List(fields.Nested(TRACK_SESSION)),
@@ -36,6 +37,7 @@ TRACK_PAGINATED = api.clone('TrackPaginated', PAGINATED_MODEL, {
 TRACK_POST = api.clone('TrackPost', TRACK)
 del TRACK_POST['id']
 del TRACK_POST['sessions']
+del TRACK_POST['font_color']
 
 
 # Create DAO
--- a/app/models/track.py
+++ b/app/models/track.py
@@ -38,9 +38,16 @@ class Track(db.Model):
         return self.name
 
     @property
+    def font_color(self):
+        h = self.color.lstrip('#')
+        a = 1 - (0.299 * int(h[0:2], 16) + 0.587 * int(h[2:4], 16) + 0.114 * int(h[4:6], 16))/255
+        return '#000000' if (a < 0.5) else '#ffffff'
+
+    @property
     def serialize(self):
         return {
             'id': self.id,
             'name': self.name,
-            'color': self.color
+            'color': self.color,
+            'font_color': self.font_color
         }
--- a/app/static/js/events/wizard/sessions-speakers/vue-components.js
+++ b/app/static/js/events/wizard/sessions-speakers/vue-components.js
@@ -1,6 +1,15 @@
 Vue.component('session-track', {
     props: ['track'],
     template: '#track-template',
+    watch: {
+        'track.color': function(value) {
+            var result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(value);
+            if(result) {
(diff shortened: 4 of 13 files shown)


app/templates/gentelella/users/events/wizard/components/tracks-microlocations-sessions.html
```diff
@@ -1,32 +1,35 @@
...
-        <div class=" col-xs-5 form-group">
+        <div class=" col-xs-4 form-group">
             <input type="text" class="form-control" v-model="track.name" placeholder="Name" required="required">
         </div>
-        <div class="col-xs-7 input-group form-group">
+        <div class="col-xs-4 input-group form-group" style="float:left;">
             <div class="input-group colorpicker-component" id="color">
                 <input type="text" class="form-control color-picker-input" placeholder="Color" required="required"
-                       v-model="track.color" title="track-color"/>
+                       v-model="track.color" title="track-color" v-on:input="changefont"/>
```
niranjan94: Use a watcher on `track.color` to trigger the font color change

app/static/js/events/wizard/sessions-speakers/vue-components.js
```diff
@@ -7,6 +7,9 @@ Vue.component('session-track', {
         },
         add: function () {
             this.$emit('add');
+        },
+        changefont: function () {
```
niranjan94: Why are you emitting this ?

app/api/sessions.py
```diff
@@ -31,6 +31,7 @@
     'id': fields.Integer(required=True),
     'name': fields.String(),
     'color': fields.Color(),
+    'font_color': fields.Color(),
```
niranjan94: Don't allow user to set font color. Font color needs to be a readonly attribute on the API

app/models/track.py
```diff
@@ -9,19 +9,21 @@ class Track(db.Model):
     description = db.Column(db.Text)
     track_image_url = db.Column(db.Text)
     color = db.Column(db.String, nullable=False)
+    font_color = db.Column(db.String)
```
niranjan94: Don't store fontcolor to data base. It's a generated property. Generate it based on track color and provide it in the API

app/api/tracks.py
```diff
@@ -24,6 +24,7 @@
     'name': fields.String(required=True),
     'description': fields.String(),
     'color': fields.Color(required=True),
+    'font_color': fields.Color(required=True),
```
niranjan94: Same comment as above.

app/static/js/events/wizard/sessions-speakers/vue-components.js
mejariamol: child component has to emit the event for parent to receive right? Ok I will go for *watch implementation*.

app/models/track.py
mejariamol: @niranjan94 how can we provide font color in the api without saving in database ?

niranjan94: Define a property in the track model (in `app/models/track.py`). Similar to the already existing `serialize` property. Then you can just use it like a db field in the API Model definition.

mejariamol: @niranjan94 ok.. so that means there will be a function in `app/models/track.py`  generating fontcolor from color right?

niranjan94:
Yes. Correct.

mejariamol:
@niranjan94 have a look at last changes.. 

1. font_color is defined as property in `app/models/track.py`. 
2. set watcher to **color** to change font_color
3. but couldn't understand how to get font_color field in the get request only in api .

niranjan94:
@mejariamol looks good. You should already get font color in the API now. 
All you have to do next is to make sure the API doesn't try to save font_color when someone makes a post/put/patch request to tracks endpoint

shubham-padia:
@mejariamol Try adding `del TRACK_POST['font_color']` on https://github.com/mejariamol/open-event-orga-server/blob/a9ba40a0275c5b2060eb481bef13237a8b22d0a2/app/api/tracks.py#L40, AFAIK that should make the field read-only.

mejariamol:
@niranjan94 @shubham-padia Travis CI is failing due to testing of color field against `red` in the function where I am parsing r g b from hex code. `int(h[0:2], 16)`.

niranjan94:
@mejariamol you need to add font color here https://github.com/fossasia/open-event-orga-server/blob/development/tests/unittests/api/utils_post_data.py#L118

mejariamol:
@niranjan94 yes. but then also it will fail while running this.. right?
```
@property
    def font_color(self):
        h = self.color.lstrip('#')
        a = 1 - (0.299 * int(h[0:2], 16) + 0.587 * int(h[2:4], 16) + 0.114 * int(h[4:6], 16))/255
        ...
```

mejariamol:
@niranjan94 travis build passed for last commit. can be merged now.

PR #3569, opened 2017-05-06
Option to provide links instead of uploading media on creating session

### LINKED ISSUE #3492: Link for Video/Slides/Audio missing - Implement option a) Upload or b) Link
Currently the link for video, slides, audio cannot be provided as only an upload option is available in the sessions form. 

Please provide an option either link or upload.

screenshot from 2017-04-14 14-05-25

poush:
fixes #3492 

screen shot 2017-05-06 at 10 52 30 am

### DIFF
--- a/app/helpers/data.py
+++ b/app/helpers/data.py
@@ -25,7 +25,8 @@ from app.helpers.notification_email_triggers import trigger_new_session_notifica
     trigger_session_state_change_notifications
 from app.helpers.oauth import OAuth, FbOAuth, InstagramOAuth, TwitterOAuth
 from app.helpers.sessions_speakers.speakers import save_speaker
-from app.helpers.storage import upload, UPLOAD_PATHS, UploadedFile, upload_local
+from app.helpers.storage import upload, UPLOAD_PATHS, UploadedFile, upload_local, \
+    is_external_file
 from app.helpers import helpers as Helper
 from app.helpers.data_getter import DataGetter
 from app.helpers.system_mails import MAILS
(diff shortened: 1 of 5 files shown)


app/helpers/data.py
```diff
@@ -204,20 +204,36 @@ def add_session_to_event(request, event_id, state=None, use_current_user=True, n
...
         form = request.form
         slide_temp_url = form.get('slides_url')
+        slide_custom_url= form.get('slides_custom_url')
+
         video_temp_url = form.get('video_url')
+        video_custom_url= form.get('video_custom_url')
+
         audio_temp_url = form.get('audio_url')
+        audio_custom_url= form.get('audio_custom_url')
+
```
shubham-padia: I think these extra fields are not needed, this can be implemented using `slides_url`, etc. itself.

poush:
> Currently while editing the proposal, the blue file bar will appear even if the user has linked to an external source which need not necessarily be a file, e.g a google drive link which is misleading.
> Please prefill the url box with the url when editing the proposal.

I already took care of both these in my PR but still if you are getting that box, I will make a recheck and retest of my code.

app/helpers/data.py
poush: The main issue I do not want show internal url of upload files. If I do not create new field, I will use field with "audio_url" as text field ( not hidden ) but then it will show uploaded paths on uploading any media file.

shubham-padia: IMO there's no problem in showing those urls. You will have to show them that anyways when the user comes back to edit the cfs (i.e when prefilling the url box in edit mode, which is not happening as of now), as there is no distinction b/w a uploaded file url and linked url right now.

niranjan94:
@shubham-padia @SaptakS @magdalenesuo please help review this.

poush:
Thanks everyone for the reviews. I have updated this PR. Can you please rereview it?

magdalenesuo:
@poush Hi Poush,
Two things not solved yet as per @shubham-padia review:
> Currently while editing the proposal, the blue file bar will appear even if the user has linked to an external source which need not necessarily be a file, e.g a google drive link which is misleading.

> One way to solve this is to get rid of the bar when editing the proposal. Instead place a small button beside the url input box, which opens the url on clicking it.

zz

I still get the blue bar on edit that seem to have no download connection. Maybe replace that with a button like said above to open the external link?

And you still seems to be having codacy issues.

poush:
@magdalenesuo I understand it was showing before but it should not now. I tried testing with google docs and various other links as well.

screen shot 2017-05-19 at 9 52 46 am

Can you please make sure you have removed all previous commits on the branch you are testing it ?

Also, the codacy issue seems to be related with template filter which is not directly used. @niranjan94 is there any way to tell codacy to ignore it?

SaptakS:
> Also, the codacy issue seems to be related with template filter which is not directly used. @niranjan94 is there any way to tell codacy to ignore it?

Yeah, seems so. Thanks for pointing out. Will modify the filter.

PR #3571, opened 2017-05-07
use specified storage for exporting zip of event

### LINKED ISSUE #3523: Make file exports take into consideration the storage setting on the app
Right now, file exports write the exported zip to the local file system by default irrespective of what the user's storage setting (google cloud, s3, local) is. Ensure the exported zip is placed in the appropriate location.

This is especially important when we have something like a kubernetes deployment where the local-storage is volatile and isolated by default. Local storage is not sharable b/w pods by default. (In our case, celery's local storage is isolated from the web app's local storage, so, creating a shared persistent storage space that is read-writable by multiple pods is both tricky and slightly unreliable during re-deployments)

poush:
fixes #3523 

1. Wait for ```shutil.make_archive``` to complete archive and store it in local storage.
2. Copy the created archive to storage ( specified by user )
3. Delete local archive created.

### DIFF
--- a/app/api/helpers/export_helpers.py
+++ b/app/api/helpers/export_helpers.py
@@ -23,6 +23,8 @@ from ..sessions import DAO as SessionDAO, SESSION, \
 from ..speakers import DAO as SpeakerDAO, SPEAKER
 from ..sponsors import DAO as SponsorDAO, SPONSOR
 from ..tracks import DAO as TrackDAO, TRACK
+from app.helpers.storage import upload, UPLOAD_PATHS, UploadedFile
+from app.settings import get_settings
 
 # DELETE FIELDS
 # All fields to be deleted go here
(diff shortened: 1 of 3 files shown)


app/api/helpers/export_helpers.py
```diff
@@ -207,7 +208,15 @@ def export_event_json(event_id, settings):
...
     # make zip
     shutil.make_archive(dir_path, 'zip', dir_path)
-    return dir_path + '.zip'
+    dir_path = dir_path + ".zip"
+
+    storage_path = UPLOAD_PATHS['exports']['zip'].format(
+        event_id = event_id
+    )
+    uploaded_file = UploadedFile(dir_path, dir_path.rsplit('/', 1)[1])
+    storage_url = upload( uploaded_file, storage_path )
```
aviaryan: PR looks good functionality-wise. Neat work !!
One thing I see here is that you are giving extra trailing spaces in function parameters. eg `upload( uploaded_file, storage_path )`. Please remove them. 
We would like to follow a consistent style everywhere in the code.

poush:
@niranjan94 will squash just after travis passes for last commit.

SaptakS:
@niranjan94 I think the commits have been squashed. Please review, so that it can be merged. Thanks.

app/api/helpers/tasks.py
```diff
@@ -46,10 +47,15 @@ def export_event_task(self, event_id, settings):
         logging.info('Exporting started')
         path = event_export_task_base(event_id, settings)
         # task_id = self.request.id.__str__()  # str(async result)
-        result = {
-            'download_url': url_for(
+        if get_settings()['storage_place'] == 'local' or get_settings()['storage_place'] == None:
+            download_url =  url_for(
```
shubham-padia: multiple spaces after `=`, use only single.

tests/unittests/api/test_export_import.py
```diff
@@ -18,7 +18,6 @@
 from tests.unittests.setup_database import Setup
 from tests.unittests.utils import OpenEventTestCase
 
-
```
shubham-padia: PEP8 expects two blank lines before a class, please dont remove this

poush:
Thanks, @shubham-padia!

PR #3573, opened 2017-05-08
CFS: Make addition of speaker details necessary before submitting ses…

SaptakS:
As discussed in https://github.com/fossasia/open-event-orga-server/issues/3560#issuecomment-298879182 implemented the workflow for the sessions. However previously submitted session which weren't implemented using this workflow still exists. 

So though doesn't exactly fix the issue, but I don't think we can do anything about already submitted session without registering speaker. We can pretty well delete or disable them from DB. @mariobehling @niranjan94 what do you think??


mariobehling:
@SaptakS Seems good for me. Cheers.

PR #3574, opened 2017-05-08
making background_image field in Event API nested

### LINKED ISSUE #3518: Make background_image field in Event API nested
@niranjan94 said (https://github.com/fossasia/open-event-orga-server/pull/3514#issuecomment-295222321): 

While we are at it .. Can we put these fields `thumbnail`, `large` and `background_image` into a nested object ? 

Something like,
```json
{
    ... other event object keys ...

    "background_image": {
             "thumbnail_image_url": "url for that",
             "large_image_url": "url for that",
             "original_image_url": "url for that"
    }
}
```
If there is no background image available `background_image` will be `null`

poush:
fixes #3518

### DIFF
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -74,7 +74,11 @@ EVENT = api.model('Event', {
     'timezone': fields.String(),
     'latitude': fields.Float(),
     'longitude': fields.Float(),
-    'background_image': fields.Upload(attribute='background_url'),
+    'background_image': {
+        'original_image_url': fields.Upload(attribute='background_url'),
+        'large_image_url': fields.Uri(attribute='large'),
+        'thumbnail_image_url': fields.Uri(attribute='thumbnail'),
+    },
     'placeholder_url': fields.PlaceHolder(attribute=lambda event: event),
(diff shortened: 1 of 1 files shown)


niranjan94:
@poush also open an issue at https://github.com/fossasia/open-event-webapp and https://github.com/fossasia/open-event-android once this is merged so that they can make changes to match the new API response

app/api/events.py
```diff
@@ -74,7 +74,11 @@
     'timezone': fields.String(),
     'latitude': fields.Float(),
     'longitude': fields.Float(),
-    'background_image': fields.Upload(attribute='background_url'),
+    'background_image': {
```
niranjan94: @poush I don't think you define nested attributes like this. I believe you need to use `fields.Nested`. Not very sure though. Could you check Flask-RESTPlus docs and confirm ?

Also, please check and fix the travis build fail.

poush: Sure

SaptakS: @poush any update on this issue?

PR #3579, opened 2017-05-11
Remove datepair.js from access_codes and discount_codes

### LINKED ISSUE #3578: Validity period for promotional codes cannot be longer than one day
When trying to enter a vailidity period for a promotional code and one selects a date on the date picker for either "VALID FROM" or "EXPIRES ON" both dates are changed. 

When I first entered a date for EXPIRES ON it was set for both. When I tried to change the VALID FROM both dates were changed to the same date.   
Now both dates are adjusted when I change one. It seems to be an offset I cannot influence.

I would like to be able to set both dates independently.

shubham-padia:
fixes #3578 (Problem with setting dates in access and discount codes)


shubham-padia:
@niranjan94 please review

PR #3581, opened 2017-05-12
Events Wizard Fix: Switching buttons on and off results in data loss

### LINKED ISSUE #3495: Switching buttons on and off results in data loss
When an organizer switches key buttons on or off all data is gone. Expected: When the organizers switches the button on again, the data should be there again, e.g.
* Ticket module on wizard step 1
* Sponsors button wizard step 2
* Sessions and Speakers wizard step 3

Secondly, switching buttons on and off in step 3 results in "all sessions" that have been submitted being deleted. Expected: Sessions should still be there, but they cannot be in the specific tracks anymore. Actually the speaker submissions are still there, but the sessions are gone. Please ensure the sessions are staying in the system in any case.

screenshot from 2017-04-14 18-05-06

shubham-padia:
fixes #3495 .
@niranjan94 @poush please review.


shubham-padia:
@SaptakS Both of those switch are working as expected atm. Resolved other codacy issues and left comments for the remaing.

app/static/js/events/wizard/sessions-speakers/vue-models.js
```diff
@@ -1,3 +1,16 @@
+Date.prototype.ddmmyyyy = function() {
```
shubham-padia: This is not overriding any preexisting methods of the said native object. So IMO it's okay to let this go.

PR #3583, opened 2017-05-13
Modified DATABASE_URL to avoid authentication error

### LINKED ISSUE #3582: Wrong Password in DATABASE_URL in basic.md lead to authentication error
Currently the DATABASE_URL in the basic.md file provided for generic installation is having a wrong password which must be set to "test" as provided while creating the user in the postgres sql instead of the "start" provided which lead to authentication error while executing manage.py file.

RamolaWeb:
Fix #3582 to avoid authentication error causing due to wrong user password

### DIFF
--- a/docs/installation/basic.md
+++ b/docs/installation/basic.md
@@ -70,7 +70,7 @@ bower install
 * **Step 4** - Create application environment variables.
 
 ```sh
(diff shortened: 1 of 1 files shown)


shubham-padia:
@niranjan94 Can you merge this please :-) .

PR #3584, opened 2017-05-13
Super admin Fix: cannot update system images

### LINKED ISSUE #3553: Updating system images through UI resulting in error and eternal upload screen
Trying to update system images is not successful at https://eventyay.com/admin/content/#

screenshot from 2017-04-28 10-56-33

shubham-padia:
fixes #3553 .
The image updation works for all deployments(local or heroku) using local storage. The error occurs when anything except local storage is used, probably s3 (e.g does not work on http://open-event-dev.herokuapp.com). So I made some conclusions of what might be causing the error and provided the fix. The fix may or may not work for external storage, but this also does not break the updation for local storage.

### DIFF
--- a/app/views/super_admin/content.py
+++ b/app/views/super_admin/content.py
@@ -15,7 +15,7 @@ from flask import flash
 from app.helpers.data import DataManager, delete_from_db, save_to_db
 from app.helpers.data_getter import DataGetter
 from app.helpers.helpers import uploaded_file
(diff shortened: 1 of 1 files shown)


shubham-padia:
:white_check_mark: Resolved. Checked on http://open-event-dev.herokuapp.com/admin/content/ and this works as of now.

PR #3596, opened 2017-05-15
Timeline block on scheduler page is spread to full screen

### LINKED ISSUE #3589: Full Screen Schedule not anymore showing on full screen
The full screen schedule is cut-off and shows a black area on the bottom.

Compare: https://eventyay.com/e/5642d9a1/schedule/

screenshot from 2017-05-15 00-57-49

mejariamol:
fixes #3589 

screenshot from 2017-05-16 04-06-46

### DIFF
--- a/app/templates/gentelella/guest/event/schedule.html
+++ b/app/templates/gentelella/guest/event/schedule.html
@@ -151,7 +151,7 @@
                     $('.overlay').addClass('visible');
                     $('.btn-pop-out').removeClass('fa-expand').addClass('fa-compress');
                     var width = $("#timeline").width() - $('.timeunits.x1').width();
-                    $('.microlocation-container').css("width", width + "px");
+                    $('.microlocation-container').css("max-width", width + "px");
                     $('.session.scheduled').popover('hide');
                 }
                 function close(){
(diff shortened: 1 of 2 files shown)


mejariamol:
@SaptakS both the pages use same template `gentelella/users/events/scheduler/components/timeline.html` for showing timeline block. 

Actually I couldn't test for scheduler in public page since it's throwing 404 for `e/<identifier>/schedule/` on localhost. So I just tested for dashboard. I think it will work correctly on public page. please check.

mejariamol:
@SaptakS yes. public page will need some extra code. re-committing.

app/templates/gentelella/guest/event/schedule.html
```diff
@@ -151,15 +151,16 @@ <h1 style="font-weight: 300; font-size: 24px">
...
-                    $('.microlocation-container').css("width", width + "px");
+                    $('.microlocation-container').css("max-width", width + "px");
                     $('.session.scheduled').popover('hide');
                 }
                 function close(){
                     $scheduler.removeClass('scheduler-pop');
                     $body.removeClass('noscroll');
                     $('.overlay').removeClass('visible');
                     $('.btn-pop-out').addClass('fa-expand').removeClass('fa-compress');
-                    $('.microlocation-container').css("width", "fit-content");
```
SaptakS: As far as I remember, the fit-content style was adding to avoid the extra space on the right side in case the number of microlocations were less. Can you test that it still works fine on making your change?

SaptakS: According to your screenshot, it seems the extra space on the right side of the timeline is back. So please test that.

mejariamol: @SaptakS Yes. I have tested that. Its not leaving extra space on the right side because now microlocation-container has only max-width property hence will wrap. the previous screenshot was before commit. 

screenshot from 2017-05-17 11-27-14

PR #3598, opened 2017-05-16
Address changed to locality location full name

SaptakS:
Fixes https://github.com/fossasia/open-event-orga-server/issues/3221#issuecomment-301774697

I have kept the previous formats only so for some latitude, longitude the locality key would be missing and not shown in the footer, which is okay I guess because anyways we are showing only the most common ones.

As of the 'P', it was the short name for the location. So I have replaced with the long_name that was being received.

### DIFF
--- a/app/helpers/data_getter.py
+++ b/app/helpers/data_getter.py
@@ -709,13 +709,20 @@ class DataGetter(object):
         return MessageSettings.query.filter_by(action=action).first()
 
     @staticmethod
+    @cache.cached(timeout=21600, key_prefix='event_locations')
     def get_locations_of_events():
         names = []
         try:
             for event in DataGetter.get_live_and_public_events():
                 if not string_empty(event.location_name) and not string_empty(event.latitude) and not string_empty(
                      event.longitude):
-                    names.append(event.location_name.split(',')[0])
+                    response = requests.get(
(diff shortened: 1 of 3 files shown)


app/helpers/data_getter.py
```diff
@@ -715,7 +715,13 @@ def get_locations_of_events():
             for event in DataGetter.get_live_and_public_events():
                 if not string_empty(event.location_name) and not string_empty(event.latitude) and not string_empty(
                      event.longitude):
-                    names.append(event.location_name.split(',')[0])
```
niranjan94: Instead of a request, try using searchable_location_name. 

Request to the API is gonna be very slow without caching

SaptakS: > Instead of a request, try using searchable_location_name.

@niranjan94 I just checked. the problem with searchable_location_name is it was storing short_name instead of the long_name. And since it is already stored in the db, it is showing the wrong output. So I can improve it for the later events, but already stored locations would still be wrong.

> Request to the API is gonna be very slow without caching

I know it's gonna be slow. So I will try implementing the cache once more and check.

niranjan94: @SaptakS okay .. sure

SaptakS: Resolved.

PR #3603, opened 2017-05-17
Ticket order pdf now prints all names of holders

### LINKED ISSUE #3601: Buy 4 - get one - Ticket issue
If you buy more than Tickets at one Computer the exported pdf will show only the name of the second last registred person.

this brings a lot off confussion

magdalenesuo:
Fixes #3601
screenshot from 2017-05-17 13-21-38
screenshot from 2017-05-17 13-21-33

### DIFF
--- a/app/templates/gentelella/guest/ticketing/pdf/ticket.html
+++ b/app/templates/gentelella/guest/ticketing/pdf/ticket.html
@@ -28,7 +28,6 @@
 
              @frame col2_frame {
                 left: 440pt; top: 170pt;
-                -pdf-frame-content: personal_content;
             }
         }
 
@@ -65,9 +64,9 @@
         <div id="qrcode">
             <img src="data:image/png;base64,{{ holder.qr_code }}" />
         </div>
-        <strong id="number_content" style="color: #ffffff; text-align: center">#{{ order.get_invoice_number() }}/{{ holder.id }}</strong>
+        <strong id="number_content" style="color: #000; text-align: center">#{{ order.get_invoice_number() }}/{{ holder.id }}</strong>
 
-        <div id="personal_content">
+        <div id="personal_content_{{ holder.id }}">
             <span>{{ holder.name }}</span><br>
             <span>{{ holder.email }}</span>
         </div>


shubham-padia:
`-pdf-frame-content` has to be removed from everywhere I think. According to https://github.com/xhtml2pdf/xhtml2pdf/blob/master/doc/source/format_html.rst#static-frames-vs-content-frames , `-pdf-frame-content` represents static frame , i.e it will be same across all pages. Currently in the screenshot you can see qr codes and other fields are also same

SaptakS:
@shubham-padia it was used exactly to make things static content. I agree it wasn't needed in this case, but it was needed in other case. As you will see in dynamic content, there is an overflow of things due to which the ticket name in case it was big was going out of space instead of wrapping causing not be printed in the pdf. That is when I changed them to static content. It also helps in making the layout much more proper and easier to control. See [my PR](https://github.com/fossasia/open-event-orga-server/pull/3513)

So that is all right.

SaptakS:
And other fields will be same anyways.

shubham-padia:
@SaptakS Okay. But i'm assuming qr_codes for each holder are different, so does it have to removed from qr code or not ?

magdalenesuo:
@shubham-padia Yes we can. It is however not hampering anything anyway so we can keep it. Removing it from everywhere will give something like this:
screenshot from 2017-05-17 13-48-10
which, I think will be more work trying to fix alignments.
So I think it is better to keep everything static and this dynamic.

shubham-padia:
@magadalenesuo agreed :smile: . But from my last comment above, I was asking whether it has to be removed from qr_codes also ?

SaptakS:
@magdalenesuo agreed. Also, if you check for a very long ticket name, then the ticket name doesn't show.

Even removing it from qr_code breaks the layout. But I checked and found qr_code to be same for all holders. Means it is also an order property.

magdalenesuo:
@SaptakS : I agree with @shubham-padia, removing from the qr_codes was a better option.

SaptakS:
@magdalenesuo @shubham-padia @niranjan94 sorry for merging the PR before the end of the discussion. Anyone of you can send a PR marking these comments.

niranjan94:
@SaptakS it's alright. No issue. 

And something else ... Yes @shubham-padia . The QR should be different for each ticket holder.

@magdalenesuo can you confirm that the QR-Code generated & displayed is different for each ticket ?

magdalenesuo:
@niranjan94 I get different QR numbers, but images seems to be the same: On scanning two tickets, I get both the same.
screenshot_2017-05-17-14-33-24-765_com eg ebit qr code and barcode qr scanner
screenshot_2017-05-17-14-33-39-113_com eg ebit qr code and barcode qr scanner

SaptakS:
@magdalenesuo is this before or after you remove static content from qr_code?

magdalenesuo:
@SaptakS Sorry I was checking without removing static content.
It is different on removing static.

SaptakS:
@magdalenesuo please make a PR removing the static content then.

PR #3605, opened 2017-05-17
Different QR Code generated

### LINKED ISSUE #3604: QR code developed same on static pdf
From discussion [#3603(comment)](https://github.com/fossasia/open-event-orga-server/pull/3603#issuecomment-302020755)
and [#3603(comment)](https://github.com/fossasia/open-event-orga-server/pull/3603#issuecomment-302028980)

magdalenesuo:
fixes #3604
screenshot from 2017-05-17 15-37-04
screenshot from 2017-05-17 15-37-14

### DIFF
--- a/app/templates/gentelella/guest/ticketing/pdf/ticket.html
+++ b/app/templates/gentelella/guest/ticketing/pdf/ticket.html
@@ -11,13 +11,11 @@
             @frame col1_frame {             /* Content frame 1 */
                 left: 80pt;  top: 30pt;
                 height: 250pt; width: 300pt;
-                -pdf-frame-content: main_content;
             }
 
             @frame qrcode {
                 left: 455pt;  top: 50pt;
                 width: 100pt; height: 120pt;
-                -pdf-frame-content: qrcode;
             }
             @frame number_frame {
                  left: 455pt;  top: 25pt;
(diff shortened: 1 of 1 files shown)


SaptakS:
@shubham-padia won't moving things out of the loop prevent them from appearing in all the tickets? Don't you think so?

shubham-padia:
@SaptakS Nope. i tried it just now. If it's static content, it has to be specified only once. Of course, everything except static should still remain in the loop.

magdalenesuo:
@shubham-padia or instead try to remove static from coll_frame too? :'D

shubham-padia:
@magdalenesuo That's also a good option :+1:

SaptakS:
Seems I didn't need to make it any of it static content. All I needed was to assign a particular width and arrange the divs properly. lol

app/templates/gentelella/guest/ticketing/pdf/ticket.html
```diff
@@ -28,7 +26,6 @@
```
shubham-padia: @magdalenesuo Should this be also removed, since `number_content` also holds holder info. I'm not sure tho. Can you please check once.

magdalenesuo: This doesn't seem to be changing anything so keeping it static is all good.
We can remove it too, will have to adjust the alignments.

shubham-padia: If it doesn't change anything, let it be as it is then for now.

PR #3612, opened 2017-05-18
Implement signals to fix ical,xcal,pentabarf export timeout

### LINKED ISSUE #3299: pentabarf, ical, xcal implementation not working
The ical, pentabarf, xcal implementations are linked in the "Export" tab. Currently using the links, e.g. https://eventyay.com/e/45da88b7/schedule/calendar.ics shows an error page.

shubham-padia:
#### Checklist

When exporting the ical/xcal/pentabarf files, the connection was timing out before the file was generated because of the event being too large.

- Add a celery task to generate a file for each export implementation everytime relevant changes in the event occur.
- Upload the file to the preffered storage and store its url in db.
- Redirect to the file url, when the user tries to access the export view.

Fixes #3299.

#### Other information:

Currently 4 signals have been implemented:
- `event_json_modified` (core changes related to the event, excludes changes in speakers, microlocations, etc)
- `sessions_modified`
- `speakers_modified`
- `microlocations_modified`

#### Conventions to follow

- Declare all your signals in `app/helpers/signals.py`, but do not put any signal handler in the said file. Each signal handler should be the file related to which it is performing its functionality.
- When sending signals, signals accept only one positional argument (the sender), rest have to be keyword arguments. 
- Thus, for the sender, From http://flask.pocoo.org/docs/0.12/signals/#sending-signals,
  `Try to always pick a good sender. If you have a class that is emitting a signal, pass self as sender. If you are emitting a signal from a random function, you can pass current_app._get_current_object() as sender.`
- We will be using `current_app._get_current_object()` as sender throughout the project.

- #### Signal handlers:
  - When sending the signal, the signal may be sending lots of information, which your signal may or may not want.
  - Thus format your signal handler in the form of: ` def name(app, **kwargs): `
  - The above format will allow future devs to attach other extra information to the signal without affecting you existing handler.Also with this format, you will not be accepting (explicitly, of course) any parameters you don't need.

#### For reviewers:
@niranjan94 @SaptakS @magdalenesuo Please review.
I have also added some conventions for easy workflow, please suggest any changes to them. 
Also if the conventions seem okay, @niranjan94 where should they be put ?


shubham-padia:
Travis failing looking into it.

app/helpers/tasks.py
```diff
@@ -35,3 +43,57 @@ def send_mail_via_smtp_task(config, payload):
...
+
+@celery.task(name='export.pentabarf')
+def export_pentabarf_task(event_id):
+    event = DataGetter.get_event(event_id)
+    try:
+        os.mkdir(app.config['TEMP_UPLOADS_FOLDER'])
+    except OSError as exc:
+        if exc.errno != errno.EEXIST:
+            raise exc
+        pass
```
SaptakS: @shubham-padia why the `pass` statement here?

app/helpers/tasks.py
```diff
@@ -35,3 +43,57 @@ def send_mail_via_smtp_task(config, payload):
...
+
+@celery.task(name='export.ical')
+def export_ical_task(event_id):
+    event = DataGetter.get_event(event_id)
+    try:
+        os.mkdir(app.config['TEMP_UPLOADS_FOLDER'])
+    except OSError as exc:
+        if exc.errno != errno.EEXIST:
+            raise exc
+        pass
```
SaptakS: Same for here.

SaptakS:
This is not a request change but just wanted to know one thing. Since you are calling 3 export tasks even if there is a small change in session or speaker or event, isn't it adding too much over head on celery and queueing up too much tasks. Maybe it might not even be needed because maybe no one exports that event. @magdalenesuo @niranjan94 what do you think? @shubham-padia am I missing something?

niranjan94:
@SaptakS what would be the alternative here ? Because doing what we do now (ie) generating every time it is requested isn't working out.

shubham-padia:
@niranjan94 Coming to think of it now, we can implement the same approach as we are following on the current. You click on export zip, a celery task runs and returns url on successful completion. What are your thoughts?

niranjan94:
@shubham-padia no. That doesn't work for most cases when it comes to PentabarfXML and ical.
We would like give our users the ability to specify the URL directly in third-party apps.

niranjan94:
@shubham-padia @SaptakS for example, In Giggity, the users should just be able to specify something like `https://eventyay.com/e/abcdefg/schedule.xml` 

screenshot_20170518-155723

SaptakS:
@niranjan94 what I was suggesting and I think @shubham-padia was also suggesting is creating the xml after clicking the export button and then showing the url (updated). Won't that work? I mean just start the task of creation on clicking pentabarf export button rather than on making every single change. Because it will heavily increase the number of queued tasks. Since session and event editing occurs frequently.

shubham-padia:
@niranjan94 Got it. Looking into the travis errors right now.

niranjan94:
No @SaptakS . For these sort of things, there shouldn't be any admin interaction required. These links will be publicly available in the event page for anyone to use.

niranjan94:
To reduce the overhead, we could maybe keep it such that, this export task will run a maximum of only once per day irrespective of the number of changes that occurs during that day.

SaptakS:
> No @SaptakS . For these sort of things, there shouldn't be any admin interaction required. These links will be publicly available in the event page for anyone to use.

But right now it is a feature only for the admins and organizers, right?

niranjan94:
@SaptakS not at all. These schedules are and must be publicly available.

SaptakS:
@niranjan94 Then only way of reducing overhead seems to be as suggested by you. But that might actually give a wrong schedule if things are updated after pentabarf was created. 
I am still not sure why permission will be required. Can't a public view send a signal and if such a signal is received, we can create the export. Shouldn't that be possible?? Because we are not directly sending request to a view which we don't have access to. Rather sending out signals.

niranjan94:
@SaptakS what permission are you talking about ?

SaptakS:
@niranjan94 @shubham-padia I mean can't something like this be done. The public click a button to show the url, a signal is sent, the export is done, url updated and then shown to the public? I haven't worked with signals so I am not sure.

niranjan94:
@SaptakS what to do when the sessions are updated ? Do we expect the public to check if schedule is updated and generate each time ?

niranjan94:
All the conference websites/pages I see just provide simple links for ical, xml, xcal exports.

SaptakS:
@niranjan94 yeah, that is a problem too. Then I guess your solution is the only solution. We can give a time interval. If not updated in that time interval, then update it.

app/helpers/wizard/sessions_speakers.py
```diff
@@ -111,9 +113,3 @@ def save_custom_forms(data, event_id):
     custom_forms.speaker_form = json.dumps(data['speaker'])
     save_to_db(custom_forms)
 
-
-def delete_all_sessions_speakers_data(event_id):
-    Microlocation.query.filter_by(event_id=event_id).delete()
-    SessionType.query.filter_by(event_id=event_id).delete()
-    Track.query.filter_by(event_id=event_id).delete()
-    CallForPaper.query.filter_by(event_id=event_id).delete()
```
shubham-padia: This function was sitting randomly, without being used anywhere, that's why its been deleted.

shubham-padia:
Closing this for some testing purposes of failing travis, don't want to send a constant barrage of codacy emails while I'm doing that .

shubham-padia:
#### What is currently happening:
- Multiple ` DetachedInstanceError: Instance <User at 0x7f358a4e9550> is not bound to a Session; attribute refresh operation cannot proceed ` errors are being whenever the celery task are being called by the appropriate signal handlers.( log [here](https://travis-ci.org/fossasia/open-event-orga-server/builds/233588618#L1161)

#### What I've noticed till now:
- Instead of running the tasks in celery queue i.e removing `@celery.task.....` removes all the above errors, indicating it's not a problem with signals i guess.
- `CELERY_ALWAYS_EAGER` is set to true, so by definition it should run as a normal task, don't know what's wrong.

I was wondering what's our current implementation on testing celery tasks.
@niranjan94 @SaptakS  Any insight on this will be appreciated.

shubham-padia:
Continued on #3612

PR #3629, opened 2017-05-21
Implementing connected social applications like twitter

### LINKED ISSUE #2803: Implement Apps similar to twitter applications page

### LINKED ISSUE #3563: Cannot sign in with Facebook Oath Error
Trying to sign up with Facebook not possible.

Getting "OAuth Authorization error. Please try again later."

screenshot from 2017-05-03 13-06-24

poush:
#### Checklist

This PR resolves an issue where the connect buttons for social medias were not working. Now through these buttons, a user can connect his social media accounts with orga server.
This PR also resolves an issue where OAuth for Facebook was not working. This was due to server asking for a field ('bio') to the Graph API while it was already deprecated and removed from the current API

- Ensured that there are views to connect all four social media accounts
- Made some changes in 'gCallback' view to ensure it works with 
- These connect button will be shown only if the required app details are provided by admin.
- Removed 'bio' field from the details requested by orga server to facebook API
- Updated update_user_details to add support for all 4 fields
- Create view '/unlink-social/<social>' to remove/unlink the account.
- Currently, the app stores just the handles/link to the profile of the user and any support to save the token can be done in a separate issue.

Fixes #2803 
Fixes #3563

### DIFF
--- a/app/templates/gentelella/users/settings/pages/applications.html
+++ b/app/templates/gentelella/users/settings/pages/applications.html
@@ -14,20 +14,27 @@
 {% block content %}
 
     <div style="display: block;">
+        {% if fb %}
         <div class="row" style="margin-bottom:20px">
-
             <div class="col-md-6 col-xs-6">
                 <label>Facebook:</label>
                 <div class="input-group">
                     <span class="input-group-addon social-addon">facebook.com/</span>
                     <input type="text" name="facebook" class="form-control"
-                           value="{{ user.user_detail.facebook|default('', true) | replace('https://facebook.com/', '') }}"
+                           value="{{ user.user_detail.facebook|default('', true) | replace('https://www.facebook.com/', '') }}"
                            style="min-width: 400px"
-                           disabled />
-                    {% if not user_id %}
+                           disabled/>
+                    {% if user.user_detail.facebook %}
                         <div class="input-group-btn">
-                            <a type="button" class="btn btn-info"
-                               style="margin-bottom: 0">{{ _("Connected") }}</a>
+                            <a  type="button" class="btn btn-info"
+                                href="/unlink-social/facebook?next=settings/applications/"
+                                style="margin-bottom: 0" title="{{ _('Click to Disconnect') }}" data-toggle="tooltip">{{ _("Connected") }}</a>
+                        </div>
+                    {% else %}
+                        <div class="input-group-btn">
+                            <a  type="button" class="btn btn-info"
+                                href="/profile/fb_connect"
+                                style="margin-bottom: 0" >{{ _("Connect") }}</a>
                         </div>
                     {% endif %}
 
@@ -36,7 +43,9 @@
             </div>
 
         </div>
+        {% endif %}
 
+        {% if twitter %}
         <div class="row" style="margin-bottom:20px">
             <div class="col-md-6 col-xs-6">
                 <label>Twitter:</label>
@@ -46,16 +55,26 @@
                     <input type="text" name="twitter" class="form-control"
                            value="{{ user.user_detail.twitter|default('', true) | replace('https://twitter.com/', '') }}"
                            style="min-width: 400px"
-                           disabled />
-                    {% if not user_id %}
+                           disabled/>
+                    {% if user.user_detail.twitter %}
+                        <div class="input-group-btn">
+                            <a  type="button" class="btn btn-info"
+                                href="/unlink-social/twitter?next=settings/applications/"
+                                style="margin-bottom: 0" title="{{ _('Click to Disconnect') }}" data-toggle="tooltip">{{ _("Connected") }}</a>
+                        </div>
+                    {% else %}
                         <div class="input-group-btn">
                             <a type="button" class="btn btn-info"
-                               style="margin-bottom: 0">{{ _("Connected") }}</a>
+                                href="/profile/tw_connect"
+                               style="margin-bottom: 0" >{{ _("Connect") }}</a>
                         </div>
                     {% endif %}
                 </div>
             </div>
         </div>
+        {% endif %}
+
+        {% if insta %}
         <div class="row" style="margin-bottom:20px">
             <div class="col-md-6 col-xs-6">
                 <label>Instagram:</label>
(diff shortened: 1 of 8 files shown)


tests/unittests/functionality/test_google_oauth.py
```diff
@@ -21,9 +21,9 @@ def test_user_already_logged_in(self):
             register(self.app, 'email@gmail.com', 'test')
             logout(self.app)
             login(self.app, 'email@gmail.com', 'test')
-            self.assertTrue('Open Event' in self.app.get('/gCallback/?state=dummy_state&code=dummy_code',
+            self.assertTrue('Open Event' in self.app.get('/gCallback/?state=dummy_state&code=dummy_code&test=1',
```
poush: @niranjan94 @SaptakS I'm not sure if this is the right way to pass tests. Can you please provide any suggestion here.

The issue is that I cannot redirect the authenticated user directly to admin index without saving the user details ( as the user presses the connect button ) but the test expect a redirect here.

app/templates/gentelella/users/settings/pages/applications.html
```diff
@@ -14,20 +14,27 @@
...
+        {% if fb %}
         <div class="row" style="margin-bottom:20px">
-
             <div class="col-md-6 col-xs-6">
                 <label>Facebook:</label>
                 <div class="input-group">
                     <span class="input-group-addon social-addon">facebook.com/</span>
-                    <input type="text" name="facebook" class="form-control"
-                           value="{{ user.user_detail.facebook|default('', true) | replace('https://facebook.com/', '') }}"
+                    <input type="text" name="twitter" class="form-control"
```
SaptakS: why name="twitter" instead of "facebook"?

tests/unittests/functionality/test_google_oauth.py
SaptakS: So what does adding test=1 do? You are checking in the function if it is for test, then redirect directly? If that is what you are doing, doesn't that make the purpose of test wrong itself?

poush: Yes that's where I'm looking for suggestion. Actually earlier the concept was 
```
 if login.current_user is not None and login.current_user.is_authenticated:
        return redirect(url_for('admin.index'))
```
but since we want to update the user details even if the user is already logged in therefore I had to move this to line 59 in util_routes.py

Originally test expect to get redirected to currect url if is already logged in and sign in from google is pressed. So, I think there are two ways to solve this. First is we can modify the test and use it to know it is actually going to save or saving the data but then we are sending the dummy_code which eventually will not give any client records from google. So may be we can just look for exception throw and test it using ```assertRaises```

Another option I got is to keep the redirect condition in 'gCallback' but will be true only if this is the test case. I feel this is not the right way therefore need your suggestion on this.

poush: Any suggestions? Here is sample of test case would be created

````
    def test_user_already_logged_in(self):
        """If the user is already logged in then on clicking 'Login with Google' the user details should be updated"""
        with app.test_request_context():
            register(self.app, 'email@gmail.com', 'test')
            logout(self.app)
            login(self.app, 'email@gmail.com', 'test')
            # an exception will be thrown since there will be no data from user
            self.assertRaises('Open Event' in self.app.get('/gCallback/?state=dummy_state&code=dummy_code',
                                                         follow_redirects=True).data)
````

**Comment by hongquan** [comment]
> we are sending the dummy_code which eventually will not give any client records from google. So may be we can just look for exception throw and test it using assertRaises

We need to fake the Google (or any other external service) response, by using [httmock](https://github.com/patrys/httmock).

The idea is, when `requests` make a HTTP request to _https://api.google.com/profile_, for example, _httmock_ will stand in the middle, stop request from going to Internet, and returns a JSON response as if the response is from Google. The content of this response is written by us in test case. We have to read Google documentation to write correct fake response.

File: app/templates/gentelella/users/settings/pages/applications.html (hunk 2 above)
Comment (poush) [inline app/templates/gentelella/users/settings/pages/applications.html]: Thanks. Fixed the typo.

**Comment by poush** [comment]
Thanks! @hongquan. Got it worked.

File: app/templates/gentelella/users/settings/pages/applications.html [hunk 3]
```diff
@@ -14,20 +14,27 @@
...
+        {% if fb %}
         <div class="row" style="margin-bottom:20px">
-
             <div class="col-md-6 col-xs-6">
                 <label>Facebook:</label>
                 <div class="input-group">
                     <span class="input-group-addon social-addon">facebook.com/</span>
                     <input type="text" name="facebook" class="form-control"
-                           value="{{ user.user_detail.facebook|default('', true) | replace('https://facebook.com/', '') }}"
+                           value="{{ user.user_detail.facebook|default('', true) | replace('https://www.facebook.com/','') }}"
```
hongquan: Why do you remove whitespace?

hongquan:
```python

fb = get_settings()['fb_client_id'] != None and get_settings()['fb_client_secret'] != None
twitter = get_settings()['tw_consumer_key'] != None and get_settings()['tw_consumer_secret'] != None
insta = get_settings()['in_client_id'] != None and get_settings()['in_client_secret'] != None
gplus = get_settings()['google_client_id'] != None and get_settings()['google_client_secret'] != None
```

Have a variable for `get_settings()`. Don't repeat function call many times (not only waste CPU but also ugly).

When you compare with `None`, use `is`, not `==`.

tests/unittests/auth_helper.py
```diff
@@ -1,9 +1,30 @@
...
 
 from app.helpers.data import DataManager, save_to_db
 from app.helpers.helpers import get_serializer
 
 
+@urlmatch(netloc='https://www.googleapis.com/userinfo/v2/me')
+def google_profile_mock(url, request):
+    headers = {'content-type': 'application/json'}
+    content = {
+            "link":"http://google.com/some_id"
```
hongquan: Over-indentation!

app/templates/gentelella/users/settings/pages/applications.html
```diff
@@ -36,26 +43,38 @@
...
+        {% if twitter %}
         <div class="row" style="margin-bottom:20px">
             <div class="col-md-6 col-xs-6">
                 <label>Twitter:</label>
                 <div class="input-group">
                     <span class="input-group-addon social-addon" id="basic-addon3">twitter.com/</span>
 
                     <input type="text" name="twitter" class="form-control"
-                           value="{{ user.user_detail.twitter|default('', true) | replace('https://twitter.com/', '') }}"
+                           value="{{ user.user_detail.twitter|default('', true) | replace('https://twitter.com/','') }}"
```
hongquan: Also whitespace. Why don't fix it?

hongquan:
Please give some thought when you are making change.

Why do you change from

```python
fb = get_settings()['fb_client_id'] != None and get_settings()['fb_client_secret'] != None
```
to
```python
fb = settings['fb_client_id'] is None or settings['fb_client_secret'] is None
```
? The logic is quite reverse!

poush:
@hongquan Yes. Actually, I tried to reverse the logic to get statement looks better. I updated it with not of the complete statement. So baically a 
__NOT of ( Comparison OR Comparison )__

app/templates/gentelella/users/settings/pages/applications.html
poush: Thanks! Fixed.

hongquan:
>  I tried to reverse the logic to get statement looks better

Give other name. You reverse the logic, but still keep old name. That causes misleading.

poush:
Updated!

PR #3643, opened 2017-05-22
Solves Scheduler not found issue 

### LINKED ISSUE #3623: Schedule Not Found
**I'm submitting a ...**  
- [x] bug report

**Current behavior:**
I am getting an error not found when I try to access the schedule of an event :

image

**Expected behavior:**
The schedule should be displayed .

**Steps to reproduce:**
* Create an event 
* Add sessions to it and assign speakers to it 
* In the **Scheduler tab**  assign sessions to the microlocations you have created.

**Other information:**
This is the screenshot of my **scheduler tab**.
image

**System information:** 
I am facing this on eventyay.com as well as on my local setup.

```
OS : Linux
```

```
Python 2.7.13
```

LuD1161:
#### Checklist

The scheduler was not showing events with **"confirmed"** states.

- Added an or query to **_get_sessions_** method in **data_getter.py**

Fixes #3623

### DIFF
--- a/app/helpers/data_getter.py
+++ b/app/helpers/data_getter.py
@@ -216,16 +216,24 @@ class DataGetter(object):
         return Track.query.filter_by(event_id=get_event_id())
 
     @staticmethod
-    def get_sessions(event_id, state='accepted'):
+    def get_sessions(event_id, state=None):
         """
         :param state: State of the session
         :param event_id: Event id
         :return: Return all Sessions objects with Event id
         """
-        return Session.query.filter_by(
-            event_id=event_id,
-            state=state
-        ).filter(Session.deleted_at.is_(None))
+        if state is None:
+            return Session.query.filter_by(
+                event_id=event_id
+            )\
+            .filter(or_(Session.state == 'accepted', Session.state == 'confirmed')) \
+            .filter(Session.deleted_at.is_(None))
+        else:
+            return Session.query.filter_by(
+                event_id=event_id,
+                state=state
+            )\
+            .filter(Session.deleted_at.is_(None))
 
     @staticmethod
     def get_image_sizes():
@@ -277,10 +285,12 @@ class DataGetter(object):
         :return: Return all Sessions objects with the current user as a speaker
         """
         if upcoming_events:
-            return Session.query.filter(Session.speakers.any(Speaker.user_id == (login.current_user.id if not user_id else int(user_id)))).filter(
+            return Session.query.filter(Session.speakers.any(
+                Speaker.user_id == (login.current_user.id if not user_id else int(user_id)))).filter(
                 Session.start_time >= datetime.datetime.now()).filter(Session.deleted_at.is_(None))
         else:
-            return Session.query.filter(Session.speakers.any(Speaker.user_id == (login.current_user.id if not user_id else int(user_id)))).filter(
+            return Session.query.filter(Session.speakers.any(
+                Speaker.user_id == (login.current_user.id if not user_id else int(user_id)))).filter(
                 Session.start_time < datetime.datetime.now()).filter(Session.deleted_at.is_(None))
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


app/helpers/data_getter.py
```diff
@@ -216,16 +216,15 @@ def get_tracks_by_event_id():
         return Track.query.filter_by(event_id=get_event_id())
 
     @staticmethod
-    def get_sessions(event_id, state='accepted'):
```
niranjan94: The `state` parameter is there for a reason. It allows us to re-use this method anywhere else we want. Maybe in someplace we want only rejected sessions. At that time we want to reuse this same method by calling `get_sessions(1, 'rejected')`. 

Please don't just blindly remove things like this without considering why it is there in the first place.

LuD1161: @niranjan94 
what do you suggest ?
If I keep that parameter then how do I pass two parameters to check with , cause in scheduler it should have both types of states **confirmed** as well as **accepted**.
And also we can't pass state as an array cause it may again break somewhere.

niranjan94: You can keep the parameter with default value as `None`. If it is `None` then show accepted and confirmed. Else show what the parameter says.

app/views/public/event_detail.py
```diff
@@ -141,8 +141,9 @@ def display_event_schedule(identifier):
         abort(404)
     tracks = DataGetter.get_tracks(event.id)
     accepted_sessions_count = get_count(DataGetter.get_sessions(event.id))
+    confirmed_sessions_count = get_count(DataGetter.get_sessions(event.id, state='confirmed'))
```
niranjan94: Now you're running two queries when it could be done in one.

LuD1161: @niranjan94 
What do you suggest ?
Should I make only one query getting all the sessions and then traversing till I get the first confirmed or accepted state session , then break the loop and set `accepted_sessions_count = 1` .
Or something else ?

niranjan94: NO! No ... Please no. Why can't you do the way I suggested in the previous comment ? (regarding the `get_sessions` method)

LuD1161: @niranjan94
I didn't see the comment upto now, will do it right away .

app/helpers/data_getter.py
LuD1161: @niranjan94 
For other uses where we need both `event_id` as well as `state` shouldn't we use `get_sessions_by_state_and_event_id` ( [https://github.com/fossasia/open-event-orga-server/blob/development/app/helpers/data_getter.py#L191](https://github.com/fossasia/open-event-orga-server/blob/development/app/helpers/data_getter.py#L191) ), I mean this function is being exclusively used for this particular function, isn't it ?

niranjan94: There are a lot of duplicate functions doing similar things. We should be removing them and they will be removed during the nextgen refactor. `get_sessions` function is preferred to `get_sessions_by_state_and_event_id`.

LuD1161:
@niranjan94 
Can you review it ?

app/helpers/data_getter.py
niranjan94: @LuD1161 

> You can keep the parameter with default value as None. If it is None then show accepted and confirmed. Else show what the parameter says.

The above was my comment.

> You can keep the parameter with default value as None.

This 👆 is done. ✅ 

> If it is None then show accepted and confirmed. Else show what the parameter says.

Where is this 👆 ?

LuD1161:
@niranjan94 
Sorry , forgot to add it.
I think it's good now.

app/helpers/data_getter.py
```diff
@@ -216,16 +216,22 @@ def get_tracks_by_event_id():
...
         """
         :param state: State of the session
         :param event_id: Event id
         :return: Return all Sessions objects with Event id
         """
-        return Session.query.filter_by(
-            event_id=event_id,
-            state=state
-        ).filter(Session.deleted_at.is_(None))
+        if state is None:
```
niranjan94: You can just use `if not state:`  right ? or is there any reason you use `if state is None:`

app/helpers/data_getter.py
```diff
@@ -216,16 +216,22 @@ def get_tracks_by_event_id():
...
+            return Session.query.filter_by(
+                event_id=event_id). \
+                filter(or_(Session.state == 'accepted', Session.state == 'confirmed')). \
+                filter(Session.deleted_at.is_(None))
+        else:
+            return Session.query.filter_by(
+                event_id=event_id, state=state). \
```
niranjan94: Please follow a consistent code style 

```python
return Session.query.filter_by(
    event_id=event_id,
    state=state
).filter(Session.deleted_at.is_(None))
```

app/helpers/data_getter.py
```diff
@@ -216,16 +216,22 @@ def get_tracks_by_event_id():
...
         :param state: State of the session
         :param event_id: Event id
         :return: Return all Sessions objects with Event id
         """
-        return Session.query.filter_by(
-            event_id=event_id,
-            state=state
-        ).filter(Session.deleted_at.is_(None))
+        if state is None:
+            return Session.query.filter_by(
```
niranjan94: Please follow a consistent code style.

app/helpers/data_getter.py
LuD1161: @niranjan94 
Nothing specific

app/helpers/data_getter.py
LuD1161: @niranjan94 
ok

app/helpers/data_getter.py
LuD1161: @niranjan94 
Will keep that in mind :smile: .

app/helpers/data_getter.py
niranjan94: Why do you keep putting the `.` at the end and then using `\` to continue on the next line when the `.` can be put in the next line ? 

```python
return Session.query.filter_by(
    event_id=event_id
)
.filter(or_(Session.state == 'accepted', Session.state == 'confirmed')))
.filter(Session.deleted_at.is_(None))
```

LuD1161:
@niranjan94 
is it wrong to write that way ?

niranjan94:
It's not wrong. `\` affects readability and make the code look ... _ugly_

LuD1161:
@niranjan94 
Does this look good ?
```
        if state is None:
            return Session.query.filter_by(
                event_id=event_id
            )\
            .filter(or_(Session.state == 'accepted', Session.state == 'confirmed')) \
            .filter(Session.deleted_at.is_(None))
        else:
            return Session.query.filter_by(
                event_id=event_id,
                state=state
            )\
            .filter(Session.deleted_at.is_(None))
```
Cause if I remove **\**  before the filter line , python gives syntax error.

niranjan94:
Cool. Looks good.

LuD1161:
@niranjan94 
I think now it's all good.

PR #3649, opened 2017-05-23
Made celery task for csv export for ticketing data

### LINKED ISSUE #3588: Error exporting attendees as csv
When trying to export the list of attendees as CSV, eventyay reports an internal server error.

The url visited is: https://eventyay.com/events/<event_no>/tickets/attendees/csv

Exporting as PDF Document works.

### LINKED ISSUE #3653: Export order csv and attendee csv through celery
- [ ] bug report
- [x] feature request
- [ ] support request 

**Current behavior:**
Right now it makes an http response and returns it to be downloaded. However this results in timeout and doesn't work for large data. As we can see in #3588

SaptakS:
#### Checklist

Fixes CSV export of attendees and also improved CSV export for orders

The CSV export now works through celery instead of HTTP response. So this handles the timeout problem for a long amount of data.

Fixes #3588 
Fixes #3653

### DIFF
--- a/app/helpers/exporters/attendee_csv.py
+++ b/app/helpers/exporters/attendee_csv.py
@@ -1,0 +1,21 @@
+from app.helpers.ticketing import TicketingManager
+
+class AttendeeCsv:
+
+    @staticmethod
+    def export(event_id):
+        (_, event_id, holders, _, _) = TicketingManager.get_attendee_export_info(event_id=event_id)
+        headers = 'Order#,Order Date, Status, First Name, Last Name, Email, Country,' \
+                'Payment Type, Ticket Name, Ticket Price, Ticket Type'
+
+        fields = ('order_invoice', 'created_at', 'status', 'firstname', 'lastname', 'email',
+                  'country', 'paid_via', 'ticket_name', 'ticket_price', 'ticket_type')
+        rows = [headers]
+        for holder in holders:
+            if holder['status'] != "deleted":
+                columns = (str(holder.get(f, '')) for f in fields)
+                rows.append(','.join(columns))
+
+        csv_content = '\n'.join(rows)
+
+        return csv_content
--- a/app/helpers/tasks.py
+++ b/app/helpers/tasks.py
@@ -8,6 +8,8 @@ from app.helpers.versioning import strip_tags
 from app.helpers.exporters.pentabarfxml import PentabarfExporter
 from app.helpers.exporters.ical import ICalExporter
 from app.helpers.exporters.xcal import XCalExporter
+from app.helpers.exporters.attendee_csv import AttendeeCsv
+from app.helpers.exporters.order_csv import OrderCsv
 from app.helpers.storage import UPLOAD_PATHS, upload, UploadedFile
 from app.helpers.data_getter import DataGetter
 from app.helpers.data import save_to_db
@@ -94,3 +96,35 @@ def export_xcal_task(event_id):
     xcal_file = UploadedFile(file_path=file_path, filename=filename)
     event.xcal_url = upload(xcal_file, UPLOAD_PATHS['exports']['xcal'].format(event_id=event_id))
     save_to_db(event)
+
+
+@celery.task(name='export.attendee.csv')
+def export_attendee_csv_task(event_id):
+    try:
+        os.mkdir(app.config['TEMP_UPLOADS_FOLDER'])
+    except OSError as exc:
+        if exc.errno != errno.EEXIST:
+            raise exc
+    filename = "attendees.csv"
+    file_path = app.config['TEMP_UPLOADS_FOLDER'] + "/" + filename
+    with open(file_path, "w") as temp_file:
+        temp_file.write(AttendeeCsv.export(event_id))
+    attendee_csv_file = UploadedFile(file_path=file_path, filename=filename)
+    attendee_csv_url = upload(attendee_csv_file, UPLOAD_PATHS['exports']['csv'].format(event_id=event_id))
+    return attendee_csv_url
+
+
+@celery.task(name='export.order.csv')
+def export_order_csv_task(event_id):
+    try:
+        os.mkdir(app.config['TEMP_UPLOADS_FOLDER'])
+    except OSError as exc:
+        if exc.errno != errno.EEXIST:
+            raise exc
+    filename = "order.csv"
+    file_path = app.config['TEMP_UPLOADS_FOLDER'] + "/" + filename
+    with open(file_path, "w") as temp_file:
+        temp_file.write(OrderCsv.export(event_id))
+    order_csv_file = UploadedFile(file_path=file_path, filename=filename)
+    order_csv_url = upload(order_csv_file, UPLOAD_PATHS['exports']['csv'].format(event_id=event_id))
+    return order_csv_url
--- a/app/helpers/exporters/order_csv.py
+++ b/app/helpers/exporters/order_csv.py
@@ -1,0 +1,29 @@
+from app.helpers.ticketing import TicketingManager
+
+class OrderCsv:
+
+    @staticmethod
+    def export(event_id):
+        orders = TicketingManager.get_orders(event_id)
+        headers = 'Order#,Order Date, Status, Payment Type, Total Amount, Quantity, Discount Code,' \
+                'First Name, Last Name, Email \n'
+
+        rows = [headers]
+        for order in orders:
+            if order.status != "deleted":
+                column = []
+                column.append(str(order.get_invoice_number()))
+                column.append(str(order.created_at) if order.created_at else '')
+                column.append(str(order.status) if order.status else '')
+                column.append(str(order.paid_via) if order.paid_via else '')
+                column.append(str(order.amount) if order.amount else '')
+                column.append(str(order.get_tickets_count()))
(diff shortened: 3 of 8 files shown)


niranjan94:
Do it like the current export task. Use the existing API endpoint to get the task status using the task ID you get when you start a task. (You can do a continuous polling using js)

SaptakS:
@niranjan94 I tried doing that and opened the task url in browser and it was giving me a 500 internal server error. Can you please review my code and tell if I am doing something wrong?

niranjan94:
@SaptakS I'm afk right now. Will review in sometime.

niranjan94:
@SaptakS what's the the error ? WHat's displayed in the logs ?

niranjan94:
@SaptakS as far as I can tell ... the code looks fine. So, It would be helpful if you tell what the error is you get ?

hongquan:
The `export_attendee_csv_task` is a Celery task, means that it is executed in Celery worker, can be a separate thread, separate process, or even separate machine from the view function that Flask is building HTTP response and send to browser.
You have to imagine the flow  is like this:

```
View function receive HTTP request -->  Process request --> Build response and return to browser.
                                                  |
                                                  +--> Send event_id to Celery task -> Celery process -> Result.
```

So, you can see, if the Celery tasks build a HTTP response, it has no way to return that response to browser.
And you should note that, Celery task is not a view function, you should not use it to build HTTP response. Celery task should be only to create some intermediate data. The result of the task then is saved to a result backend, which can be Redis or RabbitMQ, depending on your config.

In your case, even you define a Celery task to make a CSV file, you should not return the whole CSV content in `return` statement. Because all the file will go to Redis! Instead, you should save the file to somewhere and return the file path.

My experience in the case of using Celery to build CSV is:

- Case 1: Ask user email before processing. This email address is passed to Celery task. The task, after building CSV, will send the file to that email.

- Case 2: Create an Ajax endpoint (URL) to track the state of Celery task. This URL is polled by your JS script in front-end. When the Celery task is done, it inserts the file URL to its state. Your JS code then see the file URL and redirect browser to that URL to download the file.

Other thing, your code to build CSV looks not clear and _Pythonic_. You should not use much string concat. Build your rows, columns content in `list` instead and use `','.join` to make string.

SaptakS:
> You should not use much string concat.

Why is this non pythonic or not clear?

> Build your rows, columns content in list instead and use `','.join` to make string.

Why is this necessary? And as far as I know, .join() is slower compared to `+` when it is a concat of only two strings. 

@hongquan Can you please explain? Thanks.

hongquan:
> Why is this non pythonic or not clear?

First, the data is tabular (by column and row), so separate them in small parts is easier to read and manage. For example, you want to go back and forth the columns in the same rows, you can use index (`rows[1]`) when you still hold the `list`. You cannot do so when you already combine data to string `value1,value2,value3`.
> Why is this necessary? And as far as I know, .join() is slower compared to + when it is a concat of only two strings.
I read that `join` is faster. And please note that in your case, you don't concat two strings. You concat many string. This is CSV.

SaptakS:
> you want to go back and forth the columns in the same rows, you can use index (rows[1]) when you still hold the list. 

I agree but this is only applicable when you are reading or modifying columns in the same rows again and again. I don't think anything as such is happening or is necessary. If so I will surely use list.

> I read that join is faster.

For concatenation of too many strings, yes. But here at a particular time, only single concatenation is happening.

>  And please note that in your case, you don't concat two strings.

At a time I always concatenate only two strings. And since these are not generators so obviously the operation gets executed there itself. So `+` would require less time in my case of example. But first I will check if list implementation is at all needed and then do accordingly.

hongquan:
> But here at a particular time, only single concatenation is happening.

How is single concatenation? You + column 1 to column 2. Then you + last value to column 3, then + last value column 4, then + 1 row to next row, so on...

SaptakS:
Because both the operands are strings themselves. So each time it is a single concatenation. There are multiple single concatenation. Also, thinking that you have to check whether the value exists, I think the .join() operation will actually reduce readability. Please let me know what you think. And if you can suggest some way or give an example of how I should use .join in this particular scenario that maintains readability.

hongquan:
```python3
fields = ('order_invoice', 'firstname', 'lastname')
rows = [headers]
for holder in holders:
	columns = []
	for f in fields:
		columns.append(holder.get(f, ''))
	rows.append(','.join(columns))

csv_content = ','.join(rows)
```

Other version, which is more of _functional programming_ style:

```python
fields = ('order_invoice', 'firstname', 'lastname')
rows = [headers]
for holder in holders:
	columns = (holder.get(f, '') for f in fields)   # This is a generator, not list
	rows.append(','.join(columns))

csv_content = ','.join(rows)
```

Note that, if some field doesnot exist in `holder`, we cannot skip that column, or our CSV gets wrong format.

SaptakS:
@hongquan makes sense.

hongquan:
Now, even more _functional programming_, and use `tuple` instead of `list` to reduce overhead:

```python
fields = ('order_invoice', 'firstname', 'lastname')

def make_csv_row_from_holder(holder):
	# Note, we pass a generator, not intermediate list, to join
	return ','.join(holder.get(f, '') for f in fields)

rows = tuple(map(make_csv_row_from_holder, holders))

csv_content = ','.join((header,) + rows)
```

app/helpers/tasks.py
```diff
@@ -94,3 +96,36 @@ def export_xcal_task(event_id):
...
+    except OSError as exc:
+        if exc.errno != errno.EEXIST:
+            raise exc
+    filename = "order.csv"
+    file_path = app.config['TEMP_UPLOADS_FOLDER'] + "/" + filename
+    with open(file_path, "w") as temp_file:
+        temp_file.write(OrderCsv.export(event_id))
+    order_csv_file = UploadedFile(file_path=file_path, filename=filename)
+    order_csv_url = upload(order_csv_file, UPLOAD_PATHS['exports']['csv'].format(event_id=event_id))
+    print order_csv_url
```
hongquan: Remove this print

SaptakS: Sorry. Missed it. Thanks for pointing.

SaptakS: Done.

hongquan:
Please remove `print` and `console.log` from production code. They are for development but not production.

hongquan:
It seems that all export files are saved to the same folder. What if many users time request export nearly at the same time? Their files will be overwritten.

SaptakS:
@hongquan the folder is dependent on the event. So for a particular event, the information remains the same.

hongquan:
> the folder is dependent on the event. So for a particular event, the information remains the same.

Then what? The problem is still there. Files are still overwritten if 2 users are exporting the same event.

Imagine, user A request to export. The content of the file should be "A1A2B1B2". But when the file is being written "A1A2B",  the action from user B come and to the same file. The content becomes "A1A2BA1A2". Is it the one users want?

SaptakS:
@hongquan it doesn't append to the file. So if user B now sends a request, it replaces the entire content and overwrites it. And since the content is the same, overwriting doesn't create any problem.

hongquan:
> it doesn't append to the file

My example is just a simple, optimistic one, to let you imagine that we should not let 2 users work on the same file. But in fact it is worse: Two write action are doing in the same file will cause deadlock, or "race condition".

SaptakS:
@hongquan I do agree with you at that point... Right now the deadlock and OS mutex locks haven't been implemented in any of the exports. So this will cause problem in all the exports actually, right now. We need to find a way for all exports to handle this.

hongquan:
> We need to find a way for all exports to handle this.

Just add some random string to the path/filename, to make sure each user has her own file. 
We then have a periodic job (cron or systemd-timer) to clean those old files. We can define how long to consider "old". 1 hour for example.

SaptakS:
@hongquan I have created an [issue](https://github.com/fossasia/open-event-orga-server/issues/3668) to handle this situation.

hongquan:
Remove "for nothing" functions like this:

```python
def __init__(self):
    pass
```

Some of you code doesn't follow PEP8's indentation.

SaptakS:
@hongquan can you point the indentation errors by making inline comments? Would be really helpful. Thanks.

app/helpers/exporters/attendee_csv.py
```diff
@@ -0,0 +1,23 @@
...
+		pass
+
+	@staticmethod
+	def export(event_id):
+		(_, event_id, holders, _, _) = TicketingManager.get_attendee_export_info(event_id=event_id)
+		headers = 'Order#,Order Date, Status, First Name, Last Name, Email, Country,' \
+				'Payment Type, Ticket Name, Ticket Price, Ticket Type'
+
+		fields = ('order_invoice', 'created_at', 'status', 'firstname', 'lastname', 'email', \
+				  'country', 'paid_via', 'ticket_name', 'ticket_price', 'ticket_type')
```
hongquan: This one (indentation)

SaptakS: Okay. This isn't how it shows in my editor. Wait. I need to recheck this.

hongquan: This is one of the reasons why PEP8 recommends "space" over "tab". I personally prefer "tab", but...

PR #3680, opened 2017-05-28
Avoid race condition while exporting csv files

### LINKED ISSUE #3668: Handle race condition in all export files
**Current behavior:**
As pointed out by @hongquan [here](https://github.com/fossasia/open-event-orga-server/pull/3649#issuecomment-304088669) we can have race conditions in the different exports that we have right now since the exports are done using the same file in the same folder for a particular event.

**Suggestion:**
To avoid this we can follow like @hongquan has suggested here:

> Just add some random string to the path/filename, to make sure each user has her own file.
We then have a periodic job (cron or systemd-timer) to clean those old files. We can define how long to consider "old". 1 hour for example.

SaptakS:
#### Checklist

Avoid race condition while exporting csv files.

Fixes #3668

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -8,7 +8,7 @@ warnings.simplefilter('ignore', ExtDeprecationWarning)
 from pytz import utc
 
 from app.helpers.scheduled_jobs import send_mail_to_expired_orders, empty_trash, send_after_event_mail, \
-    send_event_fee_notification, send_event_fee_notification_followup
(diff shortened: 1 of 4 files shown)


hongquan:
To reduce the risk of duplicate random name, please use `uuid.uuid1()`, which counts the time of execution.

hongquan:
Do you plan to create File Clean Job here or other PR?

SaptakS:
@hongquan I will create file clean also in this in PR.

PR #3721, opened 2017-06-03
Basic api implementation for call-for-paper

### LINKED ISSUE #3715: Implement call-for-speaker api with proper permissions and relationships

magdalenesuo:
#### Checklist

- [X] My branch is up-to-date with the Upstream `nextgen` branch.

Basic api implementation for call-for-paper

- proper endpoints
- relation with events alone
-

Fixes #3715
screenshot from 2017-06-04 00-44-24

### DIFF
--- a/app/api/call_for_papers.py
+++ b/app/api/call_for_papers.py
@@ -1,0 +1,110 @@
+from flask_rest_jsonapi import ResourceDetail, ResourceList, ResourceRelationship
+from marshmallow_jsonapi.flask import Schema, Relationship
+from marshmallow_jsonapi import fields
+from app.models import db
+from app.api.helpers.permissions import jwt_required
+from sqlalchemy.orm.exc import NoResultFound
+from flask_rest_jsonapi.exceptions import ObjectNotFound
+from app.models.event import Event
+from app.models.call_for_paper import CallForPaper
+import pytz
+from datetime import datetime
+import dateutil.parser
+
+
+class CallForPaperSchema(Schema):
+    """
+    Api Schema for cfp model
+    """
+    class Meta:
+        """
+        Meta class for cfp Api Schema
+        """
+        type_ = 'call_for_paper'
+        self_view = 'v1.call_for_paper_detail'
+        self_view_kwargs = {'id': '<id>'}
+
+    id = fields.Str(dump_only=True)
+    announcement = fields.Str(required=True)
+    start_date = fields.DateTime(required=True)
+    end_date = fields.DateTime(required=True)
+    timezone = fields.Str(default="UTC")
+    hash = fields.Str()
+    privacy = fields.String()
+    event = Relationship(attribute='event',
+                         self_view='v1.call_for_paper_event',
+                         self_view_kwargs={'id': '<id>'},
+                         related_view='v1.event_detail',
+                         related_view_kwargs={'call_for_paper_id': '<id>'},
+                         schema='EventSchema',
+                         type_='event')
+
+
+class CallForPaperList(ResourceList):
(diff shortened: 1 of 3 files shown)


app/api/call_for_papers.py
```diff
@@ -0,0 +1,106 @@
...
+        if view_kwargs.get('event_id') is not None:
+            query_ = query_.join(Event).filter(Event.id == view_kwargs['event_id'])
+        return query_
+
+    def before_create_object(self, data, view_kwargs):
+        if view_kwargs.get('event_id') is not None:
+            event = self.session.query(Event).filter_by(id=view_kwargs['event_id']).one()
+            start_date = data['start_date'].replace(tzinfo=pytz.timezone(data['timezone']))
+            end_date = data['end_date'].replace(tzinfo=pytz.timezone(data['timezone']))
+            print start_date.isoformat(), end_date.isoformat()
```
magdalenesuo: @niranjan94 @shubham-padia @SaptakS I've tried doing this to make timezone aware for the date fields. While I print start_date and end_date, I am getting perfect result like this:
zzzz
But when i do a GET, I get different dates.:
screenshot from 2017-06-03 13-09-51

Is there a solution for this ?

shubham-padia: @magdalenesuo before_create_object will only be called for post methods. One thing may work, not sure, try using after_get(result) and modify the result at last.

magdalenesuo: @shubham-padia
> before_create_object will only be called for post methods

Yes, I was trying to modify the date time during POST itself.

> One thing may work, not sure, try using after_get(result) and modify the result at last.

Yes, okay. Will try it out.

app/api/call_for_papers.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def after_get(self, result):
+        tz = pytz.timezone(result['data'][0]['attributes']['timezone'])
+        start_date_str = result['data'][0]['attributes']['start_date']
+        start_date_datetime = dateutil.parser.parse(start_date_str)
+        start_date_datetime = tz.localize(start_date_datetime.replace(tzinfo=None))
+        result['data'][0]['attributes']['start_date'] = start_date_datetime.isoformat()
+        end_date_str = result['data'][0]['attributes']['end_date']
+        end_date_datetime = dateutil.parser.parse(end_date_str)
+        end_date_datetime = tz.localize(end_date_datetime.replace(tzinfo=None))
```
shubham-padia: I think the logic for tzinfo=none should be done 
in before_create_object and its equivalent method for patch requests, because currently also we are storing the date without the offset.

magdalenesuo: @shubham-padia Hi. 

I think this is not necessary. While I print the date  or GET, I find that the DateTime fields get converted to utc offset even if we store it without offset.
So changing tzinfo=none will do nothing since it will return in timezone aware format only(UTC, default).

It is not needed for PATCH patch requests because it stays timezone unaware and stored like that only.

shubham-padia: Okay

LuD1161:
@magdalenesuo @shubham-padia @SaptakS 
In `app/api/events.py` ( [here](https://github.com/fossasia/open-event-orga-server/pull/3721/commits/07bd0f6b08a909205268aaf6336f68166502e9fe#diff-364bbff4749f94ed8f7082208c36cf78R71)  )  it should be 'type_=call_for_paper' instead of `type_=sponsor`.
I have updated it in [PR#3708](https://github.com/fossasia/open-event-orga-server/pull/3708) .

PR #3724, opened 2017-06-03
Added event copyright API

### LINKED ISSUE #3714: Implement Event Copyright API with proper permissions and relationships

SaptakS:
#### Checklist

Adds the API endpoint for Event Copyright information

Fixes #3714

### DIFF
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -12,6 +12,7 @@ from app.api.sponsors import SponsorList, SponsorDetail, SponsorRelationship
 from app.api.tracks import TrackList, TrackDetail, TrackRelationship
 from app.api.call_for_papers import CallForPaperList, CallForPaperDetail, CallForPaperRelationship
 from app.api.image_sizes import ImageSizeList, ImageSizeDetail
+from app.api.event_copyright import EventCopyrightList, EventCopyrightDetail, EventCopyrightRelationship
 
 
 api_v1 = Blueprint('v1', __name__, url_prefix='/v1')
@@ -22,10 +23,11 @@ api.route(UserList, 'user_list', '/users')
 api.route(UserDetail, 'user_detail', '/users/<int:id>', '/notifications/<int:notification_id>/user')
 api.route(UserRelationship, 'user_notification', '/users/<int:id>/relationships/notifications')
 
-#notifications
+# notifications
 api.route(NotificationList, 'notification_list', '/notifications', '/users/<int:id>/notifications')
 api.route(NotificationDetail, 'notification_detail', '/notifications/<int:id>')
-api.route(NotificationRelationship, 'notification_user', '/notifications/<int:id>/relationships/user')
+api.route(NotificationRelationship, 'notification_user',
+          '/notifications/<int:id>/relationships/user')
 
 # image_sizes
 api.route(ImageSizeList, 'image_size_list', '/image_sizes')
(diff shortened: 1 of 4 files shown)


app/api/__init__.py
```diff
@@ -28,12 +29,14 @@
 api.route(EventList, 'event_list', '/events')
 api.route(EventDetail, 'event_detail', '/events/<int:id>', '/tickets/<int:ticket_id>/event',
           '/microlocations/<int:microlocation_id>/event', '/social_links/<int:social_link_id>/event',
-          '/sponsors/<int:sponsor_id>/event', '/tracks/<int:track_id>/event', '/call_for_papers/<int:call_for_paper_id>/event')
+          '/sponsors/<int:sponsor_id>/event', '/tracks/<int:track_id>/event',
+          '/event_copyright/<int:copyright_id>/event', '/call_for_papers/<int:call_for_paper_id>/event')
```
shubham-padia: I'm not sure about this endpoint as event_copyright should exist only as a relationship to event

app/api/social_links.py
```diff
@@ -39,6 +39,7 @@ def before_create_object(self, data, view_kwargs):
             event = self.session.query(Event).filter_by(id=view_kwargs['id']).one()
             data['event_id'] = event.id
 
+    view_kwargs = True
```
shubham-padia: @SaptakS What does this do? just curious, might help me somewhere.

app/api/__init__.py
SaptakS: This makes complete sense I think because this basically gives the URL for the endpoint to get the event that is related to a particular event_copyright. If /event_copyright/<int:copyright_id> makes sense then this also makes sense.

shubham-padia: okay, afaik from what I asked @niranjan94 in https://github.com/fossasia/open-event-orga-server/pull/3718#discussion_r119982046, `/event_copyright/int:copyright_id` should'nt be there also.

app/api/social_links.py
SaptakS: This is needed when you remove the /event_copyright endpoint as given in the [docs](http://flask-rest-jsonapi.readthedocs.io/en/latest/resource_manager.html#resourcelist):

> view_kwargs:	if you set this flag to True view kwargs will be used to compute the list url. If you have a list url pattern with parameter like that: /persons/<int:id>/computers you have to set this flag to True

You can see a similar line in the [last pull request](https://github.com/fossasia/open-event-orga-server/pull/3721/files#diff-256d8e06babe7f96e6b5af3457d1bf45R70) by @magdalenesuo as well. Took reference from there.

shubham-padia: okay, thanks.

app/api/__init__.py
SaptakS: > /session_types shouldn't exist

@niranjan94  never said /session_types/<int:sessiontype_id> shouldn't exist. Infact what I discussed with him, he said patch, delete should be functional on a link like /session_types/<int:sessiontype_id> . @niranjan94 your comments?

app/api/__init__.py
```diff
@@ -72,6 +76,13 @@
 api.route(TrackRelationship, 'track_sessions', '/tracks/<int:id>/relationships/sessions')
 api.route(TrackRelationship, 'track_event', '/tracks/<int:id>/relationships/event')
 
+# event_copyright
+api.route(EventCopyrightList, 'event_copyright_list', '/events/<int:id>/event_copyright')
```
shubham-padia: @SaptakS  Event_copyright is one to one right, list is not needed afaik and using <int:id> instead of <int:event_id> would give list according to the id of event_copyright and not the event, the `'/events/<int:event_id>/event_copyright'` in the line below does the same work instead

SaptakS: Yes. It is one to one alright. But I am not sure whether we can remove this exactly. I will test this out.

> using int:id instead of int:event_id would give list according to the id of event_copyright and not the event,

It will give according to the event only because of the logic implemented in before_create_object() and query()

app/api/__init__.py
shubham-padia: :white_check_mark: Resolved

app/api/__init__.py
shubham-padia: :white_check_mark: Resolved

PR #3794, opened 2017-06-13
Missing session event relationship added

### LINKED ISSUE #3779: Relationship missing between Session and Event
- [x] bug report
- [ ] feature request

There should be a relationship between session and event. Sessions always belong to a particular event.

SaptakS:
Along with session-event relationship, relationship for microlocation also fixed and removed some wrong API endpoint related to sessions

#### Checklist

* Adds missing session event relationships
* Fixes sessions microlocation endpoints
* Removes wrong session endpoints
* PEP8 fixes in most session related files in api and models

Fixes #3779

### DIFF
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -31,7 +31,7 @@ api.route(UserRelationship, 'user_notification', '/users/<int:id>/relationships/
 api.route(UserRelationship, 'user_event_invoices', '/users/<int:id>/relationships/event-invoices')
 
 # notifications
-api.route(NotificationList, 'notification_list','/users/<int:id>/notifications')
+api.route(NotificationList, 'notification_list', '/users/<int:id>/notifications')
 api.route(NotificationDetail, 'notification_detail', '/notifications/<int:id>')
 api.route(NotificationRelationship, 'notification_user',
           '/notifications/<int:id>/relationships/user')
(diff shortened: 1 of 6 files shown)


app/api/events.py
```diff
@@ -36,13 +38,13 @@ class Meta:
...
-                          many=True,
-                          type_='ticket')
+                           self_view='v1.event_ticket',
+                           self_view_kwargs={'id': '<id>'},
+                           related_view='v1.ticket_detail',
+                           related_view_kwargs={'event_id': '<id>'},
+                           schema='TicketSchema',
+                           many=True,
+                           type_='ticket')
     microlocation = Relationship(attribute='microlocation',
```
shubham-padia: Can you rename this to `microlocations` while you're at it.

SaptakS: Sure

SaptakS: Resolved.

app/api/__init__.py
```diff
@@ -86,14 +88,16 @@
 
 # sessions
 api.route(SessionList, 'session_list', '/events/<int:id>/sessions', '/tracks/<int:track_id>/sessions',
```
poush: Please remove this old one

app/api/__init__.py
```diff
@@ -73,10 +74,11 @@
 api.route(EventRelationship, 'event_tax', '/events/<int:id>/relationships/tax')
 api.route(EventRelationship, 'event_event_invoice', '/events/<int:id>/relationships/event-invoices')
 api.route(EventRelationship, 'event_discount_code', '/events/<int:id>/relationships/discount-codes')
+api.route(EventRelationship, 'event_session', '/events/<int:id>/relationships/sessions')
 
 # microlocations
 api.route(MicrolocationList, 'microlocation_list', '/microlocations',
-          '/events/<int:id>/microlocations', '/sessions/<int:id>/microlocations')
+          '/events/<int:id>/microlocations')
```
poush: I understand we are doing this in many places. But I'm not sure why we are not making convention for 
```/events/<event_id>/microlocations``` 
In case we get more than one entries here like we have in ```SessionList``` route then we cannot just say `id` instead we will have ```event_id``` so why not use it everywhere even for a single entry. And also at many places, we are using ```event_id``` even for the single entry. So as per my suggestion, we should change all entries to ```event_id``` ( related type )

app/api/__init__.py
poush: `'/events/<int:id>/sessions'` is repeated

app/api/__init__.py
SaptakS: @poush can you please mention this point [here](https://github.com/fossasia/open-event-orga-server/issues/3712)? Will fix it along with that. I know there are many wrong and repetitive endpoints as well.

app/api/__init__.py
SaptakS: I will rather remove the new one... Thanks for pointing out

SaptakS: Resolved

Issue #3738, opened 2017-06-06
Implement image and file upload API


niranjan94:
### Image upload:
`POST /v1/upload/image` - Accepts the following payload

```
{
    "data": "data:image/png;base64,iVBORw0KGgoAAAANS"
}
```

Decodes the data. Stores it in the preferred storage (in a temp storage area). Returns the URL to that.

```
{
    "url": "https://xyz.storage.com/asd/fgh/hjk/12332233.png"
}
```

This is just a temporary image store. Now, when the user saves the event, when the `PATCH` or `POST` is made to the event API, the image URL is sent as a `original_image_url`. When you process the data, if that URL has changed from what was before, do what you have to do to the image. Download from the URL. Process it. Store in different sizes. Get the new URLs for the different sizes. Add that info to the other image url fields. 

---

### Generic File upload:
`POST /upload/file` - Accepts a `multipart/form-data` payload. Where the file is directly uploaded under the key `file` if single file. Else `files` if a set of files. Store it in the prefered storage and return the full URL(s) to it as 

_for single file:_
```
{
    "url": "https://xyz.storage.com/asd/fgh/hjk/12332233.docx"
}
```

_for multiple files:_
```
{
    "url": [
        "https://xyz.storage.com/asd/fgh/hjk/12332233.docx",
        "https://xyz.storage.com/asd/fgh/hjk/66777777.ppt"
    ]
}
```

Urls should be in the same order as the `files` array files.

PR #3823, opened 2017-06-17
Image upload endpoint

SaptakS:
Image uploading to return a temp URL is done in this PR related to #3738 

Sample Input: 
```
{"data": "data:image/gif;base64,R0lGODlhEAAQAMQAAORHHOVSKudfOulrSOp3WOyDZu6QdvCchPGolfO0o/XBs/fNwfjZ0frl3/zy7////wAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACH5BAkAABAALAAAAAAQABAAAAVVICSOZGlCQAosJ6mu7fiyZeKqNKToQGDsM8hBADgUXoGAiqhSvp5QAnQKGIgUhwFUYLCVDFCrKUE1lBavAViFIDlTImbKC5Gm2hB0SlBCBMQiB0UjIQA7"}
```

Sample Output:
```
{"url":"/serve_static/media/temp/images/6e35ba8d-0652-489f-9f2f-c68b967bf5f4/Z1JjdnJHeV/1d7965cd-3098-4c81-b532-3cae2185da0b.png"}
```
And the image gets saved in the /static/media/temp/images.. folder.


poush:
You are saving it to temporary storage inside the project while niranjan comment says to store it in preferred storage.
> Store it in the prefered storage and return the full URL(s) to it as

SaptakS:
Yup. That is the prefered storage for storing the temporary images, I think. @niranjan94 ?

poush:
I think it is actual preferred storage( like s3), especially If you look at the sample endpoints he provided. But still, let @niranjan94 confirm.

SaptakS:
@poush temporary image files were stored in the local storage so that a cron job can delete them. Since the temp files are supposed to be deleted after at max a day. We can do that for s3 or gs as well. But right now it is done in that way for orga-server. Also, if I don't mention a s3 or gs id in the admin panel it's anyway going to store in local system.

niranjan94:
Saptak. Like poush said, by temp I mean preferred storage. Since the url will also be persisted in the database as the original image url in most cases. Do one thing... We actually need both cases. One that temporarily stores in the local file system and one that stores in the preferred storage (not temporary). So have a get parameter like type=temp would store in local temp that is volatile. Anything else, store in preferred storage.

app/api/helpers/images.py
```diff
@@ -0,0 +1,16 @@
...
+from app.api.helpers.storage import UploadedFile
+from flask import current_app
+
+
+def get_image_file_name():
+    return str(uuid.uuid4())
+
+
+def uploaded_image(extension='.png', file_content=None):
+    filename = get_image_file_name() + extension
```
niranjan94: Get the file extension from the data string itself.

app/api/image_upload.py
```diff
@@ -0,0 +1,27 @@
...
+
+upload_routes = Blueprint('upload', __name__, url_prefix='/v1/upload')
+
+
+@jwt_required
+@upload_routes.route('/image', methods=['POST'])
+def upload_image():
+    image = request.json['data']
+    image_file = uploaded_image(file_content=image)
+    force_local request.args.get('force-local', 'false')
```
niranjan94: call it `force_local`. Let's follow `snake_case` for all GET parameters.

PR #3836, opened 2017-06-20
Added Image Upload endpoint docs

SaptakS:
#### Checklist

- Added API docs for image upload endpoint
- Fixed the URLs for the endpoints. The prefix `/v1/` was missing

#3838

### DIFF
--- a/docs/api/api_blueprint.apib
+++ b/docs/api/api_blueprint.apib
@@ -44,7 +44,7 @@ For using the API you need(mostly) to register as an user. Registering gives you
 | `email` | Email of the user | string | **yes** |
 
 
-## Users Collection [/users{?page%5bsize%5d,page%5bnumber%5d,sort,filter}]
+## Users Collection [/v1/users{?page%5bsize%5d,page%5bnumber%5d,sort,filter}]
 + Parameters
     + page%5bsize%5d (optional, integer, `10`) - Maximum number of resources in a single paginated response.
     + page%5bnumber%5d (optional, integer, `2`) - Page number to fetched for the paginated response.
@@ -144,7 +144,7 @@ Create a new user using an email, password and an optional name.
             }
         }
 
-## User Details [/users/{user_id}]
+## User Details [/v1/users/{user_id}]
 + Parameters
     + user_id: 1 (integer) - ID of the user in the form of an integer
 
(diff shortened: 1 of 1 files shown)


docs/api/api_blueprint.apib
```diff
@@ -5251,3 +5250,36 @@ To update or get any attribute of this data layer, you will need admin access. H
...
+### Upload an Image in temporary location [POST]
+
++ Parameters
+    + force_local (optional, string, `true`) - Force storage of temporary image in local
+
++ Request
+
+    + Headers
+
+            Content-Type: application/json
```
poush: Can we change the headers to `application/vnd.api+json` so that we maintain the consistency or it is irrelevant here?

niranjan94: `application/vnd.api+json` refers to json-api spec compatible response. This isn't

SaptakS: No. Both the response and request are in the `application/json` format and not in `application/vnd.api+json` format since if you check my PR you will find that this URL is implemented separately from that of the rest jsonapi format.

poush: Okay.

shubham-padia:
Shouldn't changing the host to api.eventyay.com/v1 in the apib file do the the same instead of appending /v1 at each endpoint. 
@niranjan94 Will cahnging that affect the deployment of docs in any way

niranjan94:
@shubham-padia as said by saptak a while back

> I didn't update the host option directly because the /auth/session would then give a wrong URL. And the format keyword does denote the version of the API but it won't add the prefix.

SaptakS:
@shubham-padia there are endpoints like `/auth/session` as well which don't have `/v1/` as prefix. So I did that.

shubham-padia:
@SaptakS Sorry, I missed that comment, my bad.
But shouldn't /auth/session be moved out of the apib file since its out of version control.

niranjan94:
@shubham-padia our current documentation & blueprint is for the entire Open Event API as a whole. So, login should come inside it for now. In future, when we do have multiple API versions, then we can have multiple API blueprints like (common, v1, v2). For now that is overkill.

shubham-padia:
Okay, sounds good. Thanks for the explaination.

Issue #3695, opened 2017-05-31
Propose and setup basic testing for the api


shubham-padia:
Please comment down your implementation first.

SaptakS:
I think we can use simple python unittest. Assert the JSON responses.

niranjan94:
@shubham-padia please keep the issue updated with implementation proposal and status updates.

shubham-padia:
Already done in other issues. Closing this :)

PR #3850, opened 2017-06-21
Created Event type API and Model, relationship with event

### LINKED ISSUE #3831: Implement model & API for event types, event topic, event sub-topic
Currently, we have a static array of strings for event types, event topic, event sub-topic. Which is not very efficient and reliable on the longer run especially with a decoupled architecture. We also would like to support customisation by the admin and internalisation of strings too. 

Keeping that in mind, the following are required. 
- Model and API for event type & relationship with event
- Model and API for event topic & relationship with event, event sub-topic
- Model and API for event sub-topic & relationship with event, top
- All of the above with proper migration of existing information

The relationships are as follows:
- One to many relationship b/w event type -> event. (One type has many events)
- One to one relationship b/w event -> event type (One event belongs to one event type)
- One to many relationship b/w event topic -> event sub-topic (One event topic has many sub-topics)
- One to one relationship b/w event sub-topic -> event topic (One event sub-topic belongs to one topic)

Some special requirements for the event API
- The sub-topic, topic, type relationships should always be included in the event response. (Without having to explicitly specify the `include` query param) 

**This is a parent issue. Sub-issues to be created and worked upon.**

### LINKED ISSUE #3840: Implement model & API for event types
- [X] feature request

 Parent issue: #3831 

- **Model and API for event type & relationship with event**
 
**Relationship:**
- One to many relationship b/w event type -> event. (One type has many events)
- One to one relationship b/w event -> event type (One event belongs to one event type)

**DB:**
- `id` (Eg. 1) - Unique integer
- `name` (Eg. Appearance or Signing)
- `slug` (Eg. appearance-or-singing) - Unique string (Default: dash-cased version of the name)

magdalenesuo:
#### Checklist

Created Event type API and Model, relationship with event

- One to many relationship b/w event type -> event. (One type has many events)
`/event-types/<int:id>/relationships/events`

- One to one relationship b/w event -> event type (One event belongs to one event type)
`/events/<int:id>/relationships/event-type`

- **endpoints:**
- `/event-types` [POST and GET] for event type list
- `/event-types/<int:id>` and `/events/<int:event_id>/event-type` for event type detail
- `'/event-types/<int:event_type_id>/events'` for event list

Fixes #3840 
Partly fixes #3831


magdalenesuo:
@niranjan94 Done.
1) Made `slug` read-only and auto generated
2) Old files handling, this is how it is done:
          -  `Type` (string) from events table entered to `name` (string) of event_types table if that string is 
               not already present. Null values are ignored
          -   `Type` is converted to integer
          -   `Type` updated with the `id` of the `name` in event_type table.

app/models/event.py
```diff
@@ -74,7 +74,7 @@ class Event(db.Model):
     role_invites = db.relationship('RoleInvite', back_populates='event')
     privacy = db.Column(db.String, default="public")
     state = db.Column(db.String, default="Draft")
-    type = db.Column(db.String)
+    type = db.Column(db.Integer, db.ForeignKey('event_types.id', ondelete='CASCADE'))
```
niranjan94: rename to `event_type_id`

app/models/event.py
```diff
@@ -101,6 +101,7 @@ class Event(db.Model):
     xcal_url = db.Column(db.String)
     is_sponsors_enabled = db.Column(db.Boolean, default=False)
     discount_code = db.relationship('DiscountCode', backref='events')
+    event_type = db.relationship('EventType', backref="event")
```
niranjan94: single quotes

app/models/event_type.py
```diff
@@ -0,0 +1,43 @@
...
+
+class EventType(db.Model):
+    """Event type object table"""
+
+    __tablename__ = 'event_types'
+
+    id = db.Column(db.Integer, primary_key=True, unique=True, autoincrement=True)
+    name = db.Column(db.String, nullable=False)
+    slug = db.Column(db.String, unique=True)
+    events = db.relationship('Event', backref="event_types")
```
niranjan94: single quotes

niranjan94: And shouldn't the back reference be `event_type` ? Since one event can have only one type

migrations/versions/6bd24e8f99f0_.py
```diff
@@ -0,0 +1,46 @@
...
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('event_types',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('name', sa.String(), nullable=False),
+    sa.Column('slug', sa.String(), nullable=True),
```
niranjan94: slug should not be nullable.

migrations/versions/6bd24e8f99f0_.py
```diff
@@ -0,0 +1,46 @@
...
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('id'),
+    sa.UniqueConstraint('slug')
+    )
+    op.execute('INSERT INTO event_types(name) SELECT DISTINCT type FROM events where not exists (SELECT 1 \
```
niranjan94: Slug also needs to be created

tests/unittests/object_mother.py
```diff
@@ -22,8 +22,7 @@ def get_event():
                      location_name='India',
                      topic='Travel & Outdoor',
                      sub_topic='Climbing',
-                     is_sessions_speakers_enabled=True,
-                     type='Camp, Trip, or Retreat')
```
magdalenesuo: @niranjan94  I removed this temporarily for travis to pass. I will make a proper test with event_type because this `type` is being used in most of the files, so either I will make an issue to fix the test or solve it here #3695

migrations/versions/6bd24e8f99f0_.py
```diff
@@ -0,0 +1,50 @@
...
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('name', sa.String(), nullable=False),
+    sa.Column('slug', sa.String(), nullable=False),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('id'),
+    sa.UniqueConstraint('slug')
+    )
+    op.alter_column('events', 'type', new_column_name='event_type_id')
+    op.alter_column('events_version', 'type', new_column_name='event_type_id')
+    op.execute('INSERT INTO event_types(name, slug) SELECT DISTINCT event_type_id, replace(event_type_id, \' \', \'-\') \
```
niranjan94: For slug replace & and comma also with hyphen and make lowercase

tests/unittests/object_mother.py
niranjan94: No problem

migrations/versions/6bd24e8f99f0_.py
magdalenesuo: @niranjan94  If I replace `&` and `,`, won't (for example) 'Camp & Retreat' look like 'Camp---Retreat'? Is it fine?

magdalenesuo: @niranjan94 Fixed. The changes converts "Camp, Trip & Retreat" to "camp-trip-retreat"

niranjan94:
Hmm.. No. Not okay. We need proper slugs. I believe these is custom
postgres function online to create a proper slug. Perhaps., we could use
that? Or we could even go row by row in the migration file and slug it via
a python function already available in one of the helpers.

magdalenesuo:
@niranjan94 What the changes have been so far is stored like this:
screenshot from 2017-06-23 17-35-32
If I am not wrong, this is what we want is it not?
The slugify function in helpers gives **camp---trip-retreat**

niranjan94:
Okay .. what is there now is cool. 😄

magdalenesuo:
@niranjan94 So i do not have to change anything correct?

Issue #3824, opened 2017-06-18
Local static/temp files handling


niranjan94:
Currently, we have two static/temp file locations

1. `app/static`
2. `static`

---

- Remove `app/static` - No use since frontend is disabled and the placeholders too will be moving to the new frontend.
- Delete all folders inside `static`. Just leave a `.gitignore` file which would ignore any files except itself.
- Configure flask to serve the files inside the `static` folder at URL path `/static` **only** if a config variable `SERVE_STATIC` is set to true. Make the config variable settable  via environment. 
- Have a setting option called as `static_domain` which would default to the API domain (if not set). This will be used to generate URLs for the content stored inside the static folder. (Eg. if `static_domain` is set as `https://storage.eventyay.com`, the url of a file inside `static` folder would be `https://storage.eventyay.com/static/file_name.png`. If not set, it would just be `/static/file_name.png` (relative to root of the API server).

PR #3851, opened 2017-06-21
Static links fixed

SaptakS:
@niranjan94 please review this PR a little extra cautiously because I am deleting a lot of stuff and it wasn't possible to check everything since we still don't have that support in api or from frontend. I have also implemented the static_domain logic. Please check into it. I don't know how exactly to check that however. #3824


niranjan94:
@SaptakS I think there has been some small mis-understand here ... We need two things. 
1. `SERVE_STATIC` config option as a part of the Flask config itself that can be set during deployment. 
2. A static domain option that is configurable by the user from admin settings. 

---

- If `SERVE_STATIC` is set to true, then we will serve the static content inside the folder at `/static` along with the rest of the API. If it is set to false, it'll not be served at all. 
- The static domain option configured by the user, will be prepended to all locally stored static file urls. For example, currently if the image is being uploaded to local by the image endpoint, we return url as `/static/something/hello.jpg`. If this static domain is configured, it should return `<static-domain>/static/something/hello.jpg`

app/api/image_upload.py
```diff
@@ -25,4 +26,6 @@ def upload_image():
             image_file,
             UPLOAD_PATHS['temp']['image'].format(uuid=uuid.uuid4())
         )
+    if app.config['STATIC_DOMAIN']:
```
niranjan94: Add this logic in the `upload_local` method itself. This should apply only for local uploads.

config.py
```diff
@@ -50,8 +50,10 @@ class Config(object):
     ERROR_404_HELP = False
     CSRF_ENABLED = True
     SERVER_NAME = env('SERVER_NAME', default=None)
+    STATIC_DOMAIN = env('STATIC_DOMAIN', default=SERVER_NAME)
```
niranjan94: And this should be part of the app settings. Not config. Modifiable at will by the admin.

SaptakS: I will add to settings API then?

niranjan94: Yep :)

app/__init__.py
```diff
@@ -127,6 +127,12 @@ def create_app():
 
     sa.orm.configure_mappers()
 
+    # static subdomain
```
niranjan94: Comment is mis-leading

app/api/helpers/storage.py
```diff
@@ -149,7 +149,10 @@ def upload_local(uploaded_file, key, **kwargs):
     if not os.path.isdir(dir_path):
         os.makedirs(dir_path)
     uploaded_file.save(file_path)
-    return '/serve_' + file_relative_path
+    image_url = '/serve_' + file_relative_path
```
niranjan94: Why the `/serve_` path ? 

Make it such that, 

If the file is inside the static folder, (like `<project root>/static/filename.png`), it should be accessible at `/static/filename.png` 

And if a stattic domain is set, url returned should be `<staticdomain>/filename.png`

SaptakS: Sorry. It is made in that way only. Forgot to remove serve_. Also, should it be `<staticdomain>/filename.png` or `<staticdomain>/static/filename.png`?

niranjan94: Sorry @SaptakS .. didn't notice this comment. Should be `<staticdomain>/filename.png` for static domain case

SaptakS: @niranjan94 resolved. Please check.

PR #3819, opened 2017-06-16
add validation in all existing apis

shubham-padia:
also merge two migration heads

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The unit tests pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

#### Changes proposed in this pull request:

- add validation in all existing apis
- also merge two migration heads
- move save_to_db from api/helpers/helpers.py to api/helpers/db.py and delete api/helpers/helpers.py
- Unprocessable entity: 422 has been used for validation errors

Ref:
- http://marshmallow.readthedocs.io/en/latest/quickstart.html#validation
- http://marshmallow.readthedocs.io/en/latest/extending.html#schema-level-validation
- http://marshmallow.readthedocs.io/en/latest/_modules/marshmallow/validate.html
- http://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.Schema.validate

Fixes #3810

PR #3857, opened 2017-06-22
Add validation in patching method and correct current validation

### LINKED ISSUE #3847: POST and PATCH not working in events, sessions and some other APIs
- [x] bug report

**Current behavior:**

When PATCH an event it shows 
```
{
    "errors": [
        {
            "status": 500,
            "source": "",
            "detail": "'starts_at'",
            "title": "Unknown error"
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}
```

**Expected behavior:**
However it should just rename the 1st event.

**Steps to reproduce:**
1. Create the event
2. PATCH it's name , don't give any of the keywords amongst these `starts-at` , `ends-at` or `timezone`.

**Related code:**
The code responsible for this
```
@validates_schema
    def validate_date(self, data):
        if data['starts_at'] >= data['ends_at']:
            raise UnprocessableEntity({'pointer': '/data/attributes/ends-at'}, "ends-at should be after starts-at")

    @validates_schema
    def validate_timezone(self, data):
        if 'timezone' in data['timezone']:
            offset = timezone(data['timezone']).tz.utcoffset(datetime.strptime('2014-12-01', '%Y-%m-%d')).seconds
```

**Other information:**
Possible solutions :
- Either it should be enclosed in `try/catch` block or
- Use proper way to get an object from dict type , like this .

```
    def validate_date(self, data):
        if data.get('starts_at') and data.get('ends_at'):
            if data['starts_at'] >= data['ends_at']:
                raise UnprocessableEntity({'pointer': '/data/attributes/ends-at'}, "ends-at should be after starts-at")

    @validates_schema
    def validate_timezone(self, data):
        if data.get('starts_at') and data.get('ends_at') and data.get('timezone'):
```
**System information:** 

```
Linux  #57~16.04.1-Ubuntu SMP Wed May 24 16:22:28 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

```
Python 2.7.12
```

LuD1161:
#### Checklist

It fixes error raised due to **keyerror** , when no valid key was found in **data** dictionary.

- Rectified **keyerror** 
- Added `before_update_object` method to each class , which had validation
- changed `is_map_shown` to **Boolean** ( Wasn't done in [PR#3820](https://github.com/fossasia/open-event-orga-server/pull/3820/files) )

Fixes #3847


LuD1161:
@shubham-padia 
Can you now review it ?

shubham-padia:
@LuD1161 You've made no changes, as i said in my review before,

> Don't add validation this way i.e in before get object. Please check #3819 for examples and doc links

LuD1161:
@shubham-padia 
No, it was some connectivity issue and so it changes didn't get pushed.
Have redone it now , please check.

app/api/discount_codes.py
```diff
@@ -23,14 +23,27 @@ class Meta:
...
+        o_data = orig_data['data']['attributes']
+        discount_code = DiscountCode.query.filter_by(id=orig_data['data']['id']).one()
+        if 'min_quantity' in o_data:
+            min_quantity = o_data.get('min_quantity')
+        else:
+            min_quantity = discount_code.min_quantity
+        if 'max_quantity' in o_data:
+            max_quantity = o_data.get('max_quantity')
+        else:
+            max_quantity = discount_code.max_quantity
```
shubham-padia: Only the validation part should be put here, the above part should belong to before_create... or before_update... whichever applies .....

LuD1161: @shubham-padia
isn't it the validation of partial dataset ?

shubham-padia: umm..yeah. Let it be where it is currently. And rename orig_data to something like `original` or `original_data`

LuD1161: @shubham-padia 
Done.

shubham-padia: replace o_data and any other variable to something meaningful too.

app/models/event.py
```diff
@@ -61,7 +61,7 @@ class Event(db.Model):
     large_image_url = db.Column(db.String)
     icon_image_url = db.Column(db.String)
     organizer_name = db.Column(db.String)
-    is_map_shown = db.Column(db.Integer)
```
shubham-padia: Where is the migration for this ?

app/api/discount_codes.py
LuD1161: @shubham-padia 
Done

app/models/event.py
LuD1161: @shubham-padia 
It was already done in [PR#3820](https://github.com/fossasia/open-event-orga-server/pull/3820/files) , just renaming was missed. 
Yes , it was changed from **Integer** to **Boolean** only it was forgotten to modify here.

Check [this migration here](https://github.com/fossasia/open-event-orga-server/pull/3820/files#diff-36a2b03dc3c9199fef623a0a29d63442R41)

image

app/api/discount_codes.py
```diff
@@ -23,14 +23,27 @@ class Meta:
...
 
-    @validates_schema
-    def validate_quantity(self, data):
-        if 'max_quantity' in data and 'min_quantity' in data:
-            if data['max_quantity'] < data['min_quantity']:
+    @validates_schema(pass_original=True)
+    def validate_quantity(self, data, original_data):
+        original_data_attributes = original_data['data']['attributes']
+        discount_code = DiscountCode.query.filter_by(id=original_data['data']['id']).one()
+        if 'min_quantity' in original_data_attributes:
```
poush: Just wondering why you are checking in original_data_attributes. I know it doesn't makes any difference in output/login but it makes sense to check on validated data object i.e, `data` here. 
`data` here is already processed with initial validations. Please correct me If I'm wrong here @LuD1161 @shubham-padia

LuD1161: @poush 
As I mentioned above.
> It is the validation of partial dataset .
It is for the case if the user provided only the `max_quantity` ( or min-quantity ) and if it's less than `min-quantity` ( or max-quantity ) which is already stored in the database.

> it makes sense to check on validated data object i.e, data here.
data here is already processed with initial validations.

What **initial validations** are you talking about on **partial datasets** ( i.e. ones where only either of the parameters is given ) ?

Yes I do agree that **when both the parameters are present then we can use data attribute**.
Like in here 
```
if 'max_quantity' in original_data_attributes and 'min_quantity' in original_data_attributes:
```
and
```
if 'tickets_number' in original_data_attributes and 'min_quantity' in original_data_attributes:
```
But I kept it just for uniformity .

As far as these are concerned 
```
min_quantity = fields.Integer(validate=lambda n: n >= 0)
max_quantity = fields.Integer(validate=lambda n: n >= 0)
```
These will be checked anyhow.
Correct me if I misunderstood your question.

shubham-padia: Agreed  @poush. 
@LuD1161 I think what poush is trying to ask is why not use `data` itself, what is the need for using for using `original_data`. Either the above comment does not answer or either rephrase it.
Can't a simple check like do the work if 
```
if 'start_date' in data:
```
if only start_date is provided and not end_date

shubham-padia: Also if ends_at is not present in patch, instead of this whole long process, just fetch end_date in before_update_object and add it to the data,

poush: Yeah if any one of the two is there in input then add other from database in `before_update_object` and the validation would be simpler.

LuD1161: @shubham-padia @poush 
>Also if ends_at is not present in patch, instead of this whole long process, just fetch end_date in before_update_object and add it to the data

Still that needs to be done in the `before_update_object` ( as it had been done earlier ) and then validation would take place in `before_update_object` instead of in `validate_date`.

It was changed on your suggestion in the first review.

I think it will be better if we have all the validation in the validating function itself.

shubham-padia: > It was changed on your suggestion in the first review.

I just told you to move the existing code, nothing else

> Or do you suggest that only addition of parameters in before_update_object ?

Yes please do that for now

LuD1161: @shubham-padia 
But why is it even needed .
I don't get the whole point of moving code from validation to `before_update_object` .
Why not leave it in validation itself ?

poush: Try something like this,

```python
if 'starts_at' not in data:
data['starts_at'] = object.starts_at
``` 
Similarly for ends_at then just compare data['starts_at'] and data['ends_at']. Simple.
Also I guess validation methods are called first before `before_update_object` so splitting the process will not work.

poush: oops! not starts_at, it's min_quantity and max_quantity

LuD1161: @poush 
Thanks for the cleaner implementation.
Can you review it now ?

LuD1161:
@poush
Sorry , those commits just got mixed up.
I have squashed them and merged them up.
Please check now.

app/api/sessions.py
```diff
@@ -146,15 +154,27 @@ def before_get_object(self, view_kwargs):
             else:
                 view_kwargs['event_id'] = event.id
 
-    decorators = (api.has_permission('is_coorganizer', fetch="event_id",
-                                                       fetch_as="event_id",
-                                                       model=Session,
-                                                       methods="PATCH,DELETE"),)
+    def before_update_object(self, obj, data, view_kwargs):
+        session = Session.query.filter_by(id=obj.id).one()
+        if 'starts_at' in data:
```
poush: You already did this in `validate_date`.

app/api/sessions.py
```diff
@@ -146,15 +154,27 @@ def before_get_object(self, view_kwargs):
             else:
                 view_kwargs['event_id'] = event.id
 
-    decorators = (api.has_permission('is_coorganizer', fetch="event_id",
```
poush: Why removed the permission decorator? and added `jwt_required`.

app/api/tickets.py
```diff
@@ -22,11 +22,23 @@ class Meta:
...
-    @validates_schema
-    def validate_date(self, data):
-        if data['sales_starts_at'] >= data['sales_ends_at']:
-            raise UnprocessableEntity({'pointer': '/data/attributes/sales-ends-at'},
-                                      "sales-ends-at should be after sales-starts-at")
+    @validates_schema(pass_original=True)
+    def validate_date(self, data, original_data):
+        original_data_attributes = original_data['data']['attributes']
+        ticket = Ticket.query.filter_by(id=original_data['data']['id']).one()
+        if 'starts_at' in original_data_attributes and 'ends_at' in original_data_attributes:
```
poush: Here you are using raw data again.

app/api/sessions.py
LuD1161: @poush 
Done

app/api/tickets.py
LuD1161: @poush 
Done

app/api/sessions.py
LuD1161: @poush 
Done

LuD1161:
@niranjan94 @poush @SaptakS 
can you review this ?

app/api/sessions.py
poush: Still there.

LuD1161: @poush
sorry for that but I had done it. ok now it's done , please review it.

Issue #3802, opened 2017-06-14
Ensure basic endpoints are accessible without authentication


niranjan94:
### Event-related endpoints

- [x] Getting a list of events & details about a single event
- [x] Getting a list of sessions of an event and the details about a single session
- [x] Getting a list of microlocations of an event and the details about a single microlocation
- [x] Getting a list of tracks of an event and the details about a single track
- ~Getting a list of speakers of an event and the details about a single speaker~ [ Edited by Piyush: This need to be done separately since speakers API isn't ready yet ]
- [x] Getting a list of tickets of an event and the details about a single tickets
- [x] Getting call for speakers info of an event
- [x] Getting copyright info of an event

### Other endpoints
- [x] Getting settings (all fields except all client IDs, access tokens, secrets, passwords, email settings). But note, `stripe-publishable-key` is required.
- [x] Getting pages list & detail

poush:
I think requires the permission manager to be completed before this.

niranjan94:
@poush updated the description

niranjan94:
@poush what is the progres on this ?

poush:
Almost complete. Just need to check some end points. Also, since sessions API isn't implemented yet so that part is left.

poush:
Cool, so we have now `has-PR` label :D

niranjan94:
@poush looks like this hasn't been done for event relationships ? 

Eg. like `https://api.eventyay.com/v1/events/1/microlocations`

niranjan94:
@poush any updates ?

shubham-padia:
`sponsors` api still remaining as pointed out by @Princu7

Shailesh351:
https://api.eventyay.com/v1/events/69/social-links
https://api.eventyay.com/v1/events/69/session-types
https://api.eventyay.com/v1/events/69/event-type
https://api.eventyay.com/v1/events/69/discount-codes
https://api.eventyay.com/v1/events/69/relationships/speakers-call

https://api.eventyay.com/v1/sessions/2823/session-type

This requests also requires authentication. Fix it also.

niranjan94:
discount codes will need authentication and accessible only to event admins. We don't want someone just enumerating all the discount codes in the system and using them. 

For discount codes, we'll need two new endpoints with only JWT check.
`/v1/events/<id>/discount-codes/<discount-code>`
`/v1/discount-codes/<discount-code>`

For the rest of the endpoints, 
[...]

Issue #3712, opened 2017-06-01
Refactor duplicate endpoints.


SaptakS:
**I'm submitting a ...**  (check one with "x")
- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support requests here, instead ask your query in out Gitter channel at https://gitter.im/fossasia/open-event-orga-server

**Current behavior:**

Right now there are some duplicate endpoints that have arisen. Need to check and refactor them to avoid same endpoints refering to different views.

magdalenesuo:
@SaptakS @shubham-padia Lets do this after most of the basic APIs are done?

SaptakS:
@magdalenesuo I agree with you because there might come up more refactoring issues. So better we do it after all APIs are done in a basic format.

poush:
Along with this, we also need to update the related endpoints. Generally they are using `id` as keys in endpoints while they should be in format `model_id` like `event_id`, `session_id`, etc. This needs to be fixed so that proper utilization of these can be done in different methods.

niranjan94:
@poush this is done or is anything pending here ?

shubham-padia:
This seems to be done

PR #3858, opened 2017-06-22
Implements the permissions for Sessions API

poush:
#### Checklist

removes duplicate endpoints for Sessions Detail API. Using Permission manager, this PR ensures that  Getting a list of sessions of an event and the details about a single session do not require authentication.

Partially #3712, #3802


app/api/sessions.py
```diff
@@ -105,20 +107,14 @@ def query(self, view_kwargs):
...
-            track = self.session.query(Track).filter_by(id=view_kwargs['track_id']).one()
-            data['track_id'] = track.id
-        if view_kwargs.get('session_type_id') is not None:
-            session_type = self.session.query(SessionType).filter_by(
-                id=view_kwargs['session_type_id']).one()
-            data['session_type_id'] = session_type.id
-        if view_kwargs.get('microlocation_id') is not None:
-            microlocation = self.session.query(Microlocation).filter_by(
-                id=view_kwargs['microlocation_id']).one()
-            data['microlocation_id'] = microlocation.id
```
shubham-padia: These should not be removed

app/api/sessions.py
```diff
@@ -105,20 +107,14 @@ def query(self, view_kwargs):
...
+        if view_kwargs.get('event_id'):
+            data['event_id'] = view_kwargs['event_id']
```
shubham-padia: Please fetch the event and assign the event id

poush: Updated. Thanks!

app/api/sessions.py
```diff
@@ -105,20 +109,20 @@ def query(self, view_kwargs):
...
-            track = self.session.query(Track).filter_by(id=view_kwargs['track_id']).one()
-            data['track_id'] = track.id
-        if view_kwargs.get('session_type_id') is not None:
-            session_type = self.session.query(SessionType).filter_by(
-                id=view_kwargs['session_type_id']).one()
-            data['session_type_id'] = session_type.id
-        if view_kwargs.get('microlocation_id') is not None:
-            microlocation = self.session.query(Microlocation).filter_by(
-                id=view_kwargs['microlocation_id']).one()
-            data['microlocation_id'] = microlocation.id
```
shubham-padia: Just curious, the purpose of removing this is to ensure that the object is created only when you POST to `events/{identifier}/sessions`. Am I right or is it something else ?

poush: Yes. The session can be created only for this end point `events/{identifier}/sessions`.

But yeah I need one more suggestion here, since decorators are executed before anything in requests therefore when we are doing permission check we require `event_id` there. Now this makes it not work on `events/{identifier}/sessions` and only works for `events\{event_id}\sessions` but I remember we were using identifier only for public event viewing. So is this good ?

shubham-padia: why not put a check for identifier also here, it should work for both

poush: As discussed simply adding it here will not work since decorators are called before any method. Adding this will add unnecessary database queries and complex logic on permission manager also I'm little confused here about the requirement of this end point for POST and PATCH earlier we were using it only for public event detail. 
@niranjan94 Can you please on confirm on this requirement.

niranjan94: One should be able to use identifier anywhere where event id is used. Public or private.

poush: Done.

poush:
Please review this.

PR #3861, opened 2017-06-23
Implements the permissions for EventCopyRight

### LINKED ISSUE #3773: Fix permissions on All APIs
Pull the lastest changes and update the permissions on each API like  tickets, microlocation, sessions, tax, etc. 

```is_organizer```        => Only Organizer access, Admins

```is_coorganizer```    => Access to all organizers, Admins

```is_registrar```          => Access to Registrars, Organizers, Admins

```is_moderator```          => Access to Moderator, Organizers, Admins

```is_track_organizer```   => Access to Track Organizer, Organizers, Admins

```accessible_events```  => Adds the user_id of logged (non admin) in user to view_kwargs. Since this is used for event listing therefore it was named as accessible events. If you make use of this at other place then make sure you provide it a more generic name as well.

All decorators are second level decorators with ```jwt_required``` as parent decorator.

### PR DESCRIPTION
**Pull request by poush**
#### Checklist

- [x]  Ensure Getting copyright info of an event doesn't require authentication and add Permissions for POST, PATCH, DELETE
- [x]  Add Validation to ensure only one event copyright for one event [ One to One ]
- [x]  Fix Endpoints
- [x]  both `events` and `events_indentifier` should work.

Partially Fixes #3773, #3802, #3712

### CODE REVIEW

File: app/api/event_copyright.py [hunk 1]
```diff
@@ -37,37 +38,50 @@ class Meta:
 
 
 class EventCopyrightList(ResourceList):
-
-    def query(self, view_kwargs):
```
poush: Not required since we do not have GET requests here.

app/api/event_copyright.py
```diff
@@ -37,37 +38,50 @@ class Meta:
...
     def before_create_object(self, data, view_kwargs):
-        if view_kwargs.get('id'):
-            event = self.session.query(Event).filter_by(id=view_kwargs['id']).one()
-            data['event_id'] = event.id
-        elif view_kwargs.get('identifier'):
-            event = self.session.query(Event).filter_by(identifier=view_kwargs['identifier']).one()
+
+        # Permission Manager will ensure that event_id is fetched into
+        # view_kwargs from event_identifier
+        try:
```
poush: Not need to check for `if` because both the endpoints have at least `event` or `event_identifier` and permission manager ensures that `event_id` is there when `event_identifier` is present.

app/api/event_copyright.py
```diff
@@ -37,37 +38,50 @@ class Meta:
...
 
+        try:
+            self.session.query(EventCopyright).filter_by(event_id=data['event_id']).one()
+        except NoResultFound:
+            pass
+        else:
+            raise UnprocessableEntity({'parameter': 'event_identifier'},
+                                      "Event Copyright already exists for the provided Event ID")
+
+    methods = ['POST', ]
```
LuD1161: @poush 
Why not put this method in the decorator itself , as done in **ResourceDetail** ?

poush: Both works in a different way. Methods defined here enables the endpoints for specific methods while permission decorator makes sure it checks for permission only on specified methods.

app/api/event_copyright.py
SaptakS: So basically if this has been done in the decorators, this can be improved in all other APIs as well, right?

poush: Yes, if at that endpoint and for that *methods* you are using permission manager

enigmaeth:
@poush I tried posting event copyright for an event with a user who was not co-organizer.
When making the POST request with event_id, I get this response: `"detail": "Co-organizer access is required."` which is correct, but when using event_identifier, Internal Server Error occurs with this:
```
File "open-event-orga-server/app/api/helpers/permission_manager.py", line 164, in permission_manager
    model = kwargs['model']
KeyError: 'model'
```
Could you please check this once?

app/api/event_copyright.py
```diff
@@ -37,37 +38,50 @@ class Meta:
...
-        elif view_kwargs.get('identifier'):
-            event = self.session.query(Event).filter_by(identifier=view_kwargs['identifier']).one()
+
+        # Permission Manager will ensure that event_id is fetched into
+        # view_kwargs from event_identifier
+        try:
+            event = self.session.query(Event).filter_by(id=view_kwargs['event_id']).one()
+        except NoResultFound:
+            raise ObjectNotFound({'parameter': 'event_identifier'},
+                                 'Event: {} not found'.format(view_kwargs['event_id']))
```
enigmaeth: This looks slightly confusing.
`parameter : event_identifier` while the error message says that event with event_id not found. 
(event_id is fetched into view_kwargs, but what about events that do not exist). Please check.

poush:
@enigmaeth Fir first comment you have to take latest pull from nextgen branch to make it working. This feature was added into last PR for Sessions API

app/api/event_copyright.py
poush: Again, If you pull the latest nextgen branch, you will see it there is any `event_identifier` in url then the permission manager takes cares of it.

Also, for the error. Since it is ensured that `event_id` is there so I directly used it there. Other possible option could be directly saying `Specified Event not found`

poush: Also as per endpoints, there is only one thing {identifier} which can be event ID as well so I guess it says that user provides identifier in the form of identifier key or event ID

PR #3874, opened 2017-06-26
Set max line length to 120 in hound

shubham-padia:
79 is way too strict.


shubham-padia:
@niranjan94 Wow !! The bot is getting more and more cool :robot:

PR #3910, opened 2017-06-29
Added dredd tests for Event Copyright

SaptakS:
Fixes #

#### Checklist

Added dredd tests for event copyright

### DIFF
--- a/app/api/event_copyright.py
+++ b/app/api/event_copyright.py
@@ -15,6 +15,7 @@ from app.api.helpers.exceptions import UnprocessableEntity
 
 
 class EventCopyrightSchema(Schema):
+
     class Meta:
         type_ = 'event-copyright'
         self_view = 'v1.event_copyright_detail'
@@ -38,6 +39,7 @@ class EventCopyrightSchema(Schema):
 
 
 class EventCopyrightList(ResourceList):
+
     def before_create_object(self, data, view_kwargs):
 
         # Permission Manager will ensure that event_id is fetched into
@@ -71,7 +73,7 @@ class EventCopyrightList(ResourceList):
 
 class EventCopyrightDetail(ResourceDetail):
 
-    def before_patch(self, args, kwargs):
+    def before_patch(self, args, kwargs, data):
         if kwargs.get('event_id'):
             try:
                 event_copyright = EventCopyright.query.filter_by(event_id=kwargs['event_id']).one()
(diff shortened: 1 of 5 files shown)


app/api/event_copyright.py
```diff
@@ -71,20 +73,20 @@ def before_create_object(self, data, view_kwargs):
 
 class EventCopyrightDetail(ResourceDetail):
 
-    def before_patch(self, args, kwargs):
+    def before_patch(self, args, kwargs, data):
```
shubham-padia: Why this addition ?

docs/api/api_blueprint.apib
```diff
@@ -1956,63 +1956,14 @@ Data related to event copyright information.
 | `licence-url` | URL of the licence location | string | -  |
 | `logo-url` | URL of the logo for the Licence | string | -  |
 
-## Event Copyright [/v1/events/{event_identifier}/event-copyright{?page%5bsize%5d,page%5bnumber%5d,sort,filter}]
+## Event Copyright [/v1/events/{event_id}/event-copyright{?page%5bsize%5d,page%5bnumber%5d,sort,filter}]
```
shubham-padia: keep this event_identifier only, like the rest of the doc :)

app/api/event_copyright.py
SaptakS: Because without this, it would be (or rather was) giving an error. Because we are sending a data argument in the PATCH, but in this function, we didn't have any  data parameter. So this is necessary I think.

shubham-padia: Hmm.... Might be due to the changes in the new release.

docs/api/api_blueprint.apib
```diff
@@ -1958,61 +1958,12 @@ Data related to event copyright information.
 
 ## Event Copyright [/v1/events/{event_identifier}/event-copyright{?page%5bsize%5d,page%5bnumber%5d,sort,filter}]
 + Parameters
-    + event_identifier: 1 (string) - identifier or event id of the event. (b8324ae2 is an example of identifier)
+    + event_id: 1 (string) - identifier or event id of the event. (b8324ae2 is an example of identifier)
```
shubham-padia: I think you forgot to change it back here.

SaptakS: Yeah sorry. I realised that now. Already pushing.

app/api/event_copyright.py
SaptakS: Might be. Because even the function definition for before_patch makes sense to define data as a parameter. Check [this](https://github.com/miLibris/flask-rest-jsonapi/blob/master/flask_rest_jsonapi/resource.py#L502).

shubham-padia: Yeah, that param was added in the new release only 😉

poush: I think this has to be done in other APIs also where `before_patch` is used. Is it good to make an issue for this?

PR #3946, opened 2017-07-03
add DASHERIZE_API for dasherizing query params

### LINKED ISSUE #3944: Make all query params accept the dasherized version of the attribute names.
Currently the query params, such as 
- [x] sort
- [ ] include
- [ ] filter
- [ ] sparse fieldsets

require the attribute names to be in `snake_case`. But that isn't right since as per our response, the attribute names are dasherized. So, ensure, dasherized attribute names are accepted.

shubham-padia:
fixes #3944 

Ref: https://github.com/shubham-padia/flask-rest-jsonapi/commit/bd97bfbcc3ff355ea9ce8089c64a479b53fa17b5

- adds dasherizing support to include, sparse fieldsets and filters

Set DASHERIZE_API = True in config to enable this,I know it doesn't give much customization e.g option to specify a function to change attributes would be great, but the framework has clearly seperated resources and the api modules which I don't want to disturb in any way. 

Some customization approach can be provided in future if required, but I think this will do the job for now.

### DIFF
--- a/config.py
+++ b/config.py
@@ -56,6 +56,7 @@ class Config(object):
     DATABASE_QUERY_TIMEOUT = 0.1
     SOFT_DELETE = True
     PROPOGATE_ERROR = False
+    DASHERIZE_API = True
 
     if not SQLALCHEMY_DATABASE_URI:
         print('`DATABASE_URL` either not exported or empty')


niranjan94:
@fossasia/open-event-server-devs please review https://github.com/shubham-padia/flask-rest-jsonapi/commit/bd97bfbcc3ff355ea9ce8089c64a479b53fa17b5 too

enigmaeth:
Filtering on dasherized fields works! Checking for include and sparse fieldsets.

PR #3973, opened 2017-07-05
Refactor: add proper exception handling to APIs

### LINKED ISSUE #3972: Add proper exception handling to all apis
For example, https://open-event-api.herokuapp.com/v1/events/3d07e4f5s

The event identifier does not exist. So, I should get a `404 Not Found` error. Instead I get a `500 Internal server error`.  

Also check how relationships behave in this case. (Non-existent event should give a 404 there too)

(Note. this is only in the case of using an identifier. )

shubham-padia:
fixes #3972 
Changes:
- Adds a function to query a record with proper exception handling, ensures that in future no query goes without exception handling, increases readability.
- More than half of the functions were directly filtering like model.id == view_kwargs['something'], refactored them to fetch a record first via safe_query and the filter by that record's id.
- refactored already existing functions with proper error handling
- Remove any bugs **I could find** in the functions being refactored.
- does not refactor speakers_calls and event_copyright, they both shall be covered in the issue for https://github.com/fossasia/open-event-orga-server/pull/3927#pullrequestreview-48246632


shubham-padia:
Please review

shubham-padia:
@niranjan94 @enigmaeth Can you please review again, review dismissed due to fixing some typos, looks good to go now.

app/api/sessions.py
```diff
@@ -107,7 +107,7 @@ def query(self, view_kwargs):
             session_type = safe_query(self, SessionType, 'id', view_kwargs['session_type_id'], 'session_type_id')
             query_ = query_.join(SessionType).filter(SessionType.id == session_type.id)
         if view_kwargs.get('microlocation_id') is not None:
-            microlocation = safe_query(self, Track, 'id', view_kwargs['microlocation_id'], 'microlocation_id')
+            microlocation = safe_query(self, Microlocation, 'id', view_kwargs['microlocation_id'], 'microlocation_id')
```
enigmaeth: We still have many endpoints left to add to documentation and to dredd tests. This would have been caught then!
Rest looks good. @fossasia/open-event-server-devs please review.

shubham-padia: > We still have many endpoints left to add to documentation and to dredd tests

There will be unit tests for this stuff, documentation seems up to date and sufficient right now afaik, Only 2 apis are remaining in dredd testing.

shubham-padia: > microlocations/<int:microlocation_id>/sessions, Shouldn't this be added to the docs

@niranjan94 I'm also not sure, can you confirm ?

enigmaeth: @niranjan94 @SaptakS 
Shouldn't these endpoints too be added to docs and dredd test? 
`/microlocations/<int:microlocation_id>/sessions/` and the likes. In docs for each Resource we have two endpoints [ GET and POST ] for list and `<Resource>/<int:resource_id>` 's three endpoint [GET, PATCH, DELETE] but we have many more endpoints implemented. Also what about the relationship endpoints?

PR #4013, opened 2017-07-11
Updated discount code api for events and tickets

### LINKED ISSUE #3911: Enhance the discount code API
Two different forms of discount code can exist. (Both use the same model). Can be known by using the `used_for` column. Can have two values (`ticket` or `event`)

1. Created by the super admin and can be applied to whole events when creating an event. (this discount is on the total ticket fees that goes to eventyay). 
    1. API should be at `/v1/discount-codes`
    2. Should have a relationship `events` that returns all the events that use this discount code. (can be obtained by using `discount_code_id` field in event model.
2. Created by an event's organizer to be used on that event's tickets
    1. API is already present at `/v1/events/{event_identifier}/discount-codes`
    2. Should have two relationships.
        1. `event` - gives the event that this discount code belongs to (can be obtained by using `event_id` field in discount code model
        2. `orders` - gives all the orders that use this discount code (can be obtained by using `discount_code_id` field in the order table)

magdalenesuo:
Fixes #3911

#### Checklist

Updated discount code api for events and tickets

- Two different forms of discount code 
- Known using used_for column. values - ticket or event
- event (admin)
- ticket (organizers)

#### Detailed description of what this resolves:
1) Admin can post event discounts in `v1/discount-codes`
- GET, POST, PATCH and DELETE, done by admin alone.
2) Event Organizers and admins can post ticket discount in `v1/events/<event_id/discount-codes`.
- GET, PATCH and DELETE can be done by admin, organizers and co-organizers
3) Relationships:
- `v1/discount-codes/<id>/relationships/events` gives relationships of those events that uses the discount code (one discount code -> many events). **NOTE**: if `v1/discount-codes/<id>/event` for Event-discount-code, then we get error. 
- `v1/discount-codes/<id>/relationships/event` gives relationships of the event whose tickets uses the discount code (one event -> many discount codes). **NOTE**: if `v1/discount-codes/<id>/events` for Ticket-discount-code, then we get error. 
- similarly details `v1/discount-codes/<id>/relationships/events` (Event Discount Code) and 'v1/discount-codes/<id>/relationships/event' (Ticket Discount Code)
4) Permission changed to admin, organizer/coorganizer. No random user can GET either. 
5) Made `code`, `value`, `type` and `used_for` REQUIRED.
6) Fixed unrelated issues found along the way.


magdalenesuo:
@niranjan94 Please review. I've added the changes in description.

app/api/discount_codes.py
```diff
@@ -97,40 +170,169 @@ def before_create_object(self, data, view_kwargs):
...
+            event = safe_query(self, Event, 'id', view_kwargs['event_id'], 'event_id')
+            if event.discount_code_id:
+                view_kwargs['id'] = event.discount_code_id
+            else:
+                view_kwargs['id'] = None
+
+        if view_kwargs.get('id'):
+            discount = self.session.query(DiscountCode).filter_by(id=view_kwargs.get('id')).one()
+            if not discount:
+                raise NotFoundError()
```
poush: @SaptakS  @magdalenesuo source, detail are not optional to constructor

app/api/discount_codes.py
```diff
@@ -97,40 +170,169 @@ def before_create_object(self, data, view_kwargs):
...
+            used_for = data['used_for']
+        elif not 'used_for' in data:
+            used_for = discount.used_for
+
+        if discount.used_for == 'ticket' and (current_user.is_coorganizer(view_kwargs.get('event_id'))
+           or current_user.is_organizer(view_kwargs.get('event_id') or current_user.is_admin)) and used_for != 'event':
+            self.schema = DiscountCodeSchemaTicket
+
+        elif discount.used_for == 'event' and current_user.is_admin and used_for != 'ticket':
+            self.schema = DiscountCodeSchemaEvent
```
poush: @SaptakS @magdalenesuo Please ensure that proper schema is there for input request as well because the data will be already loaded as per the default schema in PATCH requests and any change in schema here is for responses only. Validations are already done before this point.

SaptakS: @poush agreed that the validation is done already. But there isn't any change as far as I saw the code in the schema while PATCH request. It is mainly the access who can change what. So basically before data layer seems correct to me. What do you think?

app/api/discount_codes.py
```diff
@@ -70,6 +73,63 @@ def validate_quantity(self, data, original_data):
                          type_='event')
 
 
+class DiscountCodeSchemaEvent(Schema):
```
poush: If both the schemas are almost same then maybe you can try inherit one on other. This may save us from redundancy as well.

app/api/discount_codes.py
poush: Yes, no issues if the Schema attributes and validations are same till now but yes there will be one issue when user will provide `?include=` in URL if the relationships are different.Although I see `relationship_fields = get_relationships(self.resource.schema)` is called after `before_update_object` but schema is already validated before so I guess there will be issue relationship data.

SaptakS: @poush nope. There won't be issue there either. Because it only validates the input data. ?includes is mainly used to see what all data will be shown in response. So it's perfectly fine even in that case. I just checked seeing your comment. 😅

app/api/discount_codes.py
magdalenesuo: @poush I initially tried that but I was not able to change relationships for the other after inherit. And the name `events` <-> `event` are done on purpose. I wouldn't be able to change it if I were to inherit the whole main schema?

poush: Oh okay, That could be an issue. I also checked docs to see if any option to drop attributes or redefine them. There is an option to load only specific attributes but I guess it is from SqlAlchemy Model.

app/api/discount_codes.py
poush: Okay. I'm not completely sure here as well. I think when we provide relationship data (this is done in Orders API ) in body there could be an issue but then I will create an issue for this if anything comes up :)

PR #4024, opened 2017-07-12
fix migration import error of postgresql

shubham-padia:


shubham-padia:
@SaptakS Please review

PR #4040, opened 2017-07-13
Fix bug: sentry cannot catch internal server errors

### LINKED ISSUE #4039: Bug: sentry cannot catch internal server errors
- [x] bug report
- [ ] feature request

**Current behavior:**

sentry cannot catch internal server errors, because api frameworks catches each and every exception

**Expected behavior:**

Based on config variable, framework should not catch exceptions and instead app should handle them and return proper error response

shubham-padia:
fixes #4039 

related commit on framework: https://github.com/shubham-padia/flask-rest-jsonapi/commit/33c728e330ec4936671f0a2a15d488cf8edd1530

Changes:
add optional config `API_PROPOGATE_UNCAUGHT_EXCEPTIONS`  to raise exceptions and made `PROPOGATE_ERROR` in framework

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -17,7 +17,7 @@ import os.path
 from os import environ
 from envparse import env
 import sys
-from flask import Flask
+from flask import Flask, json, make_response
 from app.settings import get_settings, get_setts
 from flask_migrate import Migrate, MigrateCommand
 from flask_script import Manager
@@ -26,6 +26,8 @@ from flask_jwt import JWT
 from datetime import timedelta
 from flask_cors import CORS
 from raven.contrib.flask import Sentry
+from flask_rest_jsonapi.errors import jsonapi_errors
+from flask_rest_jsonapi.exceptions import JsonApiException
 
 import sqlalchemy as sa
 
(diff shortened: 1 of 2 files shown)


app/__init__.py
```diff
@@ -202,5 +204,19 @@ def update_sent_state(sender=None, body=None, **kwargs):
 scheduler.add_job(send_event_fee_notification_followup, 'cron', day=15)
 scheduler.start()
 
+
+@app.errorhandler(500)
+def internal_server_error(error):
+    if 'PROPOGATE_ERROR' in current_app.config:
```
niranjan94: I don't think this check is needed. the config will always have `PROPOGATE_ERROR` right

app/__init__.py
```diff
@@ -202,5 +204,19 @@ def update_sent_state(sender=None, body=None, **kwargs):
 scheduler.add_job(send_event_fee_notification_followup, 'cron', day=15)
 scheduler.start()
 
+
+@app.errorhandler(500)
+def internal_server_error(error):
```
niranjan94: Could you also pass this error to sentry ?

shubham-padia: Sentry auto detects this as no exception is being caught by the app

app/__init__.py
shubham-padia: Yeah, this was the framework only, resolved

app/__init__.py
niranjan94: Cool 👍

config.py
```diff
@@ -54,11 +54,14 @@ class Config(object):
     SQLALCHEMY_DATABASE_URI = env('DATABASE_URL', default=None)
     SERVE_STATIC = env.bool('SERVE_STATIC', default=False)
     DATABASE_QUERY_TIMEOUT = 0.1
+    SENTRY_DSN = env('SENTRY_DSN', default=None)
+
+    # API configs
     SOFT_DELETE = True
     PROPOGATE_ERROR = env.bool('PROPOGATE_ERROR', default=False)
     DASHERIZE_API = True
+    API_PROPOGATE_UNCAUGHT_EXCEPTIONS = env.bool('API_PROPOGATE_UNCAUGHT_EXCEPTIONS', default=True)
```
niranjan94: What does this do ?

shubham-padia: This is added in the framework, in case someone wants the framework to handle all exceptions instead of the app i.e as it was happening previously.

PR #4053, opened 2017-07-14
Migration fix: The slug should be unique along with event_topic_id

### LINKED ISSUE #4052: IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "event_sub_topics_slug_key"
https://sentry.eventyay.com/eventyay/api/issues/16/

```
IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "event_sub_topics_slug_key"
DETAIL:  Key (slug)=(other) already exists.
 [SQL: "INSERT INTO event_sub_topics(name, slug, event_topic_id) SELECT DISTINCT event_sub_topic_id, lower(replace(regexp_replace(event_sub_topic_id, '& |,', '', 'g'), ' ', '-')), event_topic_id               FROM events where not exists (SELECT 1 FROM event_sub_topics where event_sub_topics.name=events.event_sub_topic_id) and event_sub_topic_id is not null"]
(25 additional frame(s) were not displayed)
...
  File "sqlalchemy/engine/base.py", line 1189, in _execute_context
    context)
  File "sqlalchemy/engine/base.py", line 1402, in _handle_dbapi_exception
    exc_info
  File "sqlalchemy/util/compat.py", line 203, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=cause)
  File "sqlalchemy/engine/base.py", line 1182, in _execute_context
    context)
  File "sqlalchemy/engine/default.py", line 470, in do_execute
    cursor.execute(statement, parameters)

IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "event_sub_topics_slug_key"
DETAIL:  Key (slug)=(other) already exists.
 [SQL: "INSERT INTO event_sub_topics(name, slug, event_topic_id) SELECT DISTINCT event_sub_topic_id, lower(replace(regexp_replace(event_sub_topic_id, '& |,', '', 'g'), ' ', '-')), event_topic_id               FROM events where not exists (SELECT 1 FROM event_sub_topics where event_sub_topics.name=events.event_sub_topic_id) and event_sub_topic_id is not null"]
```

> This occurred while running the migrations against a copy of the vintage production database

The slug should be unique along with event_topic_id. (ie) a same slug=event_topic_id pair cannot exist. 

Eg. 

Valid entries:
```
slug = other
event_topic_id = 1

slug = other
event_topic_id = 2
```
Invalid entries.
```
slug = other
event_topic_id = 1

slug = other
event_topic_id = 1
```

shubham-padia:
fixes #4052 

@magdalenesuo @niranjan94 Can you please review this once, I'm assuming this is what was required, but I'm not completely sure :grimacing:


niranjan94:
@shubham-padia actually I was talking about the unique constraint set on `slug`. 

Right now, it is something like 

```sql
CREATE UNIQUE INDEX slug_index ON event_sub_topics (event_sub_topics);
```

So, in this case, what happens is

```
id,slug,event_topic_id
1,other,3
1,other,4 --> Posgres isn't allowing this since `other` already exists
```

---

But what we need is 

```sql
CREATE UNIQUE INDEX slug_index ON event_sub_topics (event_sub_topics, event_topic_id);
```

So in this case,

```
id,slug,event_topic_id
1,other,3
1,other,4 --> Now it will be allowed
1,other,4 --> Will not be allowed
```

I think the term for this is _Composite Unique Index_. 🤔 Not sure of the proper name though.

shubham-padia:
@niranjan94 Done. Please review :)

PR #4087, opened 2017-07-18
Refactor: add direct path to resources with proper relationships

### LINKED ISSUE #4076: [Refactor] Allow direct post for endpoints of resources dependent on another resource
@fossasia/open-event-server-devs
As opposed to what we are doing currently for related objects, most of the json-api clients use the direct post path for creating records (But they pass in the relationships in the payload) 

For e.g. the clients use `GET /v1/events/1/tickets` only for getting the tickets for creating it, they use `POST /v1/tickets` and send the relationship (with event) as a part of the payload (jsonapi spec)
`POST` on `/v1/tickets` should only be possible if the event relationship data is available in the payload, else it should be rejected with a proper error.

The current implementation of post methods on `/v1/events/1/tickets` will be kept unchanged. After some of the refactor under this issue we can decide whether to keep them or not.

shubham-padia:
fixes #4076 

Changes:
- adds  endpoints like `/tickets`
- POST on endpoints like `events/id/tickets` doesn't work now, GET is working fine
- POST and DELETE not possible on required relationship endpoints like `/tickets/<int:id>/relationships/event`, so the user is not able to delete the relationship


app/api/access_codes.py
```diff
@@ -96,6 +97,9 @@ class AccessCodeList(ResourceList):
     """
     List and create AccessCodes
     """
+    def before_post(self, args, kwargs, data):
```
poush: There is one more thing in marshmallow decorators, `post_load`. Can you please check if we can use that instead of calling befoe_post and before_patch?

app/api/event_invoices.py
```diff
@@ -136,7 +139,18 @@ class EventInvoiceDetail(ResourceDetail):
                   'model': EventInvoice}
 
 
-class EventInvoiceRelationship(ResourceRelationship):
+class EventInvoiceRelationshipRequired(ResourceRelationship):
+    """
+    Event Invoice Relationship
+    """
+    decorators = (is_admin,)
```
poush: I think it is good to use permission manager to maintain uniformity and we can remove those duplicate decorators in next PRs.

shubham-padia: Yeah, that shall be done in next PRs

app/api/access_codes.py
shubham-padia: Nope, always try to use flask-rest-jsonapi methods over marshmallow's if available, and I don't know what do you mean by `meaningful` here.

app/api/__init__.py
```diff
@@ -258,19 +265,19 @@
...
 api.route(TaxDetail, 'tax_detail', '/tax/<int:id>', '/events/<int:event_id>/tax')
 api.route(TaxRelationship, 'tax_event', '/tax/<int:id>/relationships/event')
 
 # event invoices
-api.route(EventInvoiceList, 'event_invoice_list', '/events/<int:event_id>/event-invoices',
+api.route(EventInvoiceList, 'event_invoice_list', '/event-invoices', '/events/<int:event_id>/event-invoices',
           '/events/<event_identifier>/event-invoices', '/users/<int:user_id>/event-invoices')
 api.route(EventInvoiceDetail, 'event_invoice_detail', '/event-invoices/<int:id>')
-api.route(EventInvoiceRelationship, 'event_invoice_user',
+api.route(EventInvoiceRelationshipOptional, 'event_invoice_user',
```
shubham-padia: Kept above relationship as optional, please correct if wrong

app/api/__init__.py
```diff
@@ -323,13 +330,13 @@
...
-api.route(AccessCodeList, 'access_code_list', '/events/<int:event_id>/access-codes',
+api.route(AccessCodeList, 'access_code_list', '/access-codes', '/events/<int:event_id>/access-codes',
           '/events/<event_identifier>/access-codes', '/users/<int:user_id>/access-codes',
           '/tickets/<int:ticket_id>/access-codes')
 api.route(AccessCodeDetail, 'access_code_detail', '/access-codes/<int:id>')
-api.route(AccessCodeRelationship, 'access_code_event',
+api.route(AccessCodeRelationshipRequired, 'access_code_event',
           '/access-codes/<int:id>/relationships/event')
-api.route(AccessCodeRelationship, 'access_code_user',
+api.route(AccessCodeRelationshipOptional, 'access_code_user',
```
shubham-padia: Kept above relationship as optional, please correct if wrong

shubham-padia:
@fossasia/open-event-server-devs Please review

app/api/__init__.py
```diff
@@ -82,35 +87,36 @@
...
 # tickets
-api.route(TicketList, 'ticket_list', '/events/<int:event_id>/tickets',
+api.route(TicketList, 'ticket_list', '/tickets', '/events/<int:event_id>/tickets',
           '/events/<event_identifier>/tickets', '/ticket-tags/<int:ticket_tag_id>/tickets',
           '/access-codes/<int:access_code_id>/tickets')
 api.route(TicketDetail, 'ticket_detail', '/tickets/<int:id>', '/attendees/<int:attendee_id>/ticket')
-api.route(TicketRelationship, 'ticket_event', '/tickets/<int:id>/relationships/event')
-api.route(TicketRelationship, 'ticket_ticket_tag', '/tickets/<int:id>/relationships/ticket-tags')
-api.route(TicketRelationship, 'ticket_access_code', '/tickets/<int:id>/relationships/access-codes')
+api.route(TicketRelationshipRequired, 'ticket_event', '/tickets/<int:id>/relationships/event')
```
niranjan94: @shubham-padia what does required/optional mean for these endpoints ?

shubham-padia: Required means the relationship must be present in the request payload for a post request, and the relationship can be updated but it cannot be deleted. Also post on endpoints like `tickets/<int:id>/relationships/event` is disabled because it can be associated with one event. Optional relationship is just the usual relationship we had earlier, you can delete it, post it multiple times and it is not cumpolsary to provide it in the request payload

niranjan94: Okay cool. Thanks for the explanation :)

niranjan94:
I'd like a secondary review on this please. @fossasia/open-event-server-devs please review

app/api/__init__.py
SaptakS: @shubham-padia @niranjan94 I don't know about this implementation because now even POST on `/events/1/tickets` is giving me error 
```
{"errors": [{"status": 422, "source": {"pointer": "/data/relationships/event"}, "detail": "A valid relationship with event resource is required", "title": "Unprocessable Entity"}], "jsonapi": {"version": "1.0"}}
```

Please check and confirm. I don't think if the endpoint is already of the form `/events/<event_id>/tickets`, we need to send a relationship again in the data.

PR #4105, opened 2017-07-20
Add seperate resourcelist class, docs, permissions for direct endpoints of microlocations

### LINKED ISSUE #4102: Add seperate resourcelist class, docs, permissions for direct endpoints of microlocations
parent issue: #4101

shubham-padia:
fixes #4102

### DIFF
--- a/app/api/microlocations.py
+++ b/app/api/microlocations.py
@@ -11,6 +11,8 @@ from app.models.session import Session
 from app.models.event import Event
 from app.api.helpers.db import safe_query
 from app.api.helpers.utilities import require_relationship
+from app.api.helpers.permission_manager import has_access
+from app.api.helpers.exceptions import ForbiddenException
 
 
 class MicrolocationSchema(Schema):
@@ -50,13 +52,24 @@ class MicrolocationSchema(Schema):
                          type_='event')
 
 
-class MicrolocationList(ResourceList):
+class MicrolocationListPost(ResourceList):
     """
-    List and create Microlocations
+    List and create microlocations
     """
     def before_post(self, args, kwargs, data):
         require_relationship(['event'], data)
+        if not has_access('is_coorganizer', event_id=data['event']):
+            raise ForbiddenException({'source': ''}, 'Co-organizer access is required.')
 
+    schema = MicrolocationSchema
+    data_layer = {'session': db.session,
+                  'model': Microlocation}
+
+
+class MicrolocationList(ResourceList):
+    """
+    List Microlocations
+    """
     def query(self, view_kwargs):
         query_ = self.session.query(Microlocation)
         if view_kwargs.get('event_id'):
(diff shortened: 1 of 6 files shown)


app/api/microlocations.py
```diff
@@ -50,13 +51,23 @@ class Meta:
...
 
-class MicrolocationList(ResourceList):
+class MicrolocationDirectList(ResourceList):
     """
-    List and create Microlocations
+    List and create microlocations
     """
     def before_post(self, args, kwargs, data):
         require_relationship(['event'], data)
+        is_coorganizer(data['event'])
```
poush: Why rewrite the same permissions? When you can use 
```
if not has_access('is_coorganizer', event_id=1):
raise Forbidden....
````

File: app/api/microlocations.py [hunk 2]
```diff
@@ -70,24 +81,15 @@ def query(self, view_kwargs):
...
-        event = None
-        if view_kwargs.get('event_id'):
-            event = safe_query(self, Event, 'id', view_kwargs['event_id'], 'event_id')
-        elif view_kwargs.get('event_identifier'):
-            event = safe_query(self, Event, 'identifier', view_kwargs['event_identifier'], 'event_identifier')
-        if event:
-            data['event_id'] = event.id
-
     view_kwargs = True
     decorators = (api.has_permission('is_coorganizer', fetch='event_id', fetch_as="event_id", methods="POST",
```
poush: I don't there is any requirement to set this whole concept. You can use permission manager directly for it. Since you removed POST method from this resource this means only direct endpoint will take care of adding a resource

To achieve the same what you can do is 
`decorators = (api.has_permission('is_coorganizer', fetch='event_id', fetch_as="event_id", methods="GET",....`
`methods = ['GET', 'POST', ]`
```
@jwt_required
          if not has_access('is_coorganizer', event_id=data['event'])
              raise ForbiddenException(''.....")
```

app/api/tracks.py
```diff
@@ -81,22 +92,14 @@ def query(self, view_kwargs):
...
-        if view_kwargs.get('event_id'):
-            event = safe_query(self, Event, 'id', view_kwargs['event_id'], 'event_id')
-            data['event_id'] = event.id
-        elif view_kwargs.get('event_identifier'):
-            event = safe_query(self, Event, 'identifier', view_kwargs['event_identifier'], 'event_identifier')
-            data['event_id'] = event.id
-
     view_kwargs = True
     decorators = (api.has_permission('is_track_organizer', fetch='event_id', fetch_as="event_id", methods="POST"),)
+    methods = ['GET']
```
poush: Setting method to accept GET only but using permission manager for POST.

app/api/microlocations.py
shubham-padia: done

app/api/tracks.py
shubham-padia: done

app/api/microlocations.py
shubham-padia: done

poush:
@shubham-padia I don't think we need separate, "MicrolocationDirectList", before_post of this resource can be added to the "MicrolocationList"

poush:
You can easily check by 
```python
check= lambda a: bool(a) is False 
```
In case permission manager is configured for other methods in any endpoint then
```python
def check(v_k, request=request): 
if 'POST' in request.method:
   if bool(v_k) is False:
      return True
   else:
      return False
return True
```

and send it to check option

shubham-padia:
@poush 'events/id/microlocations' should be get only and so theres no way around seperate classes

poush:
@shubham-padia Actually the snippets I provided above works on the same concept that if there anything in view_kwargs and request is POST one then don't pass the request. Since decorators run before anything (practically they call the resource view on valid permission) so there will not be any issue there as well.
But yes I think we can change the error from forbidden to 404 or something else.

shubham-padia:
I don't want that functionality to rely on permission manager :smile:

shubham-padia:
@poush please review everything else, considering everything the current approach seems good to me :smile:

niranjan94:
@shubham-padia `GET /v1/events/<identifier>/microlocations` should also be documented and under dredd tests

poush:
@shubham-padia Okay, in that case just one concern, is it possible to call it something like "MicrolocationPost" or "MicrolocationListPost" or something like that?

shubham-padia:
@niranjan94 shouldn't that be converted in https://github.com/fossasia/open-event-orga-server/issues/4054 as there are many such endpoints :smile:

@poush Sure, ill name to `MicrolocationListPost`, sounds good?

niranjan94:
That issue is about sessions isn't it?

shubham-padia:
Sorry linked the wrong issue, https://github.com/fossasia/open-event-orga-server/issues/4009

niranjan94:
@shubham-padia Mmm okay .. Cool

poush:
@shubham-padia Sounds Good :)

shubham-padia:
@poush Done

PR #4112, opened 2017-07-20
Sessions: Add seperate resourcelist class, docs, permissions for dire…

### LINKED ISSUE #4110: Sessions: Add seperate resourcelist class, docs, permissions for direct endpoints
parent issue: #4101

shubham-padia:
fixes #4110

### DIFF
--- a/app/api/helpers/permission_manager.py
+++ b/app/api/helpers/permission_manager.py
@@ -1,10 +1,11 @@
 from flask import current_app as app
 from flask_jwt import _jwt_required, current_identity
-from app.api.helpers.errors import ForbiddenError, NotFoundError
-from app.api.helpers.permissions import jwt_required
 from sqlalchemy.orm.exc import NoResultFound
 from flask import request
 
+from app.api.helpers.errors import ForbiddenError, NotFoundError
+from app.api.helpers.permissions import jwt_required
+from app.models.session import Session
 from app.models.event import Event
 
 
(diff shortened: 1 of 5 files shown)


app/api/sessions.py
```diff
@@ -104,12 +106,24 @@ def validate_date(self, data, original_data):
...
 
-class SessionList(ResourceList):
+class SessionListPost(ResourceList):
     """
-    List and create Sessions
+    List Sessions
     """
     def before_post(self, args, kwargs, data):
         require_relationship(['event'], data)
+        if not has_access('is_coorganizer', event_id=data['event']):
```
niranjan94: An unauthenticated user can:
- View all accepted/approved sessions (even via relationships)

An authenticated non-event org user can:
- Create a session for any live event
- Session can have a state of either draft or pending. (Default pending)
- View/edit his/her own sessions (Cannot add/modify track/microlocation relation)
- Delete his/her own sessions

An event org can:
- Create a session with any state for that event
- List all sessions of that event
- Modify all sessions of that event and all relationships
- Delete all sessions

> event org referes to event organizer/co-organizer

shubham-padia: > Session can have a state of either draft or pending. (Default pending)

Shouldn't default be draft, so that the session is not submitted by itself ?

niranjan94: Yep. Draft.

shubham-padia: > Session can have a state of either draft or pending
Cannot add/modify track/microlocation relation

Should a 403 be raised in this case or a 422, or should it be ignored i.e relationships will be ignored and state will be changed to draft

poush: What about 400 Bad Request?

shubham-padia: @poush Bad request applies for a malformed syntax e.g a missing comma in the JSON or syntactical errors in the request 
@niranjan94 can you have a look at the above comment again please ?

SaptakS: @shubham-padia just a note. 400 isn't only for a malformed syntax. Any client side mistake in the data that the server cannot process. It can be a syntax error, invalid request or even a deceptive request according to RFC 7231. So you can use 422 Unprocessable entity also, but 400 Bad Request also covers unprocessable stuff.  It's more like 422 is used when the request is valid and processable but you can't or don't want to process the particular instruction. 400 you use when it is either syntactically wrong, or syntactically correct but invalid for your system. Though according to RFC 2616, whatever you said was true.

Just wanted to notify.

niranjan94: This is clearly a `403` error. The data is valid. The syntax is valid. But that specific user does not have _permission_ do it.

shubham-padia:
@niranjan94 Changes made as per the required permissions, please review :)

shubham-padia:
@niranjan94 please review :)

app/api/helpers/permission_manager.py
```diff
@@ -94,6 +95,35 @@ def is_coorganizer_or_user_itself(view, view_args, view_kwargs, *args, **kwargs)
...
+    if user.is_admin or user.is_super_admin:
+        return view(*view_args, **view_kwargs)
+
+    if user.is_staff:
+        return view(*view_args, **view_kwargs)
+
+    try:
+        session = Session.query.filter(Session.id == view_kwargs['id']).one()
+    except NoResultFound:
+        return NotFoundError({'parameter': 'event_identifier'}, 'Event not found.').respond()
```
enigmaeth: It should be `{'parameter': 'session_id'}, 'Session not found.'` :sweat_smile:

shubham-padia: Thanks, resolved :sweat_smile:

PR #4115, opened 2017-07-20
Tax: Add seperate resourcelist class, docs, permissions for direct en…

### LINKED ISSUE #4113: Tax: Add seperate resourcelist class, docs, permissions for direct endpoints
parent issue #4101

shubham-padia:
fixes #4113


app/api/tax.py
```diff
@@ -57,24 +72,13 @@ def query(self, view_kwargs):
...
-        event = None
-        if view_kwargs.get('event_id'):
-            event = safe_query(self, Event, 'id', view_kwargs['event_id'], 'event_id')
-        elif view_kwargs.get('event_identifier'):
-            event = safe_query(self, Event, 'identifier', view_kwargs['event_identifier'], 'event_identifier')
-        if event:
-            data['event_id'] = event.id
-
     view_kwargs = True
-    decorators = (jwt_required, )
```
SaptakS: Why remove decorators?

SaptakS: In fact I think event tax information should be visible only to organisers. @niranjan94 ?

niranjan94: Yes. Only to organizers.

shubham-padia: Why only to organisers, shouldn't take information be public, i should know the tax before paying for a ticket?

shubham-padia: *tax

niranjan94: Hmmm .. You have a point ... Then ensure only `is-tax-included` and `rate` fields are accessible to public.

shubham-padia: Why not the other fields, those are also public generally afaik ?

SaptakS: @shubham-padia we don't want to share any information regarding the organizer or whoever's billing address is being provided. That may or may not be publicly available. So no personal information like that including the company which has registered should be shown according to me.

shubham-padia: So the customer is not supposed to know the company that is billing him :smile: :smile: ?

SaptakS: Say for example you pay for an event XYZ, say it is said organized by Group ABC which is actually run by a company MNO. You will never see the name for MNO, their address or any such detail. You will know you have paid someone in Group ABC and get an email ID to contact in case of refund and all. You never know the underlying registered company. 
AFAIK. And as far as I have seen. @niranjan94 might be able to explain better.

shubham-padia: @niranjan94 your opinions?

niranjan94: @shubham-padia they need not know the company's tax ID, registered address, etc etc. They'll anyway get who is billing them when they make the payment at paypal or stripe. 

So, just tax rate and whether it is included or not is enough. The rest are all internal details that we add to the invoice we send to the event organizer at the end of each month.

shubham-padia: Okay...cool

shubham-padia:
@niranjan94 done. Please review

poush:
@shubham-padia Please check. Travis is failing.

shubham-padia:
@poush Done. Thanks, was an error in resolving conflicts :sweat_smile:

shubham-padia:
@poush, resolved, please review :)

shubham-padia:
@poush, resolved travis, please review :)

app/api/tax.py
```diff
@@ -57,64 +108,37 @@ def query(self, view_kwargs):
...
-                event = self.session.query(Event).filter_by(id=view_kwargs['event_id']).one()
-            except NoResultFound:
-                raise ObjectNotFound({'parameter': 'event_id'},
-                                     "Event: {} not found".format(view_kwargs['event_id']))
-            else:
-                if event.tax:
-                    view_kwargs['id'] = event.tax.id
-                else:
-                    view_kwargs['id'] = None
+    def before_get(self, args, kwargs):
```
poush: There is one more way here, You can instead put decorator like this

```python
@jwt_required
def before_get(....
```

poush: and then keep the same permission i.e, "is_coorganizer".

poush: Ignore, Not relevant here 😅

app/api/tax.py
```diff
@@ -43,9 +65,38 @@ class Meta:
...
+        if not has_access('is_coorganizer', event_id=data['event']):
+            raise ForbiddenException({'source': ''}, 'Co-organizer access is required.')
+
+    def before_get(self, args, kwargs):
+        if has_access('is_admin'):
+            self.schema = TaxSchema
+        else:
+            self.schema = TaxSchemaPublic
+
+    schema = TaxSchema
```
poush: Add jet_required here. Since noth endpoints require logged in user

app/api/tax.py
```diff
@@ -43,9 +65,38 @@ class Meta:
...
+    data_layer = {'session': db.session,
+                  'model': Tax}
+
+
+class TaxList(ResourceList):
+    def before_get(self, args, view_kwargs):
+        if view_kwargs.get('event_identifier'):
+            try:
```
poush: I think **save_query** can be used here.

shubham-padia: Already tried that, it requires self to have a session :)

app/api/tax.py
shubham-padia: Public schema doesnt require logged in user, while coorganizer access is already being checked in before post. While only the admin can see all the tax with private details in a list.

poush: But there is one issue. The current_identitiy will be null until jwt_required is called. So to make it work I think you have to provide `@jwt_required ` at the top of before_post method.

app/api/tax.py
```diff
@@ -43,9 +65,38 @@ class Meta:
...
 
-class TaxList(ResourceList):
+class TaxListPost(ResourceList):
     def before_post(self, args, kwargs, data):
         require_relationship(['event'], data)
+        if not has_access('is_coorganizer', event_id=data['event']):
+            raise ForbiddenException({'source': ''}, 'Co-organizer access is required.')
+
+    def before_get(self, args, kwargs):
+        if has_access('is_admin'):
```
poush: So if this TaxSchemaPublic list is visible by public then it should be "is_admin_with_jwt" also because it will throw 403 for any guest

app/api/tax.py
poush: So if session is not booting up here, Can before_get_collection method be used here? (just curious)

shubham-padia: Then changing the schema wont be possible afaik

app/api/tax.py
shubham-padia: If you look at the is_admin function, it already has a jwt_required decorator, so 401 is thrown for any unautorized guest if he/she tries to post, if he is logged in and not a co-organiser, then a 403 will be thrown

poush: Actually I was saying that current "is_admin" will throw error if the user is not logged in but I think here guests should be able to view the public schema but they would not be able to since "is_admin" has already raised the permission error. I guess this is the reason behind "is_coorganizer_Wthout_jwt.." also.

shubham-padia: @poush Okay, got it. Resolved.

app/api/tax.py
```diff
@@ -43,9 +66,41 @@ class Meta:
...
-class TaxList(ResourceList):
+class TaxListPost(ResourceList):
     def before_post(self, args, kwargs, data):
         require_relationship(['event'], data)
+        if not has_access('is_coorganizer', event_id=data['event']):
+            raise ForbiddenException({'source': ''}, 'Co-organizer access is required.')
+
+    def before_get(self, args, kwargs):
+        if 'Authorization' in request.headers:
+            if has_access('is_admin'):
```
SaptakS: I think this can be changed to be written to write like this:

```python
if 'Authorization' in request.headers and has_access('is_admin'):
    self.schema = TaxSchema
else:
    self.schema = TaxSchemaPublic
```

shubham-padia: @SaptakS Resolved

shubham-padia:
@SaptakS please review .

shubham-padia:
@SaptakS please review .

shubham-padia:
@SaptakS please review .

PR #4153, opened 2017-07-24
Add /healthcheck endpoint

### LINKED ISSUE #4151: Implement a health check endpoint.
Implement a health check endpoint similar to http://api.susi.ai/susi/status.json at `/status` which would check the following things. 

- Connection to the database. (And a few quick select queries using all the sqlalchemy models to ensure the models are inline with the migrations)
- Connection to redis
- Connection to the internet

The endpoint should return a `200` if all checks are okay. Else a `503 Service Unavailable` error to be returned along with information on which check failed. 

**Why is this required ?** 
- This will be used by kuberentes to know when the web instance is ready to receive requests

shubham-padia:
Fixes #4151 

**Changes:**
- https://github.com/Runscope/healthcheck
- Implements a healthcheck endpoint at `/healthcheck` (`/healthcheck` sounded more logical than `/status`, do tell if there is a requirement for the endpoint to be named `/status` )
Example:
```json
{
    "status": "success",
    "timestamp": 1500915121.52474,
    "hostname": "shubham",
    "results": [
        {
            "output": "celery ok",
            "checker": "health_check_celery",
            "expires": 1500915148.50881,
            "passed": true,
            "timestamp": 1500915121.50881
        },
        {
            "output": "database ok",
            "checker": "health_check_db",
            "expires": 1500915148.524729,
            "passed": true,
            "timestamp": 1500915121.524729
        }
    ]
}
```
- Implements an environment endpoint at `/environment`, secret keys, passwords etc are masked off:
https://pastebin.com/urvV7mh5


app/__init__.py
```diff
@@ -25,6 +25,7 @@
 from raven.contrib.flask import Sentry
 from flask_rest_jsonapi.errors import jsonapi_errors
 from flask_rest_jsonapi.exceptions import JsonApiException
+from healthcheck import HealthCheck, EnvironmentDump
```
shubham-padia: To be ignored, same pattern as the rest of the file

app/__init__.py
```diff
@@ -38,7 +39,7 @@
 from app.api.helpers.auth import AuthManager
 from app.models.event import Event, EventsUsers
 from app.models.role_invite import RoleInvite
-
+from app.views.healthcheck import health_check_celery, health_check_db, get_safe_config
```
shubham-padia: To be ignored, same pattern as the rest of the file

app/__init__.py
```diff
@@ -160,6 +161,13 @@ def __call__(self, *args, **kwargs):
 
 celery = make_celery(current_app)
 
+# Health-check
+health = HealthCheck(current_app, "/healthcheck")
+envdump = EnvironmentDump(current_app, "/environment", include_config=False)
```
niranjan94: No need environment. Too much data.

app/__init__.py
```diff
@@ -160,6 +161,13 @@ def __call__(self, *args, **kwargs):
 
 celery = make_celery(current_app)
 
+# Health-check
+health = HealthCheck(current_app, "/healthcheck")
```
niranjan94: Lets call it `/heath-check`

niranjan94:
In fact, @shubham-padia can we have these endpoints available on a different port altogether ?

shubham-padia:
> can we have these endpoints available on a different port altogether ?

We can, but what is the reason to do so :) ?

niranjan94:
@shubham-padia we don't want these endpoints accessible to the public. These are completely for internal use and we can have the `/environment` endpoint too.

shubham-padia:
@niranjan94 I don't think we can run a single flask app on multiple ports :grimacing: , should I remove the /environment then ?

shubham-padia:
@niranjan94 Have added a check to check whether migrations are up to date, and removed `/environment`  endpoint

app/views/healthcheck.py
```diff
@@ -0,0 +1,47 @@
...
+    :return:
+    """
+    try:
+        db.session.execute('SELECT 1')
+        return True, 'database ok'
+    except:
+        return False, 'Error connecting to database'
+
+
+def health_check_migrations():
```
niranjan94: This isn't of any use here. If there were multiple heads, the server wouldn't start and kubernetes would roll back to the previous version that worked.

What we need is the check that ensures the models are inline with the migration. A simple select query from each model would tell us that.

shubham-padia: Hi @niranjan94 this is not checking for multiple heads, the table alembic_version contains the latest revision to which the database was actually upgraded, while python manage.py db heads will give us the head according to migration files. Comparing both of them will tell whether the db has upgraded or not. Or I can replace this with a select query from each model if you want :)

shubham-padia: Although come to think of it, this may not work if migration file is missing, select queries would help in that situation, i'll replace with that :)

shubham-padia: @niranjan94 Resolved :point_up: .
Sample error message:
```json
{
            "output": "database up to date with migrations",
            "checker": "health_check_migrations",
            "expires": 1500965572.170381,
            "passed": true,
            "timestamp": 1500965545.170381
}
```

app/views/healthcheck.py
```diff
@@ -0,0 +1,60 @@
...
+            classes.append(class_)
+        except:
+            pass
+    for table in db.metadata.tables.items():
+        if table[0] in table_names:
+            models.append(classes[table_names.index(table[0])])
+
+    for model in models:
+        try:
+            db.session.query(model).all()
```
niranjan94: Don't query all. Will take up a lot of time if there are lot of rows. Query one record. (`LIMIT 1`)

niranjan94:
Also, @shubham-padia what would be the `HTTP` error code if any of the tests fail ?

shubham-padia:
>  what would be the HTTP error code if any of the tests fail ?

500 Internal Server Error

niranjan94:
Cool

app/views/healthcheck.py
shubham-padia: Resolved :)

app/views/healthcheck.py
```diff
@@ -0,0 +1,60 @@
...
+            classes.append(class_)
+        except:
+            pass
+    for table in db.metadata.tables.items():
+        if table[0] in table_names:
+            models.append(classes[table_names.index(table[0])])
+
+    for model in models:
+        try:
+            db.session.query(model).first()
```
niranjan94: What is the query generated for this ? I think there would be a order by being done.

niranjan94: If there is, then ensure there is no sorting done.

shubham-padia: From the docs at http://docs.sqlalchemy.org/en/rel_0_9/orm/query.html#sqlalchemy.orm.query.Query.first
> Return the first result of this Query or None if the result doesn’t contain any row.
first() applies a limit of one within the generated SQL, so that only one primary entity row is generated on the server side (note this may consist of multiple result rows if join-loaded collections are present).

niranjan94: Could you confirm it by checking the raw query ? Because I remember seeing it once somewhere.

shubham-padia: @niranjan94 below is the query for activities model:
```sql
[SQL: 'SELECT activities.id AS activities_id, activities.actor AS activities_actor, activities.time AS activities_time, activities.action AS activities_action \nFROM activities \n LIMIT %(param_1)s'] [parameters: {'param_1': 1}]
```

PR #4168, opened 2017-07-27
Health-check: Check if migrations are up to date only once

### LINKED ISSUE #4165: Health-check: Check if migrations are up to date only once
Currently checking if migrations are up to date or not causes a timeout when checked again everytime `/health-check` is called. We can run this on startup in create_app, store a string containing the results in maybe say app.config and whenever a call to /health-check is made, we can parse that saved string.

shubham-padia:
fixes #4165 
Changes:
- Health-check: Check if migrations are up to date only once
- add proper sentry exception catching in health-check
- add unittest to check if migrations are missing


shubham-padia:
@niranjan94 please review

Issue #4196, opened 2017-07-30
Implement Email, Notifications and Elasticsearch


niranjan94:
- Emails & notifications (High priority)
	- [x] Session submission - to organizer
	- [x] Session approval/rejection - to speakers
	- [x] Session acceptance/rejectance - to organizer
 	- [x] Event invoice - monthly - to organizer
    - [x] Event invoice - Follow up email if not paid -  to organizer
 	- [x] Ticket purchase
	    - Individual ticket pdf - to all attendees (except purchaser)
	    - Order summary and a single pdf with all tickets - to the ticket purchaser
	- [x] Ticket cancellation - To the ticket purchaser
	- [x] Email verification email - To the user who registered
	- [x] A welcome email - After registration. Before email verification. - To the user who registered
	- [x] Forgot password reset email - To the user who registered
   	- [x] Email change notification (#4210) - To both the old email address and the new email address
   	- [x] Password change email - To the user
   	- [x] Event Role creation - To the invited user's email
	- [x] Event import success (with link to imported event) /failure - To the user (To be done after #3737)
	- [x] Event export success (with the exported file attached) /failure - To the user (To be done after #3731 )
	- [ ] Speakers call invite email - To the user who is being invited
	- [x] After event thank you email - To all registrered attendees & speakers
- Elasticsearch base for event searching (Low priority. Could be moved to next milestone depending on how the email notifications go. Not to be taken up until we have a PR for all of the above.) (#3467)

@fossasia/open-event-devs

poush:
I'm thinking of one scenario here, i.e., if email changes, we have to check what is changed in after_create_object to see if email is changed and send them email only if email is changed. But what if we put this responsibility in email setter of User model. I guess it is not the responsibility of API to send the email.

SaptakS:
@niranjan94 the remaining tasks are being followed in #3467  and #4225 . Closing this.

PR #4211, opened 2017-07-31
Send an email when user email is changed

### LINKED ISSUE #4210: Send email on changing/updating email of a user
Parent #4196

poush:
added send_Email_With_Action method

Fixes #4210

#### Checklist

Parent Issue #4196 

This PR works by listening event on user model instead of relying on API Resources.


app/models/user.py
```diff
@@ -352,3 +355,19 @@ def __setattr__(self, name, value):
 @event.listens_for(User, 'init')
 def receive_init(target, args, kwargs):
     target.signup_at = datetime.now(pytz.utc)
+
+
+@event.listens_for(User, 'after_update')
```
poush: @niranjan94 This way we don't need to rely on API resource to send the email. If the email is updated user model will send the email.

poush: get_history retutn this 

`History(added=[u'me@ipiyush.comss'], unchanged=(), deleted=[u'me@ipiyush.coms'])`

poush:
@fossasia/open-event-server-devs review please.

app/api/helpers/mail.py
```diff
@@ -73,6 +73,22 @@ def send_email(to, action, subject, html):
...
+    A general email helper to use in the APIs
+    :param user: email or user to which email is to be sent
+    :param action:
+    :param kwargs:
+    :return:
+    """
+    send_email(
+        to=user,
+        action=action,
+        subject=MAILS[action]['subject'],
```
SaptakS: Subject may also need kwargs

app/models/user.py
SaptakS: @poush any particular reason why you think email service should be implemented at the model level instead of the API resource level. In fact I think we should keep models as light as possible and not implement services over there. @niranjan94 you views?

poush: @SaptakS I think we mailing and notifications can work better as events occuring and listening in our app. This way we have separate module/class/file handeling different pre & post actions. This will also ensure that if any such action is done to models directly to helpers then the required pre/post process is also executed.
Also I think this binds an SQLAlchemy event to User model so may be we can create a separate helper handling different events and put this logic there?

SaptakS: @poush I still find it a bad idea to do this task in model level rather than an API level. Model is supposed to do transactions and getter and setter tasks mostly. the event listener you are using is mainly used when you need to rollback a commit based on certain factors in the model change which is again a DB operation. So I find it a little bad idea. Had it been some logging function we were implementing I would have said kudos to your implementation. I will wait for @fossasia/open-event-server-devs  to decide.

app/api/helpers/mail.py
```diff
@@ -73,6 +74,22 @@ def send_email(to, action, subject, html):
...
+def send_email_with_action(user, action, **kwargs):
+    """
+    A general email helper to use in the APIs
+    :param user: email or user to which email is to be sent
+    :param action:
+    :param kwargs:
+    :return:
+    """
+    send_email(
+        to=user,
```
SaptakS: it always needs to be email, right? What do you mean by "or user to which email is to be sent"? I think naming email is better.

app/api/users.py
```diff
@@ -213,6 +212,13 @@ def before_update_object(self, user, data, view_kwargs):
             if data.get('icon_image_url'):
                 del data['icon_image_url']
 
+        if data.get('email') and data['email'] != user.email:
+            view_kwargs['email_changed'] = user.email
+
+    def after_update_object(self, user, data, view_kwargs):
+        if view_kwargs.get('email_changed'):
```
SaptakS: why not do the above comparison in this function itself?

poush: Because data contains the new email and user object is updated here.

app/api/helpers/mail.py
poush: Missed it in on refactoring, Actually, I added a support to directly send the user object as well.

poush:
@niranjan94 Sample link is 
```
http://eventyay.com/email/verify?token=WyJtZUBpcGl5dXNoLmNvbXNzc3MiLCAiWkdVMjk3Il0uTzN2dDlENG1ZWGEzZFkwZHZkalJwV0JDZWFN
```
Token => base64 encoded => url encoded

app/api/users.py
SaptakS: Oh yeah. My mistake.

app/api/helpers/mail.py
```diff
@@ -154,3 +177,11 @@ def send_email_after_event(email, event_name, upcoming_events):
             upcoming_events=upcoming_events
         )
     )
+
+
+def send_email_change_user_email(new_email, email):
+    s = get_serializer()
+    hash = urllib.quote(base64.b64encode(s.dumps([email, str_generator()])))
+    link = make_fe_url(path='/email/verify?token={token}'.format(token=hash))
```
niranjan94: Could we make `make_fe_url` accept a dict of query params instead of like this ?

So usage would be like:

```python
link = make_fe_url('/url/without-query')
```

```python
link = make_fe_url('/url/with-query', {
  "token": "xyzzy"
})
```

niranjan94: And the `make_fe_url` method would use urllib to properly generate a valid URL

SaptakS: Also please name it make_frontend_url. 😅

app/api/helpers/mail.py
```diff
@@ -50,7 +53,7 @@ def send_email(to, action, subject, html):
                 }
 
                 from tasks import send_mail_via_smtp_task
-                send_mail_via_smtp_task.delay(config, payload)
+                SEND_MAIL_VIA_SMTP_TASK.delay(config, payload)
```
SaptakS: @poush please check this.

app/api/auth.py
```diff
@@ -0,0 +1,34 @@
...
+
+from app.api.helpers.utilities import get_serializer
+from app.models.user import User
+
+auth_routes = Blueprint('auth', __name__, url_prefix='/v1/auth')
+
+
+@auth_routes.route('/verify-email', methods=['POST'])
+def verify_email():
+    token = request.json['data']['token']
```
niranjan94: You would need to do a base64 decode

app/api/helpers/files.py
```diff
@@ -173,11 +163,11 @@ def create_save_image_sizes(image_file, image_sizes_type, unique_identifier=None
     return new_images
 
 
-def make_fe_url(settings=None, path=None):
+def make_frontend_url(path=None, parameters=None):
```
niranjan94: Make path a required parameter.

app/api/helpers/files.py
```diff
@@ -173,11 +163,11 @@ def create_save_image_sizes(image_file, image_sizes_type, unique_identifier=None
...
     """
     Create URL for frontend
     """
-    if settings is None:
```
niranjan94: Build the url properly. Such that even if there are things like extra slashes or missing scheme in the frontend_url settings, the final url is proper. Use something like 

```python
frontend_url = urlparse.urlparse(settings['frontend_url'] if settings['frontend_url'] else '')
return urlparse.urlunparse((
		frontend_url.scheme,
		frontend_url.netloc,
		path,
		'',
		urllib.urlencode(parameters) if parameters else '', 
		''
	));
```

app/api/auth.py
```diff
@@ -0,0 +1,36 @@
...
+@auth_routes.route('/verify-email', methods=['POST'])
+def verify_email():
+    token = base64.b64decode(request.json['data']['token'])
+    s = get_serializer()
+
+    try:
+        data = s.loads(token)
+    except Exception:
+        return abort(
+            make_response(jsonify(error="Invalid Token"), 422)
```
niranjan94: Should be a `400 Bad request`. Token is present but invalid.

app/api/auth.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+
+    try:
+        user = User.query.filter_by(email=data[0]).one()
+    except Exception:
+        return abort(
+            make_response(jsonify(error="Invalid Token"), 422)
```
niranjan94: Should be a `400 Bad request`. Token is present but invalid.

app/api/auth.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+    else:
+        user.is_verified = True
+        save_to_db(user)
+
+    return jsonify({"user": user.id})
```
niranjan94: Return the following. 

- `id`
- `email`
- `name` (If no name present then return a `null`)

So the client can personalise the message shown to the user.

poush: Done

app/api/auth.py
poush: Done

app/api/auth.py
poush: Done

app/api/helpers/ticketing.py
```diff
@@ -89,7 +89,7 @@ def charge_stripe_order_payment(order, token_id):
             save_to_db(order)
 
             invoice_id = order.get_invoice_number()
-            order_url = make_fe_url(path="/{identifier}/view/".format(identifier=order.identifier))
+            order_url = make_frontend_url(path="/{identifier}/view/".format(identifier=order.identifier))
```
SaptakS: Don't think this i needed since you have commented out the lines requiring this. please verify.

PR #4238, opened 2017-08-04
Add general event statistics

### LINKED ISSUE #4007: Implement general event analytics API
Implement event analytics API for the event organizer dashboard, 
General bulk data about event should be provided by event id/identifier

@niranjan94 Please suggest Url and required fields for this one issue only, any other issues may follow the same pattern

shubham-padia:
fixes #4007 
Add general event analytics at `analytics/events/{event_identifier}/general`.
There will be another endpoint called `analytics/events/{event_identifier}/sales` in future, that's why the extra general in the current endpoint. The sales data is irrelevant for the event dashboards and they generally have a separate section.

The fields were decided on the basis of the following views,
selection_223
selection_222

Co-organizer level access is required, no matter if the event is published, public access to analytics should not be there.

### DIFF
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -48,6 +48,7 @@ from app.api.modules import ModuleDetail
 from app.api.custom_placeholders import CustomPlaceholderList, CustomPlaceholderDetail, CustomPlaceholderRelationship
 from app.api.activities import ActivityList, ActivityDetail
 from app.api.orders import OrdersList, OrderDetail, OrderRelationship, ChargeList, OrdersListPost
+from app.api.event_statistics import EventStatisticsGeneralDetail
 
 # users
 api.route(UserList, 'user_list', '/users')
@@ -385,11 +386,13 @@ api.route(StripeAuthorizationDetail, 'stripe_authorization_detail',  '/stripe-au
 api.route(StripeAuthorizationRelationship, 'stripe_authorization_event',
           '/stripe-authorization/<int:id>/relationships/event')
 
+
 # Orders API
 api.route(OrdersListPost, 'order_list_post', '/orders')
 api.route(OrdersList, 'orders_list', '/orders', '/events/<int:event_id>/orders',
           '/events/<event_identifier>/orders')
 api.route(OrderDetail, 'order_detail', '/orders/<identifier>')
+
 # Charges API
 api.route(ChargeList, 'charge_list', '/orders/<identifier>/charge', '/orders/<identifier>/charge')
 api.route(OrderRelationship, 'order_attendee', '/orders/<identifier>/relationships/attendee')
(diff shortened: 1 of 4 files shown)


app/api/__init__.py
```diff
@@ -398,3 +401,7 @@
 api.route(OrderRelationship, 'order_event', '/orders/<identifier>/relationships/event')
 api.route(OrderRelationship, 'order_marketer', '/orders/<identifier>/relationships/marketer')
 api.route(OrderRelationship, 'order_discount', '/orders/<identifier>/relationships/discount-code')
+
+# Event Analytics API
+api.route(EventAnalyticsGeneralDetail, 'event_analytics_general_detail', '/analytics/events/<int:id>/general',
```
niranjan94: Since we only have `GET` endpoints for this, make it such that it exists as a relationship to event only. 

`/events/<identifier>/general-analytics`. 

Follow json-api naming convention and consider this a related resource of event.

`/parent-resouces/<id>/related-resource`

shubham-padia: I did that initially if you see the PR email,  but hen as https://github.com/fossasia/open-event-orga-server/issues/4239 came up, I changed it to `analytics/events/{event_identifier}/general` so that all such endpoints fall under a common `/analytics` prefix. What do you think :) ?

shubham-padia: >  as a relationship to event only.

I'm not sure if that's possible currently :grimacing: , FRJ allows relationship only for existing sqlalchemy relationships

niranjan94: But according to https://flask-rest-jsonapi.readthedocs.io/en/latest/resource_manager.html#resourcerelationship , You can change the datalayer used. So, it should be possible to use a custom data layer and mimic an sqlalchemy relationship.

niranjan94: Because ... Something like `analytics/events/{event_identifier}/general` is pretty divergent from json-api spec.

shubham-padia: I'll change the url, but mocking a relationship is not possible. Data layer will change the functions mentioned in the url, not the default marshmallow-jsonapi behaviour.

shubham-padia: @niranjan94 Changed the url, please review :)

app/api/__init__.py
```diff
@@ -398,3 +401,7 @@
 api.route(OrderRelationship, 'order_event', '/orders/<identifier>/relationships/event')
 api.route(OrderRelationship, 'order_marketer', '/orders/<identifier>/relationships/marketer')
 api.route(OrderRelationship, 'order_discount', '/orders/<identifier>/relationships/discount-code')
+
+# Event Analytics API
+api.route(EventAnalyticsGeneralDetail, 'event_analytics_general_detail', '/events/<int:id>/general-analytics',
```
niranjan94: Statistics would be a better suited term here than Analytics.

shubham-padia: @niranjan94 Yep, sounds better :). Resolved.

app/api/__init__.py
```diff
@@ -398,3 +401,7 @@
 api.route(OrderRelationship, 'order_event', '/orders/<identifier>/relationships/event')
 api.route(OrderRelationship, 'order_marketer', '/orders/<identifier>/relationships/marketer')
 api.route(OrderRelationship, 'order_discount', '/orders/<identifier>/relationships/discount-code')
+
+# Event Analytics API
```
shubham-padia: Thanks, resolved :)

shubham-padia:
@niranjan94 Please review :)

PR #4277, opened 2017-08-09
Fix migration file for already existing export_jobs table

### LINKED ISSUE #4276: [Sentry]: (psycopg2.ProgrammingError) relation "export_jobs" already exists
```
ProgrammingError: (psycopg2.ProgrammingError) relation "export_jobs" already exists
 [SQL: '\nCREATE TABLE export_jobs (\n\tid SERIAL NOT NULL, \n\ttask VARCHAR NOT NULL, \n\tstarts_at TIMESTAMP WITH TIME ZONE, \n\tuser_email VARCHAR, \n\tevent_id INTEGER, \n\tPRIMARY KEY (id), \n\tFOREIGN KEY(event_id) REFERENCES events (id) ON DELETE CASCADE\n)\n\n']
(25 additional frame(s) were not displayed)
...
  File "sqlalchemy/engine/base.py", line 1189, in _execute_context
    context)
  File "sqlalchemy/engine/base.py", line 1402, in _handle_dbapi_exception
    exc_info
  File "sqlalchemy/util/compat.py", line 203, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=cause)
  File "sqlalchemy/engine/base.py", line 1182, in _execute_context
    context)
  File "sqlalchemy/engine/default.py", line 470, in do_execute
    cursor.execute(statement, parameters)
```

magdalenesuo:
Fixes #4276


magdalenesuo:
@SaptakS I think this is a better way to fix existing/non-existing table problems rather than removing the file, because it so happens that some db has the table and some don't. please check

SaptakS:
@magdalenesuo agreed. This might happen depending on whether the db was created while user was in development branch or nextgen branch since the model imports are different in the different branches. Thanks for suggesting this. Awesome!!

PR #4304, opened 2017-08-12
Pages API: Only allow `footer` or `event` in `places` attribute

### LINKED ISSUE #4303: Pages API: Only allow `footer` or `event` in `places` attribute

shubham-padia:
fixes #4303


shubham-padia:
@fossasia/open-event-server-devs Please review :)

PR #4308, opened 2017-08-12
PATCH fix: thumbnail/large/icon image-url changes to null if provided

### LINKED ISSUE #4307: Bug: thumbnail/large/icon image-url changes to null if provided in PATCH

shubham-padia:
fixes #4307 

Changes:
The fields are already read only, no need to delete them. 

cc: @iamareebjamal No error will be produced as is the standard with read-only fields, they will be ignored


shubham-padia:
@SaptakS Done :)

PR #4310, opened 2017-08-12
Add email to valid types in custom-form

### LINKED ISSUE #4309: Add email to valid types in custom-form
**Current**
Currently we are not able to set an email type to the custom-form which leads to `Error: 422`.

**Expected**
email should be a valid type for the custom-form

shubham-padia:
fixes #4309


shubham-padia:
@CosmicCoder96 @geekyd Can you please suggest a list of additional types, you guys would be needing ?

PR #4311, opened 2017-08-12
Add roles like organizers,etc to events' relationship

### LINKED ISSUE #4266: Add missing roles relationship to events
- [ ] bug report
- [x] feature request

Add missing roles relationship to events for admin/events
Related issue:
https://github.com/fossasia/open-event-frontend/issues/631

shubham-padia:
fixes #4266 :P 
The related url is `/v1/users` in all cases, having separate endpoints for this doesn't make any sense

The reverse relationships in the users API will be done in separate PR.


shubham-padia:
Nope. We do not want all those users everytime we make a request. And i don't think sqlalchemy relationships get stored in the db :)

SaptakS:
No. The relationships don't get stored. But the backref mapping does as far as I know. Secondly what about the permissions? Is everyone allowed to see who is the organizer and the details of the organizer and so on?

shubham-padia:
@SaptakS The permission was wrong for the relationship class, I've modified it

Issue #4373, opened 2017-08-17
Paypal payment mode giving 500 error in Orders API


SaptakS:
**I'm submitting a ...**  (check one with "x")
- [x] bug report ( Follow up of #4361 )
- [ ] feature request
- [ ] support request => Please do not submit support requests here, instead ask your query in out Gitter channel at https://gitter.im/fossasia/open-event-orga-server

**Current behavior:**

Following is the heroku log for making a request to Orders API with payment-mode:paypal

```
2017-08-17T11:24:19.730791+00:00 app[web.1]: Exception on /v1/orders [POST]
2017-08-17T11:24:19.730793+00:00 app[web.1]: Traceback (most recent call last):
2017-08-17T11:24:19.730794+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
2017-08-17T11:24:19.730794+00:00 app[web.1]:     response = self.full_dispatch_request()
2017-08-17T11:24:19.730795+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
2017-08-17T11:24:19.730796+00:00 app[web.1]:     rv = self.handle_user_exception(e)
2017-08-17T11:24:19.730797+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/flask_cors/extension.py", line 161, in wrapped_function
2017-08-17T11:24:19.730800+00:00 app[web.1]:     return cors_after_request(app.make_response(f(*args, **kwargs)))
2017-08-17T11:24:19.730801+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
2017-08-17T11:24:19.730802+00:00 app[web.1]:     reraise(exc_type, exc_value, tb)
2017-08-17T11:24:19.730803+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
2017-08-17T11:24:19.730804+00:00 app[web.1]:     rv = self.dispatch_request()
2017-08-17T11:24:19.730804+00:00 app[web.1]:   File "/app/.heroku/python/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
[...]

PR #4374, opened 2017-08-17
Fix payment mode 500 error in orders api

magdalenesuo:
Might Fix this #4373

Can you please check again if this fixes the error, after merge @SaptakS


SaptakS:
@magdalenesuo looks good to me. Will let you know if it breaks again.

PR #4445, opened 2017-08-28
Fix attendee POST request

### LINKED ISSUE #4441: Not allowing attendee POST request
**I'm submitting a ...**
- [x] bug report

**Current behavior:**
Created a new event with new user account. Created a free ticket for the event. for create attendee request the response is:
```
"errors": [
        {
            "status": 403,
            "source": {
                "source": "event_id"
            },
            "detail": "Access Forbidden",
            "title": "Access Forbidden"
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}
```

magdalenesuo:
Fixes  #4441

### DIFF
--- a/app/api/attendees.py
+++ b/app/api/attendees.py
@@ -30,9 +30,8 @@ class AttendeeListPost(ResourceList):
         :return:
         """
         require_relationship(['ticket', 'event'], data)
-        if not has_access('is_coorganizer', event_id=data['event']):
-            raise ForbiddenException({'source': 'event_id'}, "Access Forbidden")
 
+    decorators = (jwt_required,)
     methods = ['POST']
     schema = AttendeeSchema
     data_layer = {'session': db.session,


app/api/attendees.py
```diff
@@ -23,8 +23,8 @@ class AttendeeListPost(ResourceList):
 
     def before_post(self, args, kwargs, data):
         require_relationship(['ticket', 'event'], data)
-        if not has_access('is_coorganizer', event_id=data['event']):
-            raise ForbiddenException({'source': 'event_id'}, "Access Forbidden")
+        if not has_access('is_user_itself', id=current_identity.id):
```
poush: Didn't get this. This check is basically translating as "Is the logged in user is same as the logged in user?"

magdalenesuo: @poush yes. Its basically checks logged in like jwt_required.

poush:
@magdalenesuo  So why not just jwt_required decorator at top of before_post ?

magdalenesuo:
@poush  Yes I can do that. :D I just kept the  `is_user_itself` permission cause message was modifiable. That's all. No other reason. :laughing:  I can make use of `jwt_required` directly. Should I change?

poush:
Actually, it will not work for some cases because current_activity will be null before calling jwt_required and btw we have `auth_required` permission in permission manager :P which does exactly what you need here

PR #4643, opened 2018-03-06
Fixes #4636 Resolves SERVER_NAME Issue

### LINKED ISSUE #4636: Remove SERVER_NAME or add warning in docs
The `SERVER_NAME` config in example env shadows the possible IP addresses a machine may take. For example, if the local IP is `10.0.3.2` the SERVER_NAME parameter being set to `localhost` will shadow this and 404 error will be shown instead.

This setting should either be removed or a warning should be added in the docs as it can be quite painful to pin point what the issue is

ParthS007:
Fixes #4636 

#### Checklist

- Presently, the server shows the server name `127.0.0.1:5000` irrespective of `SERVER_NAME` Variable in `env` and due to this, it gives `404` error.
image

- Now after running the server, It shows provides the link of `SERVER_NAME` that is used in `.env` 
- I have used `localhost:5000` and now it shows:
image

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,5 +1,5 @@
 from __future__ import print_function
-
+from flask_script import Server
 import os
 
 from app.api.helpers.db import save_to_db
@@ -94,4 +94,8 @@ def prepare_kubernetes_db():
 
 
 if __name__ == "__main__":
+    host_port = app.config['SERVER_NAME'].split(":")
+    host = host_port[0]
+    port = int(host_port[1] if len(host_port) > 1 else app.config['FALLBACK_PORT'])
+    manager.add_command("runserver", Server(host=host, port=port))
     manager.run()
--- a/config.py
+++ b/config.py
@@ -49,7 +49,8 @@ class Config(object):
     SQLALCHEMY_TRACK_MODIFICATIONS = True
     ERROR_404_HELP = False
     CSRF_ENABLED = True
-    SERVER_NAME = env('SERVER_NAME', default=None)
+    SERVER_NAME = env('SERVER_NAME', default='127.0.0.1:5000')
+    FALLBACK_PORT = 80
     CORS_HEADERS = 'Content-Type'
     SQLALCHEMY_DATABASE_URI = env('DATABASE_URL', default=None)
     SERVE_STATIC = env.bool('SERVE_STATIC', default=False)


ParthS007:
@shubham-padia  Travis is not able to read `env` file, that's why Travis is failing. Please help to resolve this.

ParthS007:
Please don't  merge this, work in progress :+1:

ParthS007:
@shubham-padia @iamareebjamal @SaptakS  Please review. Thanks :+1:

manage.py
```diff
@@ -94,4 +96,12 @@ def prepare_kubernetes_db():
 
 
 if __name__ == "__main__":
+    _host = '127.0.0.1'
```
SaptakS: Don't mention the _host and _port hard-coded here.  Rather put it in the default for SERVER_NAME in config.py

SaptakS: Once you do that, the entire `if` logic below won't be necessary anymore.

manage.py
```diff
@@ -1,7 +1,9 @@
 from __future__ import print_function
-
+from flask_script import Server
 import os
 
+SERVER_NAME = __import__('config').Config.SERVER_NAME
```
SaptakS: You don't need to do this. Instead, you can use `app.config['SERVER_NAME']`

manage.py
```diff
@@ -94,4 +96,12 @@ def prepare_kubernetes_db():
...
 
 if __name__ == "__main__":
+    _host = '127.0.0.1'
+    _port = 5000
+    if SERVER_NAME:
+        host, port = SERVER_NAME.split(':')
+    else:
+        host = _host
+        port = _port
+    manager.add_command("runserver", Server(use_debugger=True, use_reloader=True, host=host, port=int(port)))
```
SaptakS: I don't think `use_debugger` and `use_reloader` is needed.

manage.py
ParthS007: @SaptakS I have tried it, but not able to import it. Then I have used this.

SaptakS: You don't need to import. You are already importing app. And the configurations are already present in the flask app. So you can directly do something like `host = app.config['SERVER_NAME'].split(":")[0]` and `port = app.config['SERVER_NAME'].split(":")[1]`

ParthS007: Got it @SaptakS

manage.py
```diff
@@ -94,4 +95,7 @@ def prepare_kubernetes_db():
 
 
 if __name__ == "__main__":
+    host = app.config['SERVER_NAME'].split(":")[0]
+    port = app.config['SERVER_NAME'].split(":")[1]
```
iamareebjamal: This will fail if the port is not specified in the configuration

iamareebjamal: Also, the splitting is done twice instead of once

ParthS007: @iamareebjamal It will take the default value i.e `127.0.0.1:5000` if we didn't specify the `SERVER_NAME` in `.env`

iamareebjamal:
```python
host_port = app.config['SERVER_NAME'].split(":")
host = host_port[0]
if len(host_port) > 1:
     port = host_port[1]
else:
    port = 80
```

SaptakS: I don't think one should write a SERVER_NAME without mentioning the port in the .env file. But yes, it would be better to handle that case, maybe using a ternary operator.

SaptakS: @iamareebjamal I think instead of hard-coding 80 here, it is better to store that also in a default config key.

ParthS007: @SaptakS @iamareebjamal I am confused what I have to do finally :confused:

rafalkowalski: @IamAdiSri You can use urlparse library to get host name and port number.
```
from urlparse import urlparse
server_name = app.config.get('SERVER_NAME')
host_name = urlparse(server_name).hostname
port = urlparse(server_name).port
```

iamareebjamal: @ParthS007 Where you have modified the SERVER_NAME parameter in config.py, add a FALLBACK_PORT=80 parameter as well and use that here

manage.py
```diff
@@ -95,7 +94,8 @@ def prepare_kubernetes_db():
 
 
 if __name__ == "__main__":
-    host = app.config['SERVER_NAME'].split(":")[0]
-    port = app.config['SERVER_NAME'].split(":")[1]
+    host_port = app.config['SERVER_NAME'].split(":")
+    host = host_port[0]
+    port = int(host_port[1]) if len(host_port) > 1 else app.config['FALLBACK_PORT']
```
iamareebjamal: `host_port[1]` will raise exception if not present

manage.py
```diff
@@ -94,4 +94,8 @@ def prepare_kubernetes_db():
 
 
 if __name__ == "__main__":
+    host_port = app.config['SERVER_NAME'].split(":")
+    host = host_port[0]
+    port = int(host_port[1]) if len(host_port) > 1 else app.config['FALLBACK_PORT']
+    manager.add_command("runserver", Server(use_debugger=True, use_reloader=True, host=host, port=int(port)))
```
iamareebjamal: A minor suggestion, can be changed to 
`int(host_port[1] if len(host_port) > 1 else app.config['FALLBACK_PORT'])`
and no need of `int(port)` will be below

PR #4674, opened 2018-04-18
refactor: Final stage changes to Python 2/3

### LINKED ISSUE #4677: Final refactoring to make project compatible with Python 2/3
These things are required to make project compatible with Python 2/3

- [x] Import future package builtins and aliases (eg. str, io, urllib)
- [x] Wrap byte return type with str(<byte>, 'utf-8') for Python 3
- [x] Wrap str type with byte(<string>) for IO functions
- [x] Replace StringIO and cStringIO with io.BytesIO for cross compatibility

iamareebjamal:
Fixes #4677 

Changes: 
- Import future package builtins and aliases (eg. str, io, urllib)
- Wrap byte return type with str(<byte>, 'utf-8') for Python 3
- Wrap str type with byte(<string>) for IO functions
- Replace StringIO and cStringIO with io.BytesIO for cross compatibility


iamareebjamal:
@poush @SaptakS @shubham-padia @niranjan94 This runs on Python3, however I don't know each and every use case of server, so please test

SaptakS:
This PR definitely makes me very very happy. ✨ ❤️ I would just want to review it once completely before merging. But awesome work. Finally, someone did it!!! In fact, I feel we should remove python 2 support completely. @niranjan94 @iamareebjamal @hongquan what do you think?

niranjan94:
@iamareebjamal Awesome work ! 🎉 😄 

---

I think now would be a great time for us to remove the python 2 code. I get why one might be hesitant to do so, but, I think removing py2 code and going with _only_ py3 code and fixing bugs as and when we find them, would be much easier than to have both next to each other.

iamareebjamal:
@niranjan94 So do you want me to do it in this PR?

iamareebjamal:
@niranjan94 Yes, I agree with the philosophy of python3 only. So do you want me to do it in this PR?

niranjan94:
Awesome. You can do it however it is comfortable for you. :)

Same PR or a separate one.

iamareebjamal:
@niranjan94 @SaptakS @poush @shubham-padia Please review

iamareebjamal:
@niranjan94 @SaptakS @poush @shubham-padia Please review. Let's merge this :)

iamareebjamal:
@niranjan94 Should this be merged?

niranjan94:
@iamareebjamal yes. We can merge this. merging now.

PR #4678, opened 2018-04-20
fix: Raise HTTP 409 on Create Ticket with Duplicate Name

### LINKED ISSUE #4672: Error 500 on Create Ticket : Unique name
- [x] bug report
- [ ] feature request

**Current behavior:**
TIcket creation with same name causes error 500
```json
{
    "errors": [
        {
            "status": 500,
            "source": {
                "pointer": "/data"
            },
            "detail": "Object creation error",
            "title": "Unknown error"
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}
```

**Expected behavior:**
There is a unique constraint on name and event id, so this should be checked before saving and shown to the user 

**Steps to reproduce:**
Try to create two tickets of the same name for the same event

Masquerade0097:
Fixes #4672 

#### Checklist

When a user tries to create ticket with same name as some other ticket for the same event then ConflictException is raised instead of HTTP 500.

- Raise ConflictException on duplicate Ticket name.

@iamareebjamal @shubham-padia @poush Please review.

### DIFF
--- a/app/api/tickets.py
+++ b/app/api/tickets.py
@@ -12,6 +12,8 @@ from app.models.access_code import AccessCode
 from app.models.order import Order
 from app.models.ticket import Ticket, TicketTag, ticket_tags_table
 from app.models.ticket_holder import TicketHolder
+from app.api.helpers.exceptions import ConflictException
+from app.api.helpers.db import get_count
 
 
 class TicketListPost(ResourceList):
@@ -30,6 +32,9 @@ class TicketListPost(ResourceList):
         if not has_access('is_coorganizer', event_id=data['event']):
             raise ObjectNotFound({'parameter': 'event_id'},
                                  "Event: {} not found".format(data['event_id']))
+
+        if get_count(db.session.query(Ticket.id).filter_by(name=data['name'], event_id=int(data['event']))) > 0:
+            raise ConflictException({'pointer': '/data/attributes/name'}, "Ticket already exists")
 
     schema = TicketSchema
     methods = ['POST', ]


app/api/tickets.py
```diff
@@ -31,6 +32,10 @@ def before_post(self, args, kwargs, data):
             raise ObjectNotFound({'parameter': 'event_id'},
                                  "Event: {} not found".format(data['event_id']))
 
+        query = db.session.query(Ticket.id).filter_by(name=data['name'])
```
niranjan94: You could use the `get_count` helper at https://github.com/fossasia/open-event-server/blob/development/app/api/helpers/db.py#L69 instead of this

Masquerade0097: Ya that will be better.

Masquerade0097:
@niranjan94 Please review.

app/api/tickets.py
```diff
@@ -31,6 +33,9 @@ def before_post(self, args, kwargs, data):
             raise ObjectNotFound({'parameter': 'event_id'},
                                  "Event: {} not found".format(data['event_id']))
 
+        if get_count(db.session.query(Ticket.id).filter_by(name=data['name'], event_id=int(data['event']))):
```
iamareebjamal: This should be get_count(...) > 0 ?

Masquerade0097: In case `get_count` return `0`, the if condition fails and the body does not get executed. So why is this change needed ?

iamareebjamal: If it returns 0, it means there is no ticket of this name, so error should not be raised

Masquerade0097: Error is geting raised only when `get_count is > 0`. In case of 0 error is not getting raised.

PR #4679, opened 2018-04-21
fix: Add defaults field values before validating schema

### LINKED ISSUE #4655: Wrong schema validation
- [x] bug report
- [ ] feature request

**Current behavior:**
Schema validation on the server is broken

```JSON
{
  "data": {
    "relationships": {
      "event": {
        "data": {
          "type": "event",
          "id": "{{event}}"
        }
      }
    },
    "attributes": {
      "name": "ticket-name",
      "description": "some description",
      "type": "VIP-ticket",
      "price": "100.00",
      "quantity": "100",
      "is-description-visible": null,
      "position": "1",
      "is-fee-absorbed": "false",
      "sales-starts-at": "2017-06-01T01:24:47.500127+00:00",
      "sales-ends-at": "2017-06-30T00:24:47.500127+00:00",
      "is-hidden": "false",
      "min-order": "1",
      "max-order": "100"
    },
    "type": "ticket"
  }
}
```
If any of the optional field is null (which is what it can be when serializing in Java or any other language with schema, it gives the following error

```JSON
{
    "jsonapi": {
        "version": "1.0"
    },
    "errors": [
        {
            "status": "422",
            "source": {
                "pointer": "/data/attributes/is-description-visible"
            },
            "title": "Validation error",
            "detail": "Field may not be null."
        }
    ]
}
```

However this is clearly an optional field. Also, all validation is done prematurely and before applying the default values, which is counter productive. If quantity is missing, it should be assigned the default value from the server rather than reporting `quantity should be higher than min-order`

Some of these changes in API are breaking and make it really hard developing mobile applications, because we cannot base our development on a particular API assuming it won't break any contracts. master branch is unusable so we switched to development, but it introduces several breaking changes without any communication in documentation.

Several times, the development in mobile (and I assume web) is halted with questions as we are not able to POST ticket, feedback, event because of some reason, and it prevents us to create new features or move forward in a direction we believe API will move. But we cannot because it is unpredictable what will change and how will we accomodate with it. Now we are essentially stopped in our tracks until the server is fixed because not everyone knows how to setup API locally and even if they do, they don't know which branch or commit they should target? And it might be easy to develop with local API on web, it isn't on mobile

~~I believe this is a side effect of the recent change in flask rest jsonapi, and the changes made in local fork might be necessary and convenient but they break several previously established behaviours~~. I think we should agree that APIs should conform to their users and not the other way around, so I urge the developers to ensure that any such breaking changes be well tested before merging and communicated in some form of documentation. Thanks

**Edit:** The dev branch was not deployed for a long time to heroku, so all the breaking changes throughout the months got introduced to us in last few weeks.

srv-twry:
Fixes #4655 

#### Checklist

Fixes the wrong schema validation. Default values are now added to the data before deserializing it.

- Add default values to the fields before schema validation

**TODO**:

- [x] Add the decorator to all the schema classes
- [x] Write a few tests
- [x] Add necessary docstrings

~~~@iamareebjamal I have opened the WIP PR so as to get feedback on the way the decorator has been implemented. Is it fine ? Moreover suggestions of the name of the decorator is also welcomed.~~~

### DIFF
--- a/app/api/schema/custom_forms.py
+++ b/app/api/schema/custom_forms.py
@@ -3,8 +3,10 @@ from marshmallow_jsonapi import fields
 from marshmallow_jsonapi.flask import Schema, Relationship
 
 from app.api.helpers.utilities import dasherize
+from utils.common import use_defaults
 
 
+@use_defaults()
 class CustomFormSchema(Schema):
     """
     API Schema for Custom Forms database model
--- a/app/api/schema/email_notifications.py
+++ b/app/api/schema/email_notifications.py
@@ -2,8 +2,10 @@ from marshmallow_jsonapi import fields
 from marshmallow_jsonapi.flask import Schema, Relationship
 
 from app.api.helpers.utilities import dasherize
+from utils.common import use_defaults
 
 
+@use_defaults()
 class EmailNotificationSchema(Schema):
     """
     API Schema for email notification Model
--- a/app/api/schema/event_invoices.py
+++ b/app/api/schema/event_invoices.py
@@ -4,8 +4,10 @@ from marshmallow_jsonapi.flask import Schema, Relationship
 
 from app.api.helpers.static import PAYMENT_COUNTRIES
 from app.api.helpers.utilities import dasherize
+from utils.common import use_defaults
 
 
+@use_defaults()
 class EventInvoiceSchema(Schema):
     """
     Event Invoice API Schema based on event invoice model
--- a/app/api/schema/events.py
+++ b/app/api/schema/events.py
@@ -9,8 +9,10 @@ from flask_rest_jsonapi.exceptions import ObjectNotFound
 from app.models.event import Event
 from app.api.helpers.exceptions import UnprocessableEntity
 from app.api.helpers.utilities import dasherize
+from utils.common import use_defaults
 
 
+@use_defaults()
 class EventSchemaPublic(Schema):
     class Meta:
         type_ = 'event'
--- a/app/api/schema/image_sizes.py
+++ b/app/api/schema/image_sizes.py
@@ -2,8 +2,10 @@ from marshmallow_jsonapi import fields
 from marshmallow_jsonapi.flask import Schema
 
 from app.api.helpers.utilities import dasherize
+from utils.common import use_defaults
 
 
+@use_defaults()
 class ImageSizeSchema(Schema):
     """
     Api schema for image_size Model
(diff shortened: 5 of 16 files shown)


srv-twry:
@iamareebjamal Ping.

srv-twry:
@niranjan94 @shubham-padia 
Please review so that i can add the decorator to all the models. 
For the context, please refer to the discussion in the issue.

srv-twry:
@niranjan94 @iamareebjamal Ready for review

tests/unittests/utils/test_common.py
```diff
@@ -0,0 +1,31 @@
...
+from pytz import timezone
+
+from app import current_app as app
+from app.api.schema.tickets import TicketSchema
+from tests.unittests.setup_database import Setup
+from tests.unittests.utils import OpenEventTestCase
+from utils.common import patch_defaults
+
+
+class TestUtils(OpenEventTestCase):
```
iamareebjamal: If it does not depend on the db, don't use OpenEventTestCase

maxlorenz:
One minor thing: Would it be better to use a new schema for the test only as every schema change in the original can break this (somewhat unrelated) test?

iamareebjamal:
Yup I agree. Test should have a test schema. And a couple more tests actually. To test the output as well

srv-twry:
@iamareebjamal @maxlorenz Please review again. I have a separate TestSchema and tests using it.

tests/unittests/test_schema.py
```diff
@@ -0,0 +1,21 @@
+from marshmallow_jsonapi import fields
+from marshmallow_jsonapi.flask import Schema
+
+from app.api.helpers.utilities import dasherize
+
+
+class TestSchema(Schema):
```
iamareebjamal: Make a file `test_schema_defaulting` and move both the class and tests in it

srv-twry:
@iamareebjamal Please suggest changes when you are free :)

app/api/schema/custom_forms.py
```diff
@@ -3,8 +3,10 @@
 from marshmallow_jsonapi.flask import Schema, Relationship
 
 from app.api.helpers.utilities import dasherize
+from utils.common import use_defaults
 
 
+@use_defaults()
```
iamareebjamal: Looks good. Remove paranthesis, as they are not needed

srv-twry: It need ```self``` right ?

srv-twry:
@niranjan94 Please review and merge at your comfort :)

PR #4682, opened 2018-04-22
fix : "last_name" in create attendee api is now required

mayank8318:
Fixes #4676 

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The unit tests pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

- [x] All the functions created/modified in this PR contain relevant docstrings.

#### Short description of what this resolves:
"last_name" field in create attendee api is a required field now, as stated in the docs.

#### Changes proposed in this pull request:

- Make "last_name" a required field since the docs state it as required.

PR #4691, opened 2018-05-02
fix: Migrate existing DB data to ensure a non-nullable last_name field for attendees

### LINKED ISSUE #4688: Migrate existing DB data to ensure a non-nullable last_name field for attendees
Follow up to #4682

> @shubham-padia: This might create problems when migrating the existing database as there maybe attendees without any last name. We need a migration to handle that. I'm not sure right now how we should go about that. Till we figure it out, we can make the field required in the API schema and let the db schema be as it was before. A follow-up issue can be made to handle change in the db schema.

mayank8318:
Fixes #4688 

#### Checklist

Last name field was made required in the API as per PR #4682. To make it a non null field in the database, custom migration had to be implemented to handle existing data with null last_name.

-Before altering the column and making it non-null, update all existing "null" last names to a space character " ".


mayank8318:
@niranjan94 @poush Please review :)

PR #4705, opened 2018-05-09
Add port specification in image urls

### LINKED ISSUE #4646: Locally generated image URLs don't take port into account
- [x] bug report
- [ ] feature request

**Current behavior:**

The URLs generated locally do not take port of the server into account. For instance, the sample event payload with image URL is converted to this format `http://localhost/static/media/events/1/large/eTNEbFJZOF/3827c2eb-5fcf-4ea3-ad46-bff1c10b69bf.jpg`

This is without the port number of the server and will create problems in consuming local APIs

**Expected behavior:**

The generated URLs should take the port of the server into account

**Steps to reproduce:**

Create an event with the example payload on locally running server

schedutron:
Fixes #4646

- I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- My branch is up-to-date with the Upstream `development` branch.

Events created on `localhost` had image urls without port specification. This PR adds the missing port specification.

- Add port number after `hostname` in locally generated image urls


app/api/helpers/storage.py
```diff
@@ -162,7 +162,7 @@ def upload_local(uploaded_file, key, **kwargs):
         return get_settings()['static_domain'] + \
             file_relative_path.replace('/static', '')
     url = urlparse(request.url)
-    return url.scheme + '://' + url.hostname + file_relative_path
+    return '%s://%s:%s%s' % (url.scheme, url.hostname, url.port, file_relative_path)
```
rafalkowalski: Use '{scheme}://{hostname}:{port}{file_relative_path}'.format(scheme=url.scheme, hostname=url.hostname, port=url.port, file_relative_path=file_relative_path)

schedutron: @rafalkowalski Thanks, I'll update it.

app/api/helpers/storage.py
```diff
@@ -162,7 +162,9 @@ def upload_local(uploaded_file, key, **kwargs):
         return get_settings()['static_domain'] + \
             file_relative_path.replace('/static', '')
     url = urlparse(request.url)
-    return url.scheme + '://' + url.hostname + file_relative_path
+    return '{scheme}://{hostname}:{port}{file_relative_path}'.format(
```
iamareebjamal: If the port is 80 and scheme is http and if the port is 443 and scheme is https, we can omit the :port 

I think this should be accomodated

schedutron:
@iamareebjamal The tests are failing because they are looking for filenames with ports, which are absent. Should I add port number while creating these files or should I strip the port numbers in the test code before comparisons take place?

iamareebjamal:
Port number was not available before. So I can't understand. Files should be independent of port number. Just url should have that. Can you post an assertion failure here.

schedutron:
Here's an error from my local `nosetests` run:
```
======================================================================
ERROR: test_create_save_image_sizes (tests.unittests.api.helpers.test_files.TestFilesHelperValidation)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/binaryboy/Desktop/open-event-server/tests/unittests/api/helpers/test_files.py", line 138, in test_create_save_image_sizes
    resized_width_large, _ = self.getsizes(resized_image_file_large)
  File "/Users/binaryboy/Desktop/open-event-server/tests/unittests/api/helpers/test_files.py", line 22, in getsizes
    im = Image.open(file)
  File "/usr/local/lib/python3.6/site-packages/PIL/Image.py", line 2312, in open
    fp = builtins.open(filename, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '/Users/binaryboy/Desktop/open-event-server:5000/static/media/events/53b8f572-5408-40bf-af97-6e9b3922631d/large/UFNNeW5FRF/5980ede1-d79b-4907-bbd5-17511eee5903.jpg'
```

iamareebjamal:
Seems like same method is used to get the file from file system. This should not be done. Check why is it happening

app/api/helpers/files.py
```diff
@@ -113,6 +114,8 @@ def create_save_resized_image(image_file, basewidth=None, maintain_aspect=None,
     uploaded_url = upload(upfile, upload_path)
     os.remove(temp_file_path)
 
+    uploaded_url = re.sub(r'\:\d+/', '/', uploaded_url)
```
schedutron: Is this the right place to strip the port? I'm unsure as this is outside the test framework. Also, this is different from the code for stripping in the case of protocol-port matches you mentioned earlier (`http`: `80` and so on). I'll add code for it after this part gets reviewed :)

iamareebjamal: I don't think you should really strip the port. This is the uploaded url being returned, it should have the port as well. You should check where the file is being loaded, the path should not have port in it

app/api/helpers/files.py
```diff
@@ -113,6 +114,8 @@ def create_save_resized_image(image_file, basewidth=None, maintain_aspect=None,
     uploaded_url = upload(upfile, upload_path)
     os.remove(temp_file_path)
 
+    uploaded_url = re.sub(r'\:(\d+)|None/', '/', uploaded_url)
```
schedutron: @iamareebjamal Is this the right place for stripping the port? I'm unsure as this is outside the test framework. Now the files fetched don't involve port number in their names.

app/api/helpers/storage.py
```diff
@@ -162,7 +162,9 @@ def upload_local(uploaded_file, key, **kwargs):
...
+        scheme=url.scheme, hostname=url.hostname, port=url.port,
+        file_relative_path=file_relative_path).replace(':None', '')
```
maxlorenz: can we replace `:None` here with the empty string? This would remove the need for a generic placeholder that can be stripped

schedutron: @maxlorenz So `:None` should be retained?

app/api/helpers/files.py
iamareebjamal: I think port shouldn't be stripped here at all. Because this is returning the url as well. It should be handled correctly when opening file. How is the url affecting the file name. This is the problem

tests/unittests/api/helpers/test_files.py
```diff
@@ -111,6 +112,7 @@ def test_create_save_resized_image(self):
             upload_path = 'test'
             resized_image_url = create_save_resized_image(image_url_test, width, aspect_ratio, height, upload_path, ext='png')
             resized_image_file = app.config.get('BASE_DIR') + resized_image_url.split('/localhost')[1]
+            resized_image_file = re.sub(r'\:(\d+)|None/', '/', resized_image_file)
```
schedutron: I've moved the port stripping part here.

tests/unittests/api/helpers/test_files.py
```diff
@@ -125,6 +127,9 @@ def test_create_save_image_sizes(self):
             width_icon = 75
             image_sizes = create_save_image_sizes(image_url_test, image_sizes_type)
 
+            for image in image_sizes:  # So that file names below don't contain port spec.
+                image_sizes[image] = re.sub(r'\:(\d+)|None/', '/', image_sizes[image])
```
schedutron: And here. Are these the right places?

maxlorenz: As far as the standard is concerned, a URL-port string can contain _zero_ or more digits. This would not be stripped and stripping just `:` would not make much sense. Imho, we should use [`urllib.parse`](https://docs.python.org/3.0/library/urllib.parse.html) here to remove the port

app/api/helpers/storage.py
maxlorenz: Well `:None` is not a valid port while the empty string is so I guess having `:None` does not really do much here?

tests/unittests/api/helpers/test_files.py
schedutron: @maxlorenz I cannot think of a string which would break the above regex. Can you please provide an example that will? I'm open to using `urllib.parse`, if that's more readable and Pythonic.

maxlorenz: Well the only I can find would be the empty port example `http://test.com:/`. But urllib.parse seems more pythonic to me (the existing solution should work too, however). Maybe put the regex in a helper function for readability

schedutron: @maxlorenz I suspect such urls would never be generated. Also, I'm not able to think of a one-liner involving `urllib.parse`, so maybe wrapping in a helper function is the right way to go?

schedutron: @maxlorenz Anyways, are these the right places for stripping the ports off?

iamareebjamal: @schedutron What does the function return. Can you paste the content here? I'm trying to wrap my head around as to why is there a need to strip ports.

schedutron: @iamareebjamal If the ports aren't stripped, the tests code tries looking for file paths with ports, and thus fails.

tests/unittests/api/helpers/test_files.py
```diff
@@ -111,6 +112,7 @@ def test_create_save_resized_image(self):
             upload_path = 'test'
             resized_image_url = create_save_resized_image(image_url_test, width, aspect_ratio, height, upload_path, ext='png')
             resized_image_file = app.config.get('BASE_DIR') + resized_image_url.split('/localhost')[1]
```
iamareebjamal: The problem is actually here. This is assuming the localhost part of the URL to be static. You can split the URL using urlpase instead of this logic and have risk free path without regex replacement

```python
from urllib.parse import urlparse
o = urlparse('http://www.cwi.nl:80/%7Eguido/Python.html')
>> o.path
'/%7Eguido/Python.html'
```

tests/unittests/api/helpers/test_files.py
iamareebjamal: Got it, use urlparse here as well

iamareebjamal:
```python
image_sizes = [urlparse(url).path for url in create_save_image_sizes(image_url_test, image_sizes_type)]

File: app/api/helpers/storage.py [hunk 9]
```diff
@@ -162,7 +164,15 @@ def upload_local(uploaded_file, key, **kwargs):
...
-    return url.scheme + '://' + url.hostname + file_relative_path
+
+    # No need to specify scheme-corresponding port
+    port = url.port
+    if port and url.scheme == SCHEMES.get(url.port, None):
+        port = None
+
+    return '{scheme}://{hostname}:{port}{file_relative_path}'.format(
+        scheme=url.scheme, hostname=url.hostname, port=port,
+        file_relative_path=file_relative_path).replace(':None', '')
```
iamareebjamal: Looks good, if you can add tests for this logic, it'd be great

iamareebjamal: Create 3-4 new ones testing different types of URL, you'll have to move this logic in a different function for isolated testing

schedutron: @iamareebjamal Just curious, shouldn't test methods `test_upload_image_local()`, `test_upload_single_file()` and `test_upload_multiple_file()` be in a different test file, say `test_storage.py`?

schedutron: @iamareebjamal How do I simulate requests going to different ports during tests?

schedutron: Okay, I thought so because the url is generated via the `request.url` call within this function.

iamareebjamal: That's the point of Unit Testing. Isolate components from real-world dependencies so that they can be tested for each and every use case. Unit tests use mock arguments to test the correctness of a function so that they can check it is working as expected without simulating real world constraints

schedutron: Thanks, here's what I've come up with so far:
```
def test_upload_local(self):
    upfile = UploadedFile('/path/to/image.png', 'image.png')
    with app.test_request_context():
        upload_url = upload_local(upfile, key='test')
    ...
```
The value of `upload_url` at this point is `http://localhost:5000/static/media/test/clZCY2xkNz/image.png`. Now I want to have different values of port so that I can check the match between expected and actual urls for different ports and schemes. I'm working on tuning the port parameter.

tests/unittests/api/helpers/test_storage.py
```diff
@@ -0,0 +1,51 @@
...
+            upload_url = upload_local(upfile, key=test_key)
+
+        expected_url = urljoin('http://localhost:5000',
+                               os.path.join('static/media/',
+                                            test_key,
+                                            test_hash,
+                                            test_name)
+                              )
+
+        self.assertEqual(upload_url, expected_url)
```
schedutron: @iamareebjamal Am I going in the right direction now?

schedutron: @iamareebjamal Am I going in the right direction now?

tests/unittests/api/helpers/test_storage.py
```diff
@@ -0,0 +1,55 @@
...
+
+        expected_url = urljoin(
+            '{url.scheme}://{url.netloc}'.format(url=urlparse(test_url)),
+            os.path.join('static/media/',
+            test_key,
+            test_hash,
+            test_name)
+            )
+
+        self.assertEqual(upload_url, expected_url)
```
schedutron: @iamareebjamal Am I going in the right direction now?

iamareebjamal: You don't need app context and all of this to test that feature. This is an integration test, they are expensive, slow and not needed to test a simple function.

Move that logic in a new function, let's say `create_url` with parameters request_url and file_relative_path

```
url = urlparse(request.url)

# No need to specify scheme-corresponding port
port = url.port
if port and url.scheme == SCHEMES.get(url.port, None):
    port = None

return '{scheme}://{hostname}:{port}{file_relative_path}'.format(
       scheme=url.scheme, hostname=url.hostname, port=port,
       file_relative_path=file_relative_path).replace(':None', '')
```

Just call this method and test the output string

app/api/helpers/storage.py
```diff
@@ -250,3 +252,17 @@ def generate_hash(key):
     """
     phash = generate_password_hash(key, get_settings()['secret'])
     return str(b64encode(phash), 'utf-8')[:10]  # limit len to 10, is sufficient
+
+
+def create_url(request_url, file_relative_path):
```
iamareebjamal: Good job, this should be as close to the calling method as possible, ideally just below it

tests/unittests/api/helpers/test_storage.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TestStorageHelperValidation(OpenEventTestCase):
+    """Test class for testing storage helper functions."""
+
+    def setUp(self):
+        """Method to set up app for testing."""
+        self.app = Setup.create_app()
+
+    def test_create_url(self):
```
iamareebjamal: One test should handle one URL. Create 3 different tests to check

- http 80
- https 443
- arbitrary port

tests/unittests/api/helpers/test_storage.py
```diff
@@ -0,0 +1,34 @@
+"""Test file for storage functions."""
+
+from app.api.helpers.storage import create_url
+from tests.unittests.setup_database import Setup
+from tests.unittests.utils import OpenEventTestCase
+
+
+class TestStorageHelperValidation(OpenEventTestCase):
```
iamareebjamal: We don't need database for this test, so don't extend OpenEventTestCase

tests/unittests/api/helpers/test_storage.py
```diff
@@ -0,0 +1,34 @@
...
+    """Test class for testing storage helper functions."""
+
+    def setUp(self):
+        """Method to set up app for testing."""
+        self.app = Setup.create_app()
```
iamareebjamal: No need for this as well

poush:
@schedutron from next onwards, make sure to squash your commits. There was no point to have 17-18 commits for one issue.

PR #4718, opened 2018-05-12
Patch it

Biswajee:
Added a db.commit statement and updated http links to https


niranjan94:
@Biswajee this is not required at all. `get_or_create` already takes care of committing the DB transaction. 

Closing the PR.

Also, always please create an issue first.

PR #4728, opened 2018-05-14
Change Feedback Type

### LINKED ISSUE #4708: Rating field of feedback should be float
**I'm submitting a ...**  
- [x] bug report
- [ ] feature request

**Current behavior:**
The `rating` field of `feedback` is set to `String` 

**Expected behavior:**
`Rating` field should be `float`

screen shot 2018-05-10 at 11 45 49 am

schedutron:
Fixes #4708 

#### Checklist

- I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- My branch is up-to-date with the Upstream `development` branch.

The feedback score should be a `float`, not a `string`. This PR changes the feedback type to `float` and updates the API docs accordingly.


schedutron:
I think I need to write/edit associated tests for this.

schedutron:
Travis build is successful, but Codecov shows red flag. What should I do?

schedutron:
@poush @simarsingh24 We can round it to the nearest 0.5 though, before storing, as `(rating * 2) -> round off -> divide the result by 2`. With this formula, `4.2` becomes `4.0` and `4.3` becomes `4.5`.

schedutron:
Should I implement that?

bhaveshAn:
I think the rating which a user gives should be an Integer. And we should have a average_rating attribute (additional) of type Float in the event model to describe the average rating of an event.

schedutron:
@bhaveshAn Average rating sounds great! @poush @simarsingh24 Should I implement that?

poush:
I will suggest to create new issue with concept of average rating. There are various ways to do it. for example creating the separate end point to get average, calculating  the average every time event detail api is called, or updating average column in event every time feedback is updated.
Let's not block this PR. Just change it to integer, create a new issue to finalise that.

schedutron:
@poush Okay, will create a separate issue and PR for that.

app/models/feedback.py
```diff
@@ -5,15 +5,15 @@ class Feedback(db.Model):
     """Feedback model class"""
     __tablename__ = 'feedback'
     id = db.Column(db.Integer, primary_key=True)
-    rating = db.Column(db.String, nullable=False)
+    rating = db.Column(db.Integer, nullable=False)
```
poush: I think this should have migration file

iamareebjamal:
Generally, all mobile UI widgets (iOS and Android) support 0.5 step for rating. We can round it easily to using 0.5 step

This will be a server limitation if only integers are allowed. Also, there should be a min-max limit of 0-5 on the field in marshmallow schema (preferably in db as well)

schedutron:
@iamareebjamal So should I revert to float and use the rounding-off formula mentioned before?

iamareebjamal:
This should also be added in the server. But let other maintainers weigh in

schedutron:
@iamareebjamal @poush @simarsingh24 So the rating field should remain of type `Float`?

iamareebjamal:
@bhaveshAn Yes

iamareebjamal:
- [ ] Round to 0.5 step
- [ ] Implement min max constraints

schedutron:
@iamareebjamal Should I implement the limits in the DB as well?

app/models/feedback.py
```diff
@@ -5,15 +5,20 @@ class Feedback(db.Model):
...
     user_id = db.Column(db.Integer,
                         db.ForeignKey('users.id', ondelete='CASCADE'))
     event_id = db.Column(db.Integer,
                          db.ForeignKey('events.id', ondelete='CASCADE'))
 
     def __init__(self, rating=None, comment=None, event_id=None, user_id=None):
-        self.rating = rating
+        rating = float(rating)
+        if not 0 < rating < 5:
+            raise ValueError, "Rating should be in range [0, 5]."
```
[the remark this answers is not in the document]
iamareebjamal: Use marshmallow fields and validators instead

iamareebjamal:
Yes, ideally they should be there as well

app/models/feedback.py
```diff
@@ -5,15 +5,20 @@ class Feedback(db.Model):
...
     event_id = db.Column(db.Integer,
                          db.ForeignKey('events.id', ondelete='CASCADE'))
 
     def __init__(self, rating=None, comment=None, event_id=None, user_id=None):
-        self.rating = rating
+        rating = float(rating)
+        if not 0 < rating < 5:
+            raise ValueError("Rating should be in range [0, 5].")
+        else:
+            self.rating = round(rating*2, 0) / 2  # Rounds to nearest 0.5
```
iamareebjamal: This is still manual checking rather than marshmallow

poush: @schedutron You need to add validations on feedback schema as well, which will allow throwing proper response based on these limits.

schedutron: @poush Do you intend the code to be added here: `open-event-server/app/api/schema/feedbacks.py`?

schedutron:
@iamareebjamal For implementing the limits in DB, I will have to create a custom domain as: 

`CREATE DOMAIN rating AS float CHECK (value BETWEEN 0 AND 5);`

Is this the right way to go? I may have to execute raw SQL, as SQLAlchemy perhaps doesn't provide API for custom domains.

app/models/feedback.py
iamareebjamal: Yes

app/api/schema/feedbacks.py
```diff
@@ -35,3 +38,10 @@ class Meta:
                         schema='UserSchema',
                         type_='user'
                         )
+
+    @validates_schema
+    def validate_rating(self, data):
+        """Method to validate rating"""
+        rating = float(data['rating'])
+        if 0 <= rating <= 5:
+            raise ValidationError("Rating should be in range [0, 5].")
```
iamareebjamal: There are already min max constraints in marshmallow, I recommend you going through marshmallow documentation

iamareebjamal:
> CREATE DOMAIN rating AS float CHECK (value BETWEEN 0 AND 5);

I guess then marshmallow validation will be enough

schedutron:
Travis build is passing, but I don't understand why Codacy is failing. Can anyone please explain it to me? :)

schedutron:
@poush How should we go about type-casting existing values?

iamareebjamal:
Yes. I think we should do best effort migration where we parse the string into float and save and catch the exception to save 0. That's the best we can do

schedutron:
Thanks @poush, @iamareebjamal. Can you please suggest some entry points as to where I do the type casting?

schedutron:
@poush @mayank8318 Thanks!

migrations/versions/194a5a2a44ef_.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def upgrade():
+    feedbacks_table = sa.sql.table('feedback',
+                                   sa.Column('rating', sa.VARCHAR())
+                                )
+    
+    op.execute(
+        feedbacks_table.update().values(
+            rating=str(update_rating(feedbacks_table.c.rating))
```
schedutron: This fails with the error
```
TypeError: <flask_script.commands.Command object at 0x103f520f0>: float() argument must be a string or a number, not 'Column'
```
How do I perform a self-correlated update here? What I wish to do is something like
```
UPDATE feedback SET rating = update_rating(rating);
```

migrations/versions/194a5a2a44ef_.py
```diff
@@ -14,8 +14,25 @@
 revision = '194a5a2a44ef'
 down_revision = '4cac94c86047'
 
+def update_rating(rating):
+    """Updates the old rating to new rating."""
+    rating = float(rating)
```
iamareebjamal: Not each value will be float, handle exception

migrations/versions/194a5a2a44ef_.py
poush: check this migration `versions/dbb9670cd902_.py`

schedutron:
@poush Thanks, it seems to be working now.
@iamareebjamal @bhaveshAn @srv-twry @mayank8318 Please review.

migrations/versions/194a5a2a44ef_.py
```diff
@@ -0,0 +1,64 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '194a5a2a44ef'
+down_revision = 'caf96244e10b'
+
+class ReplaceableObject(object):
```
bhaveshAn: @schedutron Was this class created automatically or hardcoded.

schedutron:
It is hardcoded.

schedutron:
Should I delete the procedure after the update completes?

poush:
@schedutron which procedure?

migrations/versions/194a5a2a44ef_.py
```diff
@@ -0,0 +1,62 @@
...
+    END;
+    $$
+    STRICT
+    LANGUAGE plpgsql IMMUTABLE;
+    """)
+
+
+def upgrade():
+    op.create_or_replace_sp(update_rating_func)
+    op.execute(UPDATE feedback SET rating=update_rating(rating)")
```
iamareebjamal: As reported, invalid syntax

iamareebjamal:
@schedutron Still the conflict has been resolved in wrong manner. I'd recommend that you squash your commits

schedutron:
I have rebased it.

iamareebjamal:
@schedutron Still squash your commits as it is best practice

schedutron:
@iamareebjamal Done.

PR #4758, opened 2018-05-21
Fix create_db where the parser is not assigned properly

### LINKED ISSUE #4757: Create_db is not setting arguments to a variable
- [x] bug report
- [ ] feature request

**Current behavior:**

Basically it's not correctly identifying command line arguments because the object was not set to another variable
**Expected behavior:**

you should be able to use the program by setting parameters on the terminal
**Steps to reproduce:**

run the program with command options
**Related code:**

https://github.com/fossasia/open-event-server/blob/70bb78a9204a322e80d8e647b7cf8c617176b30a/create_db.py#L41

Kreijstal:
Fixes #4757

#### Checklist

It turns out you had the assign the parser result to a variable and then use that object


create_db.py
```diff
@@ -38,9 +38,9 @@ def create_default_user(email, password):
     parser = argparse.ArgumentParser()
     parser.add_argument("email", nargs='?', help="The email for super_admin.", default='')
     parser.add_argument("password", nargs='?', help="The password for super_admin.", default='')
-    parser.parse_args()
-    email = parser.email if hasattr(parser, "email") else ''
-    password = parser.password if hasattr(parser, "password") else ''
+    parse = parser.parse_args()
+    email = parse.email if hasattr(parse, "email") else ''
```
schedutron: We can use the `getattr` BIF:

`email = getattr(parse, 'email', '')`

mayank8318: :+1:

poush:
@bhaveshAn https://github.com/kreijstal-contributions/open-event-api

create_db.py
poush: This can be simply replaced with `email = parser.get('email')` which will return None if not found.

schedutron: @poush If he doesn't want empty strings as default values for specific reasons, then `parser.get()` call seems to be the right way to go.

create_db.py
```diff
@@ -38,11 +38,9 @@ def create_default_user(email, password):
...
     parser.add_argument("password", nargs='?', help="The password for super_admin.", default='')
-    parser.parse_args()
-    email = parser.email if hasattr(parser, "email") else ''
-    password = parser.password if hasattr(parser, "password") else ''
+    parse = parser.parse_args()
     with current_app.app_context():
         db.create_all()
         stamp()
-        create_default_user(email, password)
+        create_default_user(getattr(parse, "email"), getattr(parse, "password"))
```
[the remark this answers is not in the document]
bhaveshAn: Agree

poush: It is better to use `parser.get('email')` here instead.
@schedutron Regarding `None`. Please check `create_default_user` method. The way it check will return same effect for empty string and None and None makes much sense than empty string.
Also, if you see the method, if the values are not provided it prompts the user for it so I don't think it is the good idea to provide default value unless we are intended to it.

Kreijstal: There is no parser.get

Kreijstal:
@poush 
When I try `parser.get` , I get 
```
Traceback (most recent call last):
  File "create_db.py", line 42, in <module>
    parser.get('email')
AttributeError: 'ArgumentParser' object has no attribute 'get'
```
When I try `parse.get` I get
```
Traceback (most recent call last):
  File "create_db.py", line 42, in <module>
    parse.get('email')
AttributeError: 'Namespace' object has no attribute 'get'
```

None of them work.

create_db.py
```diff
@@ -38,11 +38,9 @@ def create_default_user(email, password):
...
     parser.add_argument("password", nargs='?', help="The password for super_admin.", default='')
-    parser.parse_args()
-    email = parser.email if hasattr(parser, "email") else ''
-    password = parser.password if hasattr(parser, "password") else ''
+    parse = parser.parse_args()
     with current_app.app_context():
         db.create_all()
         stamp()
-        create_default_user(email, password)
+        create_default_user(getattr(parse, "email", ""), getattr(parse, "password", ""))
```
maxlorenz: Wouldn't it be better to use the Python 3 way of [argparse](https://docs.python.org/3/howto/argparse.html) here and use `if parse.email`? E.g.: 
```python
import argparse
parser = argparse.ArgumentParser()
parser.add_argument("--verbosity", help="increase output verbosity")
args = parser.parse_args()
if args.verbosity:
    print("verbosity turned on")
```

Kreijstal: I think what happens is that when there is no argument it comes up with error.

schedutron: It seems `getattr(object, attr, default_val)` is the right way to go.

poush:
So on reading the docs, argparse doesn't provide dict on calling `parser.parse_args()` therefore the `get()` won't work. We can pass the return value through `vars(parser.parse_args())` to get dict but we don't need it because we don't have to check for default values as they are already defined in argparse itself.
No need to check if attributed is defined and have some defined value as it will be always there with the default set in argparse.

Something like this has to be done 
```python
    parser = argparse.ArgumentParser()
    parser.add_argument("email", nargs='?', help="The email for super_admin.", default='')
    parser.add_argument("password", nargs='?', help="The password for super_admin.", default='')
    args = parser.parse_args()
    with current_app.app_context():
        db.create_all()
        stamp()
        create_default_user(args.email, args.password)
        populate()
```
I believe it should work but let me know if it doesn't.

Kreijstal:
@poush It doesn't work if the attribute is undefined  which is why I used the hasattr in the first place.

Kreijstal:
@poush through later inspection, it seems the issue was that I was not using the args variable, but I suppose it's okay now, take a look

create_db.py
```diff
@@ -38,11 +38,9 @@ def create_default_user(email, password):
     parser = argparse.ArgumentParser()
     parser.add_argument("email", nargs='?', help="The email for super_admin.", default='')
     parser.add_argument("password", nargs='?', help="The password for super_admin.", default='')
-    parser.parse_args()
-    email = parser.email if hasattr(parser, "email") else ''
-    password = parser.password if hasattr(parser, "password") else ''
+    parse = parser.parse_args()
```
schedutron: Agree, there can be a better name for `parser.parse_args()`.

bhaveshAn:
@Kreijstal Please wait for approval of atleast one Maintainer to merge your PR.

PR #4779, opened 2018-05-24
Fixes #4778 Feat: Add System Image for all the Event Topic (Category)

### LINKED ISSUE #4778: Add System Image for all the Event Topic (Category)

System Image is a default image for each Category of event, which will be set as event image in case the user doesn't provides it.
At frontend, this image should be visible at route http://localhost:4200/admin/content/system-images/1 for each event topic (category).

Allow just the admin to add this default image and visible to all the users.
Adding of system_image attribute in event model is required. Which would check if event_image is not provided by the event organizer then system image is used for event image.

**Additional context**

screenshot from 2018-05-24 23-02-14

**Wanna work on this issue**

bhaveshAn:
Fixes #4778 Feat: Add System Image for all the Event Topic (Category)

#### Checklist

Adds System Image feature for event topics.

- Add `system_image_url` field in event-topic model. and using this url to set default image of event in attribute `original_image_url` when `original_image_url` of event is not present in event creation.
-
-

@iamareebjamal @maxlorenz @poush @SaptakS Please review the code. Thanks!


app/models/event.py
```diff
@@ -208,7 +209,8 @@ def __init__(self,
         self.location_name = location_name
         self.description = clean_up_string(description)
         self.external_event_url = external_event_url
-        self.original_image_url = original_image_url
+        if self.original_image_url is None:
```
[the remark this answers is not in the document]
bhaveshAn: No, A user can provide it. In that case we won't be using system image of event topic.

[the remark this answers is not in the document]
srv-twry: What i meant is you forgot to set it.
You have to do something like this
```python
self.original_image_url = original_image_url
if self.original_image_url is None:
    // do stuff
```

app/models/event.py
```diff
@@ -262,6 +265,14 @@ def __setattr__(self, name, value):
         else:
             super(Event, self).__setattr__(name, value)
 
+    def set_default_event_image(self, event_topic_id):
+        if self.event_topic_id is None:
```
poush: What is the point of having a parameter when you are not using it? or using it some place. Better to remove the parameter from method and the property of the class.

bhaveshAn: How about 
```
def set_default_event_image(self, event_topic_id):
        if event_topic_id is None:
            return None
        else:
            event_topic = EventTopic.query.filter_by(
                event_topic_id=event_topic_id).first()
            return event_topic.system_image_url
```

bhaveshAn: I think, since we are validating an attribute before creating the object in __init__ method Constructor so cannot use `self.event_topic_id` in creation.

migrations/versions/20b836df74cc_.py
```diff
@@ -0,0 +1,40 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '20b836df74cc'
+down_revision = '4cac94c86047'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('microlocation')
```
niranjan94: @bhaveshAn why are we dropping microlocation here ?

niranjan94: @bhaveshAn why are we dropping microlocation here ?

Please check your migrations properly before pushing.

bhaveshAn: @niranjan94 Please review now.

bhaveshAn:
No, as I have edited just parameters in the parameters list of event-topics.

bhaveshAn:
@poush @SaptakS Please review !

app/models/event_topic.py
```diff
@@ -20,6 +20,7 @@ class EventTopic(db.Model):
 
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
     name = db.Column(db.String, nullable=False)
+    system_image_url = db.Column(db.String)
```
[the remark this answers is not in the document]
poush: @bhaveshAn if you add something on schema, it will be reflected on json response as well.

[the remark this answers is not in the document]
bhaveshAn: Yes, I tried adding system_image_url in EventTopicFactory and API Blueprint but the hooks were failing and so Travis Please see Build history (https://travis-ci.org/fossasia/open-event-server/builds/383386054). So Can we deal it in some other PR ?

bhaveshAn:
@poush Multiple Migration heads error is there. Can you please help to rectify it.
Last night Travis has passed when Niranjan and Areeb had approved. Now don't know what happened.

poush:
run `python manage.py db heads` to see the heads. You can merge them by `python manage.py db merge [HEAD1] [HEAD2] ....`

bhaveshAn:
@poush 
Resolved Migration head error. Now Travis failing due to API blueprint. 
Also, I have added the attr system_image_url as required one so should I add it at EventTopicFactory as well. Did it earlier but cannot get which Event Topic APIs has to be modified in API blueprints.
Please tell .

iamareebjamal:
If you added it as required, what is used as default value for pre-existing rows?

app/models/event_topic.py
```diff
@@ -20,6 +20,8 @@ class EventTopic(db.Model):
 
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
     name = db.Column(db.String, nullable=False)
+    system_image_url = db.Column(db.String, nullable=False,
+        default='https://www.w3schools.com/html/pic_mountain.jpg')
```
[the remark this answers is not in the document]
iamareebjamal: Unfortunately, although this is default image used in many places in the project, in my opinion, a more suitable contextual creative commons free icon should be used instead

bhaveshAn:
@poush @iamareebjamal @niranjan94 Ready for review ! Please review.

bhaveshAn:
@poush @iamareebjamal @niranjan94 Earlier added commits due to which your approvals have been dismissed ! Please review/approve again . Travis passing though

bhaveshAn:
@iamareebjamal @niranjan94 Earlier added commits due to which your approvals have been dismissed ! Please review/approve again .

app/models/event_topic.py
```diff
@@ -20,6 +20,10 @@ class EventTopic(db.Model):
 
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
     name = db.Column(db.String, nullable=False)
+    system_image_url = db.Column(
+        db.String, nullable=False,
+        default='http://www.cih.org/resources/images/event_header.jpg'
```
iamareebjamal: This resource should be generic like the empty avatar when you create a new account

bhaveshAn: @iamareebjamal 
Do you mean this 
`default=''`

iamareebjamal: No I mean for avatar images like this are used - http://hotchillitri.co.uk/wp-content/uploads/2016/10/empty-avatar.jpg

Similarly such generic image should be used with appropriate creative commons license

bhaveshAn: How will this work http://vitriviera.com/static/media/pre.17254662.jpg

iamareebjamal: It should be generic. Remember how you use a service and expect it to have sane defaults for you

Example - https://github.com/fossasia/open-event-orga-app/blob/development/app/src/main/res/drawable/header.png

Kreijstal:
Which image do we choose though?

app/models/event_topic.py
maxlorenz: and please download the image. Otherwise the project depends on an external URL that might change

iamareebjamal:
And use the image provided by me as we have the rights to use that image

app/models/event_topic.py
bhaveshAn: I have uploaded that image at https://image.ibb.co/iTrd5y/system_image_common.png 
Which URL will not change 
Will it be okay @maxlorenz @iamareebjamal

iamareebjamal: This is based on external service which can be down without any notice

app/models/event_topic.py
```diff
@@ -22,17 +22,22 @@ class EventTopic(db.Model):
     name = db.Column(db.String, nullable=False)
     system_image_url = db.Column(
         db.String, nullable=False,
-        default='https://image.ibb.co/iTrd5y/system_image_common.png'
+        default='http://localhost:5000/static/header.png'
```
iamareebjamal: Not really good. If it is hosted on heroku, this should not point to localhost

SaptakS: @bhaveshAn can you maybe use the `SERVER_NAME` or use something like `url_for`

app/models/event_topic.py
```diff
@@ -13,22 +14,34 @@ def get_new_slug(name):
         return '{}-{}'.format(slug, uuid.uuid4().hex)
 
 
+default_image = current_app.config.get('SERVER_NAME') + '/static/header.png'
```
bhaveshAn: @SaptakS @iamareebjamal Can you help with this.
https://travis-ci.org/fossasia/open-event-server/builds/388320978

iamareebjamal: You cannot access `app` outside of context

bhaveshAn: Please tell how to deal as this PR is open from very long.
How to modify it.
`default_image = current_app.config.get('SERVER_NAME') + '/static/header.png' `

iamareebjamal: Use this inside a method as not as a static constant

poush:
First fix your merge heads. Either jump the down version of your migration file to the conflicting head or merge your heads. Also, try use `app.settings`

srv-twry:
@bhaveshAn Please resolve conflicts.

bhaveshAn:
@iamareebjamal @poush Please help to fix the travis build. tried a lot but not able to fix that

srv-twry:
@bhaveshAn The problem is due to multiple migration heads.

migrations/versions/e1aee9f7409b_.py
```diff
@@ -0,0 +1,20 @@
...
+revision = 'e1aee9f7409b'
+down_revision = ('194a5a2a44ef', '63fb9eb9a2bf')
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
+    pass
```
srv-twry: Delete this file

migrations/versions/63fb9eb9a2bf_.py
```diff
@@ -0,0 +1,20 @@
...
+revision = '63fb9eb9a2bf'
+down_revision = ('20b836df74cc', 'caf96244e10b')
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
+    pass
```
srv-twry: Delete this file

migrations/versions/355d1d2992be_.py
```diff
@@ -0,0 +1,24 @@
...
+revision = '355d1d2992be'
+down_revision = ('e1aee9f7409b', '4bbba854e30b')
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
+    pass
```
srv-twry: Delete this file

migrations/versions/20b836df74cc_.py
```diff
@@ -0,0 +1,30 @@
...
+Create Date: 2018-05-25 01:46:53.292127
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from flask import current_app
+
+# revision identifiers, used by Alembic.
+revision = '20b836df74cc'
+down_revision = '4cac94c86047'
```
srv-twry: Change it to : `ebf41490ac7d`

migrations/versions/20b836df74cc_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message
+Revision ID: 20b836df74cc
+Revises: 4cac94c86047
```
srv-twry: Change it to `ebf41490ac7d`

srv-twry:
Explainations for the requested changes:
When you started working on this PR, the db head was at `4cac94c86047`. So your migrations were from `4cac94c86047` to `20b836df74cc` . But when your PR was being reviewed, other PRs were merged which meant the new db head was not `4cac94c86047` any more. As of now, it is at `ebf41490ac7d` , so you need to update your migration file to `ebf41490ac7d` i.e change it's `down_revision` to `ebf41490ac7d`. Also delete the other ones since they don't do anything.

.env.example
```diff
@@ -1,4 +1,5 @@
 DATABASE_URL=postgresql://john:start@127.0.0.1:5432/oevent
+SERVER_URL=http://localhost:5000
```
iamareebjamal: This created a lot of problems for us in the past

srv-twry:
@bhaveshAn Now the build fails purely due to programming reasons. You can fix it now.

.env.example
bhaveshAn: But for accessing the default system image (`header.png` as suggested by you). I'll require this

iamareebjamal: https://stackoverflow.com/questions/15974730/how-do-i-get-the-different-parts-of-a-flask-requests-url

bhaveshAn: `flask.request.url_root` is not letting pass the travis build.
Failed due to that : https://travis-ci.org/bhaveshAn/open-event-server/builds/395595002
Passing if using localhost:5000 : https://travis-ci.org/fossasia/open-event-server/builds/395581727

iamareebjamal: This is not some global module request from flask. This is the request coming to the server. Did you test the implementation?

bhaveshAn: @niranjan94 Please help to fix travis .

bhaveshAn:
@iamareebjamal Please review

app/models/event_topic.py
```diff
@@ -20,20 +22,31 @@ class EventTopic(SoftDeletionModel):
...
+            if system_image_url is None else system_image_url
         self.slug = get_new_slug(name=self.name)
         self.deleted_at = deleted_at
 
+    @classmethod
+    def set_default_sys_image(cls):
+        with app.test_request_context():
+            url = urllib.parse.urlparse(request.url)
+            image_url = url.scheme + '://' + url.netloc + '/static/header.png'
+            return image_url
```
iamareebjamal: This is test_request context

Pass real request here

bhaveshAn: app.request_context() requires argument `environ`
```
with app.request_context():
            url = urllib.parse.urlparse(request.url)
            image_url = url.scheme + '://' + url.netloc + '/static/header.png'
            return image_url
```
Even it is test_request_context() it works fine. Also, RequestContexttask using it  https://github.com/fossasia/open-event-server/blob/development/app/api/helpers/request_context_task.py#LC41

iamareebjamal: Celery tasks are separate things because they are detached from actual requests, here you are sending response from a request, why can't you simply take that request and pass it here?

bhaveshAn: without using `with app.test_request_context():` it's not working.

bhaveshAn: without `test_request_context` error comes
```
    raise RuntimeError(_request_ctx_err_msg)
RuntimeError: Working outside of request context.

This typically means that you attempted to use functionality that needed
an active HTTP request.  Consult the documentation on testing for
information about how to avoid this problem.
```

bhaveshAn:
As this PR is getting late due to accessing the local image for system image url, so Now it just adds a system_image_url column in Event Topic Model. Later I will add the implementation of accessing the local image 'header.png' and updation of API blueprints also in a single PR.

iamareebjamal:
@bhaveshAn I'm sorry but just merging your own PR because it is delayed is not ideal. The solution is half-baked and there are already so many half-baked merges in the server which make it unusable to consume the API for clients

The issue of displaying URLs to the consumers is extremely crucial and very trivial, even showing relative URL would have sufficed, or if the complete URL was to be shown, it could be done by just extracting the URL from request, but you kept accessing URL from outside the app context rather than from the controller

bhaveshAn:
>  but you kept accessing URL from outside the app context rather than from the controller

@iamareebjamal I did that too but it was just returning `system_image_url = "http://localhost/static/header.png" ` and misses the port even when i added ` url.hostname +  url.port` and accessed it in request context.

bhaveshAn:
@iamareebjamal I think the reason for this could be 
since we create event topics in populate_db.py so when make fresh db and run create_db.py at that time we are not running server so it missed the port. in command `python create_db.py` what do you think on this.

iamareebjamal:
```Python
from flask import Flask, request
app = Flask(__name__)

@app.route("/gopher/nested/url")
def hello():
    return "The root URL is " + request.url_root
```

I'm getting complete URL with protocol, host name, and port

srv-twry:
@bhaveshAn Your migration file isn't right either. May be you made changes after creating the migrations file but never updated it.

When I ran `python manage.py db migrate`, it added 
```
op.alter_column('event_topics', 'name',
                    existing_type=sa.VARCHAR(),
                    nullable=True)
```
to my migrations file but I never touched this model.

srv-twry:
Please open a PR to fix it.

bhaveshAn:
```from flask import Flask, request
app = Flask(__name__)

@app.route("/gopher/nested/url")
def hello():
    return "The root URL is " + request.url_root
```
> I'm getting complete URL with protocol, host name, and port

@iamareebjamal Yes, even I can get the complete URL using this example, but the point here is we can get the complete url after running the server and not when we run or populate db ([creating records of event-topics](https://github.com/fossasia/open-event-server/blob/development/populate_db.py#LC125))using create_db.py for the first time. Getting URL using routes in this way makes sense only when we run the server but while we populate db we are actually not running server. So since we are not running server in step `python create_db.py` then no point that we are receiving a request.

**At the time of `db.create_all()` we have no access to any request or route of our current App**
**After running `python manage.py runserver` we will be getting request and can access the routes**

iamareebjamal:
You're still not understanding the strategy. You have to store relative paths in database, not full paths. You never store full paths in any database. This issue of URL is not related to database at all

PR #4823, opened 2018-05-30
fix: server not returning hidden tickets

### LINKED ISSUE #4822: Server does not return hidden tickets
**Describe the bug**
Currently server is returning only those tickets whose `is-hidden` property is false. Whereas server should return all the tickets and frontend should be ready to display them to organizer and not to public. So in current case event organiser is not able to access/edit the ticket that he made hidden during a edit.

**To Reproduce**
1. Create a event and have two tickets one hidden and other one as public.
2. Now save the event.
3. Now go back to edit the event and ticket details.
4. You see that ticket you made hidden is not listed.

**Expected behavior**
Even hidden tickets should be available to organiser to edit.

pradeepgangwar:
Fixes #4822 

#### Checklist

Currently server is returning only those tickets whose is-hidden property is false. In this way organizer is also not able to edit the tickets at the edit route. Server should be returning all the tickets for admin and organizer. Also it should return only visible tickets for unauthenticated and normal users.

- If a user is logged in server returns all the tickets if logged in user is super user or admin or organizer of the event.
- If logged in user is normal user then it send all the tickets except hidden one.
- If query is done without logging in then it sends all the tickets except hidden one.

### DIFF
--- a/app/api/tickets.py
+++ b/app/api/tickets.py
@@ -1,5 +1,7 @@
+from flask import request, current_app
 from flask_rest_jsonapi import ResourceDetail, ResourceList, ResourceRelationship
 from flask_rest_jsonapi.exceptions import ObjectNotFound
+from flask_jwt import current_identity as current_user, _jwt_required
 
 from app.api.bootstrap import api
 from app.api.helpers.db import safe_query
@@ -66,7 +68,18 @@ class TicketList(ResourceList):
         :param view_kwargs:
         :return:
         """
-        query_ = self.session.query(Ticket).filter_by(is_hidden=False)
+
+        if 'Authorization' in request.headers:
+            _jwt_required(current_app.config['JWT_DEFAULT_REALM'])
+            if current_user.is_super_admin or current_user.is_admin:
+                query_ = self.session.query(Ticket)
+            elif view_kwargs.get('event_id') and has_access('is_organizer', event_id=view_kwargs['event_id']):
+                query_ = self.session.query(Ticket)
+            else:
+                query_ = self.session.query(Ticket).filter_by(is_hidden=False)
+        else:
+            query_ = self.session.query(Ticket).filter_by(is_hidden=False)
+
         if view_kwargs.get('ticket_tag_id'):
             ticket_tag = safe_query(self, TicketTag, 'id', view_kwargs['ticket_tag_id'], 'ticket_tag_id')
             query_ = query_.join(ticket_tags_table).filter_by(ticket_tag_id=ticket_tag.id)


Kreijstal:
What srv-trw says makes sense, make it so, so that it returns hidden events when you're an admin.

pradeepgangwar:
Ok, I will try to do it this way. Thanks @srv-twry :+1:

pradeepgangwar:
@srv-twry @Kreijstal One more doubt. There are 2-3 places where tickets are listed. The one visible to audience will be on the public (/e/<event_id>) route. So if organiser or admin himself wants to preview the event then he/she will still see all the tickets, whereas since he is in preview mode(/e/<event_id>) he should see actual thing that will be visible to public.
Anything we can do here?

srv-twry:
One solution could be to remove the JWT token of the logged in organiser i.e. don't send any token in the request when fetching public data for the event. It will then be shown the public schema. Is it possible ?

app/api/tickets.py
```diff
@@ -37,9 +38,6 @@ def before_post(self, args, kwargs, data):
         if get_count(db.session.query(Ticket.id).filter_by(name=data['name'], event_id=int(data['event']))) > 0:
             raise ConflictException({'pointer': '/data/attributes/name'}, "Ticket already exists")
 
-        if get_count(db.session.query(Event).filter_by(id=int(data['event']), is_ticketing_enabled=False)) > 0:
-            raise MethodNotAllowed({'parameter': 'event_id'}, "Ticketing is disabled for this Event")
-
```
srv-twry: Why delete this ?

app/api/tickets.py
```diff
@@ -67,6 +65,14 @@ def query(self, view_kwargs):
         :return:
         """
         query_ = self.session.query(Ticket).filter_by(is_hidden=False)
+
+        if 'Authorization' in request.headers:
+            _jwt_required(current_app.config['JWT_DEFAULT_REALM'])
+            if current_user.is_super_admin or \
+               current_user.is_admin or \
+               has_access('is_organizer', event_id=view_kwargs['event_id']):
+                    query_ = self.session.query(Ticket).filter_by()
```
srv-twry: Else statements also needed here.

app/api/tickets.py
```diff
@@ -67,6 +69,14 @@ def query(self, view_kwargs):
         :return:
         """
         query_ = self.session.query(Ticket).filter_by(is_hidden=False)
```
pradeepgangwar: @srv-twry I included this as default behavior. Do i still need else statement after if in code below?

srv-twry: I think the issue in this approach is that you will querying twice when the user is the organizer or the admin. 
first the default one and then in case when the `if` condition is executed. 
I would suggest that you `else` conditions. The code would look slightly more complicated but it won't query the db twice. Views ?

pradeepgangwar: Yes, that's right. Let me make changes.

pradeepgangwar:
@srv-twry @Kreijstal @bhaveshAn Please review.

app/api/tickets.py
```diff
@@ -66,7 +67,19 @@ def query(self, view_kwargs):
         :param view_kwargs:
         :return:
         """
-        query_ = self.session.query(Ticket).filter_by(is_hidden=False)
+        
+        if 'Authorization' in request.headers:
+            _jwt_required(current_app.config['JWT_DEFAULT_REALM'])
+            if current_user.is_super_admin or current_user.is_admin:
+                query_ = self.session.query(Ticket).filter_by()
+            else if view_kwargs.get('event_id'):
```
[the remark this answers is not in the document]
srv-twry: `elif`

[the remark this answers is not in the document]
pradeepgangwar: fixed :sweat_smile: . Just too much javascript is dangerous.

pradeepgangwar:
@srv-twry Please check. I think it should not block now when querying with access code

app/api/tickets.py
```diff
@@ -66,7 +68,19 @@ def query(self, view_kwargs):
...
         :return:
         """
-        query_ = self.session.query(Ticket).filter_by(is_hidden=False)
+
+        if 'Authorization' in request.headers:
+            _jwt_required(current_app.config['JWT_DEFAULT_REALM'])
+            if current_user.is_super_admin or current_user.is_admin:
+                query_ = self.session.query(Ticket).filter_by()
+            elif view_kwargs.get('event_id'):
+                if has_access('is_organizer', event_id=view_kwargs['event_id']):
```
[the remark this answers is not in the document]
pradeepgangwar: Yes that's right. Done.

app/api/tickets.py
```diff
@@ -66,7 +68,18 @@ def query(self, view_kwargs):
         :param view_kwargs:
         :return:
         """
-        query_ = self.session.query(Ticket).filter_by(is_hidden=False)
+
+        if 'Authorization' in request.headers:
+            _jwt_required(current_app.config['JWT_DEFAULT_REALM'])
+            if current_user.is_super_admin or current_user.is_admin:
+                query_ = self.session.query(Ticket).filter_by()
```
srv-twry: you can remove the `filter_by()`. Sorry noticed it just now :sweat_smile:

pradeepgangwar: Removed it. Thanks :smile:

PR #4851, opened 2018-06-04
WIP: Enforce one-to-one relationship between ticket holders and tickets

### LINKED ISSUE #4835: Multiple ticket_holders can be mapped to the same ticket
**Describe the bug**
Server allows multiple ticket holders to have the same ticket. However, only one ticket holder should be mapped to a given ticket.

**To Reproduce**
1. Create a ticket (`POST`)
2. Create multiple attendees all having the ticket id of the above ticket (`POST`)

**Expected behavior**
Server should validate that each ticket corresponds to *one and only one* attendee.

I'm working on fixing this.

schedutron:
Fixes #4835

#### Checklist

There is no constraint that prevents multiple ticket_holders to attend on the same ticket. This PR tries to implement this constraint.


schedutron:
Shouldn't the various `id`s be integers?

https://github.com/schedutron/open-event-server/blob/c5495780ae6fe15e8edfb97d76a2a94b524efbc2/app/api/schema/attendees.py#L46

app/api/attendees.py
```diff
@@ -31,11 +35,26 @@ def before_post(self, args, kwargs, data):
...
+                )
         if ticket.event_id != int(data['event']):
-            raise UnprocessableEntity({'pointer': '/data/relationships/ticket'}, "Ticket belongs to a different Event")
+            raise UnprocessableEntity(
+                {'pointer': '/data/relationships/ticket'},
+                "Ticket belongs to a different Event"
+            )
+
+        if db.session.query(TicketHolder.id).filter_by(
+                ticket_id=int(data['ticket']), event_id=int(data['event'])
```
srv-twry: Checking only for the `ticket_id` should be sufficient. If the `ticket_id` already exists, then that already means that we have someone on this ticket. 
There won't be a case when the `ticket_id` is same but `event_id` is different. Hence only filtering by `ticket_id` is sufficient IMO. 

Thoughts @bhaveshAn @Kreijstal @mayank8318

schedutron: I was suspecting that. Let's see what others have to say in this.

mayank8318: I agree with @srv-twry

app/api/attendees.py
```diff
@@ -31,11 +35,26 @@ def before_post(self, args, kwargs, data):
...
         require_relationship(['ticket', 'event'], data)
 
-        ticket = db.session.query(Ticket).filter_by(id=int(data['ticket'])).first()
+        ticket = db.session.query(Ticket).filter_by(
+            id=int(data['ticket'])
+            ).first()
         if ticket is None:
-            raise UnprocessableEntity({'pointer': '/data/relationships/ticket'}, "Invalid Ticket")
+            raise UnprocessableEntity(
+                {'pointer': '/data/relationships/ticket'}, "Invalid Ticket"
```
Kreijstal: it's not that the ticket is invalid but more like the ticket was not found, isn't it?

app/api/attendees.py
Kreijstal: I don't think it matters much, since it will pretty much bring the same result (unless something extraordinary happens), but if it improves redability to just filter by ticket, then be my guest.

app/api/attendees.py
schedutron: Yeah, I'll fix that. Thoughts @srv-twry @mayank8318 @iamareebjamal?

srv-twry: yeah do it. :+1:

migrations/versions/9d21de792967_.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '9d21de792967'
+down_revision = '194a5a2a44ef'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint('ticket_event', 'ticket_holders', ['ticket_id', 'event_id'])
```
maxlorenz: Does this migrate well? I mean when one has a non-unique relationship in the db? Or does the migration crash then?

schedutron: @maxlorenz It crashes. I tested it locally and it crashed. How should we handle that?

iamareebjamal: Delete all except one

schedutron: @iamareebjamal How do we decide the one to keep? The earliest one? The one with lowest `id`?

iamareebjamal: Yes, the first one should be kept

schedutron:
@iamareebjamal @srv-twry @Kreijstal please review :)

app/api/attendees.py
srv-twry: I noticed it just now. IMO it would raise NoResultFound exception and return HTTP 500  when ticket doesn't exist. Please test it :)

app/api/attendees.py
```diff
@@ -31,11 +35,26 @@ def before_post(self, args, kwargs, data):
         """
         require_relationship(['ticket', 'event'], data)
 
-        ticket = db.session.query(Ticket).filter_by(id=int(data['ticket'])).first()
+        ticket = db.session.query(Ticket).filter_by(
+            id=int(data['ticket'])
+            ).first()
```
srv-twry: Let's wrap this in try-catch block and except a NoResultFound error. Please test it in the case of a random ticket id which doesn't exist.

schedutron: I tried it with a random ticket and 500 wasn't returned. Instead, I got this:
```
{
    "errors": [
        {
            "status": 422,
            "source": {
                "pointer": "/data/relationships/ticket"
            },
            "title": "Unprocessable Entity",
            "detail": "Invalid Ticket"
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}
```
So I guess no need to wrap?

srv-twry: alright then. If it works then fine. :smile:

schedutron:
@niranjan94 @SaptakS @bhaveshAn Please review :)

srv-twry:
@schedutron Please squash the commits

schedutron:
@srv-twry Will do once a mentor approves this as well, otherwise I may need to add more commits after squashing :). Do you think the PR should be accompanied by related tests?

srv-twry:
It would be great if you add tests. But I don't see it as a compulsory step in order to merge this.

iamareebjamal:
Tests are good and should be added with PRs but when PRs take long to merge, tests become additional blockers

schedutron:
@srv-twry @iamareebjamal Thanks, I'll now squash the commits and add tests in a separate PR so that they don't hinder the merge. Tests may seem as blockers in the short term, but they're a great asset in the long run.

iamareebjamal:
@schedutron Yes, I agree. I'm the kind of person who pushes to write tests for even the most mediocre of methods. I had also created an issue to fix the testability of the codebase as in its current form, it is not possible to unit test it

PR #4869, opened 2018-06-08
Append "s" at the end of stripe api endpoint url

### LINKED ISSUE #4868: Change Stripe api endpoint to include s at the end of url
**Describe the bug**
Currently ember by default makes calls to stripe-authorizations whereas server has end point as stripe-authorization

**Expected behavior**
-  All the other endpoints use "s" in the end i.e. the plural sense. There should be consistency.
-  FE is in Ember which follows conventions. If we don't add "s" to it then we will have to unnecessary modifications in the adapter for the FE.

Hence better to follow conventions. It will make life easier for us in the FE.

pradeepgangwar:
Fixes #4868 

#### Checklist

Currently ember by default makes calls to stripe-authorizations whereas server has end point as stripe-authorization. All the other endpoints use "s" in the end i.e. the plural sense. There should be consistency. Ember follows conventions hence this should be added to make frontend easier to implement. This PR appends s at the end of url of stripe api. 

- Append "s" to the url of stripe API.

### DIFF
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -171,7 +171,7 @@ api.route(EventDetail, 'event_detail', '/events/<int:id>', '/events/<identifier>
           '/access-codes/<int:access_code_id>/event', '/email-notifications/<int:email_notification_id>/event',
           '/attendees/<int:attendee_id>/event', '/custom-forms/<int:custom_form_id>/event',
           '/orders/<order_identifier>/event', '/faqs/<int:faq_id>/event', '/faq-types/<int:faq_type_id>/event',
-          '/feedbacks/<int:feedback_id>/event', '/stripe-authorization/<int:stripe_authorization_id>/event')
+          '/feedbacks/<int:feedback_id>/event', '/stripe-authorizations/<int:stripe_authorization_id>/event')
 api.route(EventRelationship, 'event_ticket', '/events/<int:id>/relationships/tickets',
           '/events/<identifier>/relationships/tickets')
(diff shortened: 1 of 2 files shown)


app/api/__init__.py
```diff
@@ -463,11 +463,11 @@
           '/feedbacks/<int:id>/relationships/user')
 
 # Stripe Authorization API
-api.route(StripeAuthorizationListPost, 'stripe_authorization_list_post', '/stripe-authorization')
-api.route(StripeAuthorizationDetail, 'stripe_authorization_detail',  '/stripe-authorization/<int:id>',
-          '/events/<int:event_id>/stripe-authorization', '/events/<event_identifier>/stripe-authorization')
+api.route(StripeAuthorizationListPost, 'stripe_authorization_list_post', '/stripe-authorizations')
+api.route(StripeAuthorizationDetail, 'stripe_authorization_detail',  '/stripe-authorizations/<int:id>',
+          '/events/<int:event_id>/stripe-authorizations', '/events/<event_identifier>/stripe-authorizations')
```
srv-twry: Please don't change the url for the event relationship.
It should still be `/events/<int:event_id>/stripe-authorization`. You can use event copyright as a reference :smile:

srv-twry:
> The weird thing would be that the server would just return always one, wouldn't it?

@Kreijstal No it's not weird. It won't always return one. For eg. The admin can get the whole list of stripe authorization objects. It's the convention that is used everywhere in the app and I wasn't aware of that when I was implementing stripe authorization endpoint.

PR #4908, opened 2018-06-15
feat: Add soft deletion support in user and event related schemas and models

### LINKED ISSUE #4879: Add deleted_at field to every user & event related model and schema

Since we are using soft deletion by default, we should add field `deleted_at` to all the schemas and models which are directly related to event or user.

Each and every resource that supports deletion should have a delete_at field. 

**Additional context**

Required to fix #4874 

CC: @niranjan94 @shubham-padia @SaptakS

srv-twry:
Fixes #4879 

#### Checklist

In order to add soft deletion support for Event and Users, we need to add a column "deleted_at" in each of the related schemas.

Added a SoftDeletionSchema and SoftDeletionModel in order to add the field.


srv-twry:
The docs will be updated in a PR following this one. The PR is already long.

srv-twry:
**Session**
sessions

**Speakers**
track

It's not possible to show screenshot for each and every model/schema defined :)

app/models/base.py
```diff
@@ -0,0 +1,10 @@
+from app.models import db
+
+
+class SoftDeletionModel(db.Model):
```
schedutron: The motive behind inheriting from this model is to avoid adding `deleted_at` attributes for all the models that support soft deletion, right? Are there other advantages of having a whole new model just for soft deletion?

srv-twry: Yep, just so that we don't repeat code. There won't be a database table associated with this though.

Kreijstal: maybe you can as well abstract it even further say, using a common open-event schema, but maybe in the future.

app/models/event_invoice.py
```diff
@@ -97,6 +99,7 @@ def __init__(self,
         self.last4 = last4
         self.stripe_token = stripe_token
         self.paypal_token = paypal_token
+        self.deleted_at = deleted_at
```
Kreijstal: But why are you adding this property when it inherits from SoftDeletionModel?

srv-twry: I am not adding it. It is already declared from the base SoftDeletionModel, I am initialising it. Am I wrong in this ? @schedutron @mayank8318

srv-twry:
@mayank8318 @schedutron @Kreijstal Please review. It's a very simple PR to review.

srv-twry:
@maxlorenz Actually it was implemented like this before. I am just adding it to a few more tables. Check this blog post: https://blog.fossasia.org/handling-soft-and-hard-deletes-in-the-open-event-server-api/

But I do agree with you on this that this. It doesn't seem to be the right way to do it. May be we could discuss with the mentors on this in the channel.

iamareebjamal:
I think using a professional entity versioning system would be better for long term 

http://sqlalchemy-continuum.readthedocs.io/en/latest/intro.html

SaptakS:
@iamareebjamal I think a proper versioning system will definitely help in making the entire thing better to not soft delete but to create version object on both update and delete. But apart from that, I would still have soft delete at db level to allow the admins to have a quick recovery from some messy delete. What do you think?

PR #4958, opened 2018-06-23
fix: Update queries to handle soft-deletion

### LINKED ISSUE #4946: Update query statements to handle soft deletion case throughout the project

After the addition of the `deleted_at` column in every user and event related model/schema in #4879 , we must now ensure that all the query statements inside the project take care of this.

**Reason**
Let's say we have an event. Now we create a tax for that event. Since tax is also an event related field,so it will also have a deleted_at column.
Now for some reason, I decide to delete the tax. Since by default we have soft deletion enabled. The tax would be soft deleted but the entry would still be in the database.
Now let's say I i.e. the event organizer wants to add tax again. So I will make another post request. Now before creating a tax object for an event, we check if a tax object already exists in the db for that event or not like this :
```Python
if self.session.query(Tax).filter_by(event_id=data['event']).first():
            raise ConflictException({'pointer': '/data/relationships/event'}, "Tax already exists for this event")
```
Now even though the tax entry was deleted by the user, this snippet would still raise ConflictException since that deleted entry is still present in the db. 

Make all the queries inside the project take care of this case. i.e. replace it with something like this:
```Python
if self.session.query(Tax).filter_by(event_id=data['event'] and deleted_at=None).first():
            raise ConflictException({'pointer': '/data/relationships/event'}, "Tax already exists for this event")
```

srv-twry:
Fixes #4946 

#### Checklist

After the addition of the deleted_at column in every user and event related model/schema, we must now ensure that all the query statements inside the project take care of this. Please check issue for a detailed explaination and example.

- Adds SoftDeletion support to SpeakerCalls, Tax and Event copyright.
- Updates queries to take care of soft deletion.


srv-twry:
@Kreijstal @mayank8318 @bhaveshAn Please review :)

app/api/attendees.py
```diff
@@ -49,7 +49,7 @@ def before_post(self, args, kwargs, data):
             )
 
         if db.session.query(TicketHolder.id).filter_by(
-                ticket_id=int(data['ticket']), event_id=int(data['event'])
+                ticket_id=int(data['ticket']), event_id=int(data['event']), deleted_at=None
```
bhaveshAn: Why is the need to filter the queries by `deleted_at=None` as we are already doing 
https://github.com/fossasia/flask-rest-jsonapi/commit/7b1b0865b5b2ac016333a33ab1999419a2bb13d1#diff-1c38b04e7b5726a5388ec960bb8a4bca **get_object**
https://github.com/fossasia/flask-rest-jsonapi/commit/066a050e1dcda884ecbaaa614e37fb341a1a3ef7#diff-1c38b04e7b5726a5388ec960bb8a4bca **get_collection**

srv-twry: From the linked issue:
> Let's say we have an event. Now we create a tax for that event. Since tax is also an event related field,so it will also have a deleted_at column.
Now for some reason, I decide to delete the tax. Since by default we have soft deletion enabled. The tax would be soft deleted but the entry would still be in the database.
Now let's say I i.e. the event organizer wants to add tax again. So I will make another post request. Now before creating a tax object for an event, we check if a tax object already exists in the db for that event or not like this :
```
if self.session.query(Tax).filter_by(event_id=data['event']).first():
            raise ConflictException({'pointer': '/data/relationships/event'}, "Tax already exists for this event")
```
Now even though the tax entry was deleted by the user, this snippet would still raise ConflictException since that deleted entry is still present in the db.

srv-twry: The library filters the `get` requests. All the changes that I have made in this PR correspond to the POST requests.

PR #4985, opened 2018-06-29
Add endpoint to resend email confirmation

### LINKED ISSUE #4968: Add an end point for resending the verification mail to the user.

Currently, when the user registers, a confirmation mail is sent out. However if the link expires or the mail gets deleted, there should be an option for the user to make the server resend a verification mail.

image
Currently the link on the FE doesn't work, because there is no proper support for that.

schedutron:
Fixes #4968

#### Checklist

This PR adds endpoint to resend email confirmation in circumstances where it is required - e.g, when the original confirmation email is deleted by mistake.

Added an endpoint `/resend-confirmation-email` for re-sending email confirmation for verifying a new account.


app/api/auth.py
```diff
@@ -38,6 +40,27 @@ def verify_email():
...
+def resend_verification_email():
+    try:
+        email = request.json['data']['email']
+    except TypeError:
+        return BadRequestError({'source': ''}, 'Bad Request Error').respond()
+
+    try:
+        user = User.query.filter_by(email=email).one()
+    except NoResultFound:
+        return UnprocessableEntityError({'source': ''}, 'User not found').respond()
```
srv-twry: `User with email: {email} not found`.

app/api/auth.py
```diff
@@ -38,6 +40,27 @@ def verify_email():
...
+    else:
+        s = get_serializer()
```
CosmicCoder96: `s` is not a good variable name.

schedutron:
I will add the docs, but before that, I need to configure sender's email locally and test the actual API response, as I mentioned in the Gitter channel.

mayank8318:
@schedutron How did you test this locally?

schedutron:
@mayank8318 I'm trying to do that locally by configuring the sender's email. This blog post was useful: https://blog.fossasia.org/implement-email-in-open-event-server/

poush:
@schedutron Looks good. Have you been able to run it locally? The email goes well and the link works? Also, you will need to add this in API blueprint as well.

schedutron:
I've tried to fix the email problem, and the following endpoint now works:

screen shot 2018-07-08 at 11 27 17 am
I got the following mail (but it got classified as spam):
screen shot 2018-07-08 at 11 28 22 am

However, there is a problem: when I was testing the Sendgrid API in the interpreter, it was working fine. But within the server's celery context, it was giving an annoying bug:
```
SSLError(SSLError("bad handshake: Error([('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')],)",),))
```

So I had to pass a `verify=False` flag to disable SSL verification. It's weird that it only happens within the celery context.

schedutron:
@srv-twry @Kreijstal Thanks. I'll just add the docs before now, before merge.

srv-twry:
Please fix the build @schedutron

schedutron:
I don't know why the last test is failing. As far as I know, that error message shouldn't be sent at all if the app has testing configuration. And if that message is sent, a "Sendgrid key not defined" message should appear in the log, which is absent :(

schedutron:
I have removed the general error message. Because the place where the chances of errors occurring is high is celery - and since the celery task runs in a separate thread, the general error message wouldn't help in catching it. Travis should pass now.

PR #4996, opened 2018-06-30
fix: Permissions for updating orders

### LINKED ISSUE #4995: Modify permissions for updating Order

We need to modify the permissions for updating the order so that we can follow the plan added in the parent issue.
Also remove the required field constraints from the schema.

Parent: #4852

srv-twry:
Fixes #4995 

#### Checklist

We need to modify the permissions for updating the order so that we can follow the plan added in the parent issue.

Modified permissions:
1. admin can update all the fields.
2. event organizer
 a. own orders: he/she can update selected fields.
 b. other's orders: can only update the status that too when the order mode is free. No refund system.
3. order user can update selected fields of his/her order when the status is pending.

Also removed the required field constraints.

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -191,20 +191,50 @@ class OrderDetail(ResourceDetail):
 
     def before_update_object(self, order, data, view_kwargs):
         """
+        before update object method of order details
+        1. admin can update all the fields.
+        2. event organizer
+            a. own orders: he/she can update selected fields.
+            b. other's orders: can only update the status that too when the order mode is free. No refund system.
+        3. order user can update selected fields of his/her order when the status is pending.
+        The selected fields mentioned above can be taken from get_updatable_fields method from order model.
         :param order:
         :param data:
         :param view_kwargs:
         :return:
         """
-        # Admin can update all the fields while Co-organizer can update only the status
-        if not has_access('is_admin'):
-            for element in data:
-                if element != 'status':
-                    setattr(data, element, getattr(order, element))
+        if (not has_access('is_coorganizer', event_id=order.event_id)) and (not current_user.id == order.user_id):
+            raise ForbiddenException({'pointer': ''}, "Access Forbidden")
 
-        if not has_access('is_coorganizer', event_id=order.event.id):
-            raise ForbiddenException({'pointer': 'data/status'},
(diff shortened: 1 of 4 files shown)


app/models/order.py
```diff
@@ -17,6 +17,13 @@ def get_new_order_identifier():
         return get_new_order_identifier()
 
 
+def get_updatable_fields():
```
schedutron: Would it be a good practice to have this as a static attribute of the model itself?

app/api/orders.py
```diff
@@ -188,20 +188,50 @@ def before_get_object(self, view_kwargs):
...
+        if (not has_access('is_coorganizer', event_id=order.event_id)) and (not current_user.id == order.user_id):
+            raise ForbiddenException({'pointer': ''}, "Access Forbidden")
 
-        if not has_access('is_coorganizer', event_id=order.event.id):
-            raise ForbiddenException({'pointer': 'data/status'},
-                                     "To update status minimum Co-organizer access required")
+        if has_access('is_coorganizer_but_not_admin', event_id=order.event_id):
+            if current_user.id == order.user_id:
+                # Order created from the tickets tab.
+                for element in data:
```
schedutron: You can have some parts of this as a helper function (subroutine)? That way you can use it below as well. Would that be a good idea or an overkill?

srv-twry: We could but I feel like if we have it like this then it is great for the readability of the code. You can just read the docstring and read the code and everything will be clear to you. Not sure this would be the case if I encapsulate it into a helper function.

Kreijstal:
Travis is failing!

app/models/order.py
```diff
@@ -109,6 +109,13 @@ def __str__(self):
     def get_invoice_number(self):
         return 'O' + str(int(time.mktime(self.created_at.timetuple()))) + '-' + str(self.id)
 
+    @staticmethod
+    def get_updatable_fields(cls):
```
maxlorenz: `cls` is unused, right? Also wouldn't be a public constant better here (as this has to be evaluated on every call)

srv-twry: This is exactly what I did earlier but I was told to add it inside the model class. The point about getting evaluated on every call never occured to me then otherwise I would have countered that argument. Thanks, will move it outside the model class.

PR #5036, opened 2018-07-06
Add checkout times for attendees

### LINKED ISSUE #4936: API for Attendee Check Out Times
There should be API for check out times of an Attendee. Organizer should be able to create, update and get check out times of an Attendee. The feature is present in EventBrite Organizer App and should be included in Open Event Orga app which requires API support.

Development of API to provide csv data of checkout times.

**Additional context**
Please refer EventBrite Organizer app.

schedutron:
Fixes #4936

#### Checklist

This PR implements the `checkout-times` attribute for attendees.


app/api/schema/attendees.py
```diff
@@ -47,7 +47,9 @@ class Meta:
     ticket_id = fields.Str(allow_none=True)
     is_checked_in = fields.Boolean()
     checkin_times = fields.Str(allow_none=True)
+    checkout_times = fields.Str(allow_none=True)
     attendee_notes = fields.Str(allow_none=True)
+    is_checked_out = fields.Boolean()
```
srv-twry: is_checked_out = not is_checked_in. We already have `is_checked_in` in the schema. Why add a redundant one ?

app/api/attendees.py
```diff
@@ -149,6 +149,18 @@ def before_update_object(self, obj, data, kwargs):
                 if obj.checkin_times and data['checkin_times'] not in obj.checkin_times.split(","):
                     data['checkin_times'] = '{},{}'.format(obj.checkin_times, data['checkin_times'])
 
+        if 'is_checked_out' in data and data['is_checked_out']:
+            if 'checkout_times' not in data:
+                raise UnprocessableEntity(
+                    {'pointer': '/data/attributes/checkout_times'},
+                    "Check out time missing while trying to check out"
+                    "attendee")
```
srv-twry: This means we're expecting the client to send in the checkout_times. Why can't we just use the current time in the server for it ?

app/api/schema/attendees.py
schedutron: I thought about it, but I guess when `is_checked_in` is true and `is_checked_out` is false, the person is currently attending the event?

app/api/attendees.py
schedutron: So the column should be present in the model but this field should be absent from the attendee schema? Also, after what action should the current time be specified as the checkout_time in the server?

app/api/schema/attendees.py
srv-twry: I am not sure about it. I think `is_checked_in` = true implies currently attending and if it is false, then it means not attending currently. 

@iamareebjamal Can you please confirm ?

app/api/attendees.py
srv-twry: sorry, but I didn't understand what you meant by the first question. 
All I am trying to say is that once a PATCH request is made to checkout the attendee, then do save the current time as one of the checkout times in the after_update_object method.

app/api/schema/attendees.py
schedutron: Not attending can have 2 states - not arrived and, arrived but currently absent?

app/api/attendees.py
schedutron: Okay, I think I misinterpreted you - I thought you were saying that `checkout_times` should be a read-only value.

srv-twry: We do something similar in columns such as `created_at` although it is a one time action so it is done inside the constructor. Here attendee can be checked out multiple times. So just check for it in the `before_update_object` and add the current time into the data as one of the checkout times.

**Note**

In my above comment, I accidently said `after_update_object`. Use `before_update_object`.

app/api/schema/attendees.py
poush: Yeah, both being false can also be interpreted as if the attendee is/was absent. Although, I have few concerns, 
first why the name is plural "checkin_times"? I believe it will have only one timestamp right?
second, why you even need a flag for it? Simply, checked in timestamp being null represents that attendee never checked in and not null as the timestamp of check-in. This although affects the frontend as well if they are already dependent on flags but will make it easier and redundant free.

app/api/attendees.py
poush: Yes, makes sense. If we see the use case of the checkout button. It will be called whenever user checkout from the event so this we can definitely make use of current timestamps. And yes this way the `checkout_times` will be read only.

app/api/attendees.py
```diff
@@ -149,6 +151,19 @@ def before_update_object(self, obj, data, kwargs):
...
+            else:
+                if obj.checkout_times and data['checkout_times'] not in \
+                        obj.checkout_times.split(","):
+                    data['checkout_times'] = '{},{},{}'.format(
```
poush: I didn't get the point of last one. If you are expecting a checkout_time from the frontend then why are you appending both the user one and the current timestamp ?

schedutron: Since we just established that it will be a read-only value, we can eliminate the checkout time sent by the FE?

poush: Yeah we can.

schedutron:
@poush @srv-twry @mayank8318 @bhaveshAn @mishuvs Please review. Travis is fixed now.

app/api/attendees.py
```diff
@@ -154,6 +156,12 @@ def before_update_object(self, obj, data, kwargs):
                 if obj.checkin_times and data['checkin_times'] not in obj.checkin_times.split(","):
                     data['checkin_times'] = '{},{}'.format(obj.checkin_times, data['checkin_times'])
 
+        if 'is_checked_out' in data and data['is_checked_out']:
```
srv-twry: Logic is wrong. Try to simulate things in order to check for logic. 
This will add the checkout_time for every time an update request is made and `is_checked_out` is true. Is this correct ? Absolutely not. You have to check that the value of `is_checked_out` was initially false and then this PATCH request is making it true.

PR #5126, opened 2018-07-17
fix: Shift cookie_policy column from users table to settings

### LINKED ISSUE #5125: Shift cookie_policy column from users table to settings
**Describe the bug**
As pointed out by @srv-twry here "In reference to #5068 
@mayank8318 Looking into this issue in detail, I found that we don't need that column for cookie policy in the user model. Users don't have to explicitly approve it. They are approving it just by using the site.
The requirement of this issue is that we should allow the admin of the system to add this information to the site. Also, add a default cookie policy with the message that @mariobehling has mentioned.
Please revert your PR and simply add a cookie-policy in the settings column."

mayank8318:
Fixes #5125 

#### Checklist

Revert changes of #5113  and add the cookie_policy column to settings table

- Revert changes of #5113 
- Add the cookie_policy column to settings table
- Include the migration


app/models/setting.py
```diff
@@ -209,6 +223,7 @@ def __init__(self,
         self.email_from = email_from
         self.email_from_name = email_from_name
         self.frontend_url = frontend_url
+        self.cookie_policy = cookie_policy
```
bhaveshAn: You have added `cookie_policy` and `cookie_policy_link` in factories but __init__ function is only adding `cookie_policy` IMO this is the reason for travis dredd failure. Also if you added both attrs in schema then there is need to update blueprints too

mayank8318: Oh okay I will do that :+1:

PR #5131, opened 2018-07-17
fix: HTTP 500 on patch order

### LINKED ISSUE #5130: Key error 'user_id' in PATCH order.
**Describe the bug**

The server returns HTTP 500 when sending a PATCH request to the /orders/<identifier> endpoint.

srv-twry:
Fixes #5130 

#### Checklist

HTTP 500, key error 'user_id' when hitting the /orders/<identifier>.

Fixed it.

### DIFF
--- a/app/api/helpers/permission_manager.py
+++ b/app/api/helpers/permission_manager.py
@@ -354,8 +354,8 @@ def permission_manager(view, view_args, view_kwargs, *args, **kwargs):
         try:
             order = Order.query.filter_by(identifier=view_kwargs['order_identifier']).one()
         except NoResultFound:
-            return NotFoundError({'parameter': 'order_identifier'}, 'Event not found').respond()
-        view_kwargs['id'] = order.identifier
+            return NotFoundError({'parameter': 'order_identifier'}, 'Order not found').respond()
+        view_kwargs['id'] = order.id
 
     # If event_identifier in route instead of event_id
     if 'event_identifier' in view_kwargs:
@@ -435,7 +435,6 @@ def permission_manager(view, view_args, view_kwargs, *args, **kwargs):
                 kwargs[kwargs['fetch']] = fetched
         else:
             return NotFoundError({'source': ''}, 'Object not found.').respond()
-
     if args[0] in permissions:
         return permissions[args[0]](view, view_args, view_kwargs, *args, **kwargs)
     else:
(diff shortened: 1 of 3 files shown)


app/api/orders.py
```diff
@@ -302,7 +302,7 @@ def before_delete_object(self, order, view_kwargs):
             raise ConflictException({'source': ''}, 'You cannot delete a placed/completed paid order.')
 
     decorators = (jwt_required, api.has_permission(
-        'is_user_itself', methods="PATCH,DELETE", fetch="id", fetch_as="order_id", model=Order),)
+        'is_user_itself', methods="PATCH,DELETE", fetch="user_id", model=Order),)
```
schedutron: Why `user_id`?

srv-twry: Because we need user_id. `is_user_itself` needs the user id. We need to check if this user is owner of the order. So we compare the current logged in user with the `user_id` of the order owner.

schedutron: But doesn't `before_update_object()` does that already?

app/api/helpers/permission_manager.py
```diff
@@ -354,8 +354,8 @@ def permission_manager(view, view_args, view_kwargs, *args, **kwargs):
         try:
             order = Order.query.filter_by(identifier=view_kwargs['order_identifier']).one()
         except NoResultFound:
-            return NotFoundError({'parameter': 'order_identifier'}, 'Event not found').respond()
-        view_kwargs['id'] = order.identifier
+            return NotFoundError({'parameter': 'order_identifier'}, 'Order not found').respond()
+        view_kwargs['id'] = order.id
```
schedutron: For me, using `id` instead of `identifier` here threw an error, because it then compared order `id` with `identifier` later, in the library code.

srv-twry: For me it is working. Not sure why it wasn't working for you.

app/api/orders.py
srv-twry: 1. You added this decorator, so I thought it is needed. 
2. If this is redundant then we can remove this later or in another PR. This is super needed right now in the frontend.

srv-twry:
Don't merge this guys.

app/api/helpers/permission_manager.py
schedutron: I tried using `id` instead of `identifier` and it's working now. I don't recall why it was not working then. Maybe because I was using a modified order schema.

app/api/orders.py
schedutron: The code to add `id` key in `view_kwargs` requires having this decorator, that's why I added it. Yes, we can indeed fix redundancies in another PR. If the current code does the job for now, no issues with that :)

PR #5143, opened 2018-07-18
Fixes #5142 Allow super-admin to add admins

### LINKED ISSUE #5142: For the edit action button in admin/users the super admin should be allowed to make the users as an admin.
**Describe the bug**

The patch for the users is working but is_admin is a readonly property even for the super user. Used the patch request for users and changed the value for is_admin from true to false. But there is no change after the request to the value of is_admin. 

**To Reproduce**
1. Go to v1/users/user_id
2. Send a patch request
3. Change the value of is_admin for some user from false to true.

image

After the request
image

bhaveshAn:
Fixes #5142 Allow super-admin to add admins

#### Checklist

Fixes #5142 Allow super-admin to add admins

- Fixes #5142 Allow super-admin to add admins
screenshot from 2018-07-19 03-12-51

### DIFF
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -184,7 +184,7 @@ class UserDetail(ResourceDetail):
             view_kwargs['email_changed'] = user.email
 
         if data.get('is_admin') != user.is_admin:
-            data['is_admin'] = not user.is_admin
+            user.is_admin = not user.is_admin
 
     def after_update_object(self, user, data, view_kwargs):
         """
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -183,6 +183,9 @@ class UserDetail(ResourceDetail):
         if data.get('email') and data['email'] != user.email:
             view_kwargs['email_changed'] = user.email
 
+        if data.get('is_admin') != user.is_admin:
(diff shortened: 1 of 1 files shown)


app/api/users.py
```diff
@@ -183,6 +183,9 @@ def before_update_object(self, user, data, view_kwargs):
         if data.get('email') and data['email'] != user.email:
             view_kwargs['email_changed'] = user.email
 
+        if data.get('is_admin') != user.is_admin:
+            data['is_admin'] = not user.is_admin
```
srv-twry: What is this even doing ??
This is doing absolutely nothing IMO. 
Let `data.get('is_admin')` = A and `user.is_admin` = B
1. A = true, B = true - The if statement isn't true so nothing happens
2. A = false B = True - So the if statement is correct. Now you set A = not B and hence A = not True = False. But A is already false. 
3. A = true, B = False - If statement is true. Now you set A = not B and hence A = not False = True, but A is already true.
4. A = False, B = false - If statement isn't true and hence it won't run. 

How does this have any impact on the values ???

bhaveshAn:
@srv-twry @Kreijstal This PR canbe merged

PR #5366, opened 2018-09-09
Confirming the password in create_db.py

### LINKED ISSUE #5364: Password confirmation while creating tables in local installation
When I was installing the project locally, after running create_db.py, it didn't ask me to re-confirm the password, and password can easily be mistyped in the first time.  

So it would be better if we add password confirmation in that part.

prateekj117:
Ask the user to enter the password again in case
the password entered is less than 4 characters or
the re-entered password do not match the password
typed earlier

Fixes #5364

### DIFF
--- a/create_db.py
+++ b/create_db.py
@@ -20,8 +20,17 @@ def _validate_email(email):
 def _validate_password(password):
     if len(password) < 4:
         print('\nPassword should have minimum 4 characters')
-        sys.exit(1)
+        password = getpass.getpass("\nEnter your password again      :")
+        _validate_password(password)
 
+
+def _validate_repassword(password, repassword):
+    if password != repassword:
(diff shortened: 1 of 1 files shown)


create_db.py
```diff
@@ -20,8 +20,17 @@ def _validate_email(email):
...
+        _validate_password(password)
+
 
+def _validate_repassword(password, repassword):
+    if password != repassword:
+        print('\nPassword did not match')
+        password = getpass.getpass("\nEnter your password again      :")
+        _validate_password(password)
+        repassword = getpass.getpass("Enter your password again to confirm : ")
+        _validate_repassword(password, repassword)
```
iamareebjamal: This is recursion, you should use a loop instead

PR #5367, opened 2018-09-09
Confirm the password by entering it again 

### LINKED ISSUE #5364: Password confirmation while creating tables in local installation

prateekj117:
Confirm the password by entering it again and ask the user to enter the password again in case the password entered is less than 4 characters or the re-entered password do not match the password typed earlier

Fixes #5364

### DIFF
--- a/create_db.py
+++ b/create_db.py
@@ -20,8 +20,17 @@ def _validate_email(email):
 def _validate_password(password):
     if len(password) < 4:
         print('\nPassword should have minimum 4 characters')
-        sys.exit(1)
+        password = getpass.getpass("\nEnter your password again      :")
+        _validate_password(password)
 
+
+def _validate_repassword(password, repassword):
+    if password != repassword:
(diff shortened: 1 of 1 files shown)


create_db.py
```diff
@@ -20,17 +20,26 @@ def _validate_email(email):
 def _validate_password(password):
     if len(password) < 4:
         print('\nPassword should have minimum 4 characters')
-        sys.exit(1)
+        _enter_password()
```
iamareebjamal: This is still recursion

iamareebjamal:
The logic is very simple

```
ask = True
while ask:
  password = ...
  if invalid_password:
    continue
  re_password = ...
  if not matches:
    continue
  ask = False
```

Simple, no recursion, no extra functions

PR #5396, opened 2018-10-03
Add Pipfile and Pipfile.lock and closes fossasia/open-event-server#5391

### LINKED ISSUE #5391: Using Pipenv for Packaging
As an alternative to `virtualenv/venv` which is currently used in the project(requirements.txt), we can use pipenv also for easy installation. Also now `pipenv` is officially recommended python package and virtualization manager.
Please refer https://stackoverflow.com/questions/46330327/please-explain-the-usage-of-pipfile-and-pipfile-lock

ghost:
Fixes #5391

Add Pipenv and Pipenv.lock to configure the project with pipenv.

All the dependencies are locked, as is the python version.

Installation of Pipenv is required (Pyenv is optional).

After installation, only run `pipenv sync`


tonythomas01:
I think you will have to also update the installation docs. Like `python3 manage.py ...` to `python manage.py etc` ?

bhaveshAn:
@castrogeraldo Please help adding the docs for installing app using pipenv within this PR. Thanks!

PR #5516, opened 2019-01-14
Set up logging correctly

### LINKED ISSUE #5498: Set up logging correctly
Currently, logging is hardcoded in the app and never really used. Logging should be used almost on every action with proper label as:

- trace
- debug
- info
- warn
- error

Right now, logger is hardcoded to show ERROR only. This needs to be configurable through the command line. 

Also, I think default should be `INFO` or even `DEBUG` for the app itself and warn for other 3rd party libraries.

prateekj117:
Fixes #5498 

Right now, logger is hardcoded to show ERROR only. The default should be INFO for the app itself and warn for other 3rd party libraries.


iamareebjamal:
Write `app.logger.debug('Hello')`

This should not have been printed but it is

iamareebjamal:
My laptop's broken and this needs to be verified. Please ask your peers to review and verify this

app/__init__.py
```diff
@@ -223,7 +244,7 @@ def update_sent_state(sender=None, body=None, **kwargs):
 # register celery tasks. removing them will cause the tasks to not function. so don't remove them
 # it is important to register them after celery is defined to resolve circular imports
 
-from .api.helpers import tasks
+from app.api.helpers import tasks
```
iamareebjamal: Revert this

PR #5545, opened 2019-01-26
Read from environment variables when creating the super admin user

### LINKED ISSUE #5522: Read from environment variables when creating the super admin user

When deploying on Kubernetes or Openshift, the super admin account points to the default one, that is hardcoded in manage.py file. https://github.com/fossasia/open-event-server/blob/fix-celery/manage.py#L89

Extend manage.py file to read environment variables on database initialization. If the user doesn't provide these variables then use the default ones.

prateekj117:
Fixes #5522 

Extend manage.py file to read environment variables on database initialization. If the user doesn't provide these variables then use the default ones.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -79,16 +79,18 @@ def initialize_db(credentials):
                 print("[LOG] Could not create tables. Either database does not exist or tables already created")
             if populate_data:
                 credentials = credentials.split(":")
-                create_super_admin(credentials[0], credentials[1])
+                admin_email = os.environ.get('SUPER_ADMIN_EMAIL', credentials[0])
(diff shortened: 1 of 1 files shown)


manage.py
```diff
@@ -79,16 +79,18 @@ def initialize_db(credentials):
...
+                admin_password = os.environ.get('SUPER_ADMIN_PASSWORD', credentials[1])
+                create_super_admin(admin_email, admin_password)
                 populate()
         else:
             print("[LOG] Tables already exist. Skipping data population & creation.")
 
 
 @manager.command
-def prepare_kubernetes_db():
+def prepare_kubernetes_db(credentials='open_event_test_user@fossasia.org:fossasia'):
```
iamareebjamal: This will override the environment variables

PR #5560, opened 2019-01-30
Event model: Add info about presence of sessions,speakers

### LINKED ISSUE #5529: Update docs on Events end point to accomodate additional fields  for Front end's sidebar
The frontend facilitates all the public aspects of an event through a side menu as shown below.
image

However, it is desired to only show a menu item, if it's corresponding entries exist. For instance, display the item sessions, if sessions are present, display speakers if speakers are present.  Display schedule if the schedule is published etc.

To achieve this functionality the frontend has to make a call to the corresponding endpoint to check if sessions or speakers exist, but that results in sending 100s of network requests for data, user is probably not going to view anyway. 

Thus there needs to be an api endpoint or property values in the event schema itself, which tells if **sessions**,  **speakers** are present for a given event. Current production environment has revealed that querying sessions and speakers results in a significant overhead.

CosmicCoder96:
Fixes #5529 

Eliminates the need to fetch all sessions and speakers of an event just to check if they exist.

- Add field `has_sessions` to the event model
- Add field `has_speakers` to the event model

### DIFF
--- a/app/api/schema/events.py
+++ b/app/api/schema/events.py
@@ -82,6 +82,8 @@ class EventSchemaPublic(SoftDeletionSchema):
     organizer_name = fields.Str(allow_none=True)
     is_map_shown = fields.Bool(default=False)
     has_organizer_info = fields.Bool(default=False)
+    has_sessions = fields.Bool(default=0, dump_only=True)
+    has_speakers = fields.Bool(default=0, dump_only=True)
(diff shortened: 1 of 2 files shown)


iamareebjamal:
@CosmicCoder96 This will slow down the list queries by a lot. This should be handled by a separate relationship to event

iamareebjamal:
As mentioned, this slowed down queries by large amount

PR #5632, opened 2019-02-20
fix : ensure distinct organisers in no. of organisers

### LINKED ISSUE #5631: admin dashboard : number of organisers > number of actual users
**Describe the bug**

Currently, the admin dashboard is effectively reporting the number of `published + draft` events in the `No. of organizers` tab which should not be the case.
In my example, the number of total organisers is 3 while the number reported here is 7

**To Reproduce**
1. Go to Admin Dashboard
2. See error

**Expected behavior**

It should be reporting the actual number of organizers

image
image

**Additional details (please complete the following information):**
 - OS: [e.g. MacOS, Ubuntu, CentOS] Ubuntu
 - Python Version [e.g. `3.5`, `3.6`] 3.5

**Additional context**

Working on the issue

uds5501:
Fixes #5631 

#### Checklist

It makes sure that distinct users on a particular role are being counted instead of the whole query result

- Creating a local helper to give the length of distinct users on an event role
- Using that helper instead of get_count helper

image

### DIFF
--- a/app/api/admin_statistics_api/users.py
+++ b/app/api/admin_statistics_api/users.py
@@ -12,7 +12,6 @@ from app.models.users_events_role import UsersEventsRoles
 from app.models.role import Role
 from app.api.helpers.db import get_count
 
-
 class AdminStatisticsUserSchema(Schema):
     """
     Api schema
@@ -49,21 +48,21 @@ class AdminStatisticsUserSchema(Schema):
 
     def get_all_user_roles(self, role_name):
         role = Role.query.filter_by(name=role_name).first()
-        uers = UsersEventsRoles.query.join(UsersEventsRoles.event).join(UsersEventsRoles.role).filter(
-            Event.deleted_at.is_(None), UsersEventsRoles.role == role)
-        return uers
+        newquery = User.query.join(UsersEventsRoles.user).join(UsersEventsRoles.role).filter(
+            UsersEventsRoles.role == role).distinct()
+        return newquery
 
     def organizer_count(self, obj):
-        return get_count(self.get_all_user_roles('organizer'))
+        return self.get_all_user_roles('organizer').count()
 
     def coorganizer_count(self, obj):
-        return get_count(self.get_all_user_roles('coorganizer'))
+        return self.get_all_user_roles('coorganizer').count()
 
     def track_organizer_count(self, obj):
-        return get_count(self.get_all_user_roles('track_organizer'))
+        return self.get_all_user_roles('track_organizer').count()
 
     def attendee_count(self, obj):
-        return get_count(self.get_all_user_roles('attendee'))
+        return self.get_all_user_roles('attendee').count()
 
 
 class AdminStatisticsUserDetail(ResourceDetail):


app/api/admin_statistics_api/users.py
```diff
@@ -53,17 +53,28 @@ def get_all_user_roles(self, role_name):
             Event.deleted_at.is_(None), UsersEventsRoles.role == role)
         return uers
 
+    def get_distinct_roles_count(self, query_result):
+        """
+        Returns the length of set of users for the given query
+        """
+        distinct_entry = set([i.user for i in query_result])
```
iamareebjamal: what is i?

app/api/admin_statistics_api/users.py
```diff
@@ -53,17 +53,28 @@ def get_all_user_roles(self, role_name):
...
+        return len(distinct_entry)
+
     def organizer_count(self, obj):
-        return get_count(self.get_all_user_roles('organizer'))
+        query_result = self.get_all_user_roles('organizer')
+        return self.get_distinct_roles_count(query_result)
```
iamareebjamal: Very unoptimized approach. The query itself should use distinct count. Not load up every user in memory and then add in a set

Please do the same for others as well

app/api/admin_statistics_api/users.py
uds5501: @iamareebjamal I will improvise on the variable choice in the next commit

app/api/admin_statistics_api/users.py
uds5501: @iamareebjamal Understood, will try and improve it in the next commmit

uds5501: @iamareebjamal Please check out the query now.

uds5501:
@iamareebjamal @CosmicCoder96 @niranjan94 @srv-twry Please review

app/api/admin_statistics_api/users.py
```diff
@@ -47,34 +46,26 @@ def verified_count(self, obj):
...
-        """
-        Returns the length of set of users for the given query
-        """
-        distinct_entry = set([i.user for i in query_result])
-        return len(distinct_entry)
+        newquery = User.query.join(UsersEventsRoles.user).join(UsersEventsRoles.role).filter(
+            UsersEventsRoles.role==role).distinct()
+        counter = 0
+        for element in newquery:
+            counter += 1
```
iamareebjamal: Still the same problem as before but much worse

uds5501: @iamareebjamal could you please elaborate on the similarity and the approach to be adopted for tackling this issue?

I am using `distinct` method for the query already, how do you want me to optimise it. I would really appreciate your guidance here

iamareebjamal: See the get_count method. You should not use any loop or set of any other pythonic construct to get the count from the database

uds5501: @iamareebjamal oh, so the counting is the problem. Understood, I shall look into it

app/api/admin_statistics_api/users.py
```diff
@@ -47,23 +47,23 @@ def verified_count(self, obj):
...
-        uers = UsersEventsRoles.query.join(UsersEventsRoles.event).join(UsersEventsRoles.role).filter(
-            Event.deleted_at.is_(None), UsersEventsRoles.role == role)
-        return uers
+        newquery = User.query.join(UsersEventsRoles.user).join(UsersEventsRoles.role).filter(
+            UsersEventsRoles.role==role).distinct()
+        return newquery
 
     def organizer_count(self, obj):
-        return get_count(self.get_all_user_roles('organizer'))
+        return self.get_all_user_role('organizer').count()
```
iamareebjamal: Use get_count

app/api/admin_statistics_api/users.py
uds5501: @iamareebjamal Better now?

app/api/admin_statistics_api/users.py
uds5501: @iamareebjamal `get_count` function is returning a wrong count (of 8 TBP) instead of desired 3

iamareebjamal: If you pass `self.get_all_user_role('organizer')`?

iamareebjamal:
What if you return `get_count(self.get_all_user_role('organizer'))`?

app/api/admin_statistics_api/users.py
uds5501: @iamareebjamal Yes, it's giving 8 then

uds5501:
@iamareebjamal I get a count of 8 on using the `get_count(self.get_all_user_role('organizer'))`

uds5501:
@iamareebjamal  I found a few more schemas using `.count()` instead of `get_count`
Please see:
https://github.com/fossasia/open-event-server/blob/324faf6acb0c4e53150ae558c6605dfa83d47018/app/api/schema/event_statistics.py#L35-L60

PR #5636, opened 2019-02-22
fix: Donations settings bug(PATCH)

### LINKED ISSUE #5618: Donations PATCH error
**Describe the bug**
Donations aren't getting enabled on the admin panel. Throws a 409 conflict error.

**To Reproduce**
1. Go to 'Admin Panel'
2. Click on 'Modules Section'
3. See unexpected error

**Additional context**
Working on this.
screenshot from 2019-02-18 15-12-59

mrsaicharan1:
Fixes #5618 

#### Checklist

- Fixed the bug for donations where `before_patch `was checking for the wrong condition


mrsaicharan1:
@CosmicCoder96 @iamareebjamal Please review

mrsaicharan1:
@iamareebjamal Shouldn't it be checking if `donations_include` is  False?

uds5501:
@mrsaicharan1 Yes, it should be checking that. On the side note, I think you accidentally created your PR on a merged branch. (was it intended though?)

app/api/modules.py
```diff
@@ -29,7 +29,7 @@ def before_patch(self, args, kwargs, data=None):
         :param data:
         :return:
         """
-        if data.get('donation_include'):
+        if not data.get('donation_include'):
```
iamareebjamal: ???

This is wrong logic AFAIK

mrsaicharan1: After these changes, I am able to turn the donations on. But turning them off, unexpected error occurs. There was vice-versa state before these changes. What could the error possibly be?

iamareebjamal: You are checking if the data has donation_include or not. You should check if the settings has donations enabled

CosmicCoder96: @mrsaicharan1 can you update this PR please.

mrsaicharan1: @CosmicCoder96 Yes. Will update by today. Should I be checking for donations_include in `before_patch` itself?

CosmicCoder96: >You are checking if the data has donation_include or not. You should check if the settings have donations enabled

Did you get the point he made? 
You checked if the request received contains a filed called donation_include. 
In addition to checking that, you also need to check if that field is present in the response, has its value as `true` or `false`. `data.get` only checks the existence of the field, `data['donation_include']` will give you the value of the field.

mrsaicharan1: Got it. Will update it right now.

CosmicCoder96: Also this check shouldnt be here, it should be present in the events api. If an event organizer tries to pass a value for donations, and they are not enabled in the system, an exception should be thrown.

mrsaicharan1: @CosmicCoder96 Please tell me if there's anything else which is required

mrsaicharan1: @CosmicCoder96 Can I get an update on this please?

mrsaicharan1: @iamareebjamal Can you look into this as well?

app/api/modules.py
```diff
@@ -29,9 +28,6 @@ def before_patch(self, args, kwargs, data=None):
         :param data:
         :return:
         """
-        if data.get('donation_include'):
```
iamareebjamal: Why remove this

mrsaicharan1: This Check should be present in the events api where if the donations are disabled and the organiser tries to pass something for donations, an exception should be thrown(which has already been incorporated in the code). As quoted by @CosmicCoder96, I think this check shouldn't be present here.

CosmicCoder96: yeah @iamareebjamal I believe this exact check should be placed in the events api, and this very exception should be raised *after* checking from the modules api if donation is enabled in the system. If the admin has not enabled it, then this field should not be present in the request body at all.

mrsaicharan1:
@CosmicCoder96 I believe that this check is already present in the events
API.

mrsaicharan1:
@CosmicCoder96 I believe that this check is already present in the events
API.
https://github.com/fossasia/open-event-server/blob/development/app/api/events.py

mrsaicharan1:
@iamareebjamal Please review

PR #5637, opened 2019-02-22
fix: prevent deleted users from logging in

uds5501:
Fixes fossasia/open-event-frontend#2216

#### Checklist

It adds a utilisation of  default `deleted_at` attribute to the User model and in the authentication query, makes sure that deleted_at is `None`

- setting the field to None and querying for the same

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -78,7 +78,6 @@ class User(SoftDeletionModel):
     was_registered_with_order = db.Column(db.Boolean, default=False)
     last_accessed_at = db.Column(db.DateTime(timezone=True))
     created_at = db.Column(db.DateTime(timezone=True), default=datetime.now(pytz.utc))
-    deleted_at = db.Column(db.DateTime(timezone=True), default=None)
     speaker = db.relationship('Speaker', backref="user")
     favourite_events = db.relationship('UserFavouriteEvent', backref="user")
     session = db.relationship('Session', backref="user")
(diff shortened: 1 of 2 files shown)


uds5501:
@iamareebjamal @CosmicCoder96 Please have a look and see if anything else is required?

app/models/user.py
```diff
@@ -78,6 +78,7 @@ class User(SoftDeletionModel):
     was_registered_with_order = db.Column(db.Boolean, default=False)
     last_accessed_at = db.Column(db.DateTime(timezone=True))
     created_at = db.Column(db.DateTime(timezone=True), default=datetime.now(pytz.utc))
+    deleted_at = db.Column(db.DateTime(timezone=True), default=None)
```
CosmicCoder96: user model class is a subclass of `SoftDeletionModel` by virtue of which it already has a `deleted_at` column. In fact that's where the `deleted_at` timestamp is stored. What are you trying to do here? Making changes to model files actually adds a column in the database. There already is a `deleted_at` column for users.

uds5501: @CosmicCoder96 Actually, I was just trying to  change the query so that the `deleted_at=None` can be executed in the filter. I was aware of the presence of `deleted_at` element but it was not working as expected (giving a `401 Authentication Error`)

So i added a deleted_at field manually to see if it works then and it did. 
I think I will retest and see if it was some error from my side

uds5501: @CosmicCoder96 My bad, it was incorrect testing from my side. Please review now

PR #5657, opened 2019-03-02
feat: add conflict check for TicketFee

uds5501:
Fixes fossasia/open-event-frontend#2169

#### Checklist

this makes sure no more than 1 object of a particular Currency-Country combination is being formed

- implement conflict checks for currency-country combinations


uds5501:
ezgif com-video-to-gif 2

Only unique objects are present now:

unqieucombos

uds5501:
@iamareebjamal @CosmicCoder96 Please review :) :beers:

uds5501:
@CosmicCoder96 @iamareebjamal According to my API conditions, this particular test case should've failed. So, how do I fix that Travis build?

app/api/ticket_fees.py
```diff
@@ -10,10 +12,35 @@ class TicketFeeList(ResourceList):
...
+                except NoResultFound:
+                    pass
+                else:
+                    raise ConflictException(
+                        {'pointer': ''},
+                        "({}-{}) Combination already exists".format(data['currency'], data['country']))
+            else:
+                raise UnprocessableEntity({'source': ''}, "Country or Currency cannot be NULL")
+        else:
+            raise UnprocessableEntity({'source': ''}, "Country or Currency Attribute is missing")
```
iamareebjamal: Use marshmallow for validation

uds5501: @iamareebjamal could you please elaborate?

iamareebjamal: https://marshmallow.readthedocs.io/en/3.0/

uds5501: Understood, shall commit soon

uds5501:
@iamareebjamal I was using `@validate_schema` decorator to apply the condition  checks on schema level, but it can't differenciate between a `POST` and a `PATCH` request. 
My objective is to allow PATCH requests but deny any POST request which aims to create a new object of same currency and country which is already existing.
What shall I do in this case??

iamareebjamal:
The same thing is applied to all types of schema like event, speaker, etc.

Check those. 

You can use before hooks but don't validate the schema there.

uds5501:
@iamareebjamal Thanks! Fixed it now, committing

shreyanshdwivedi:
@uds5501 can you please update this ? It seems quite an important bug to me as it deals with the fees charged.

uds5501:
@shreyanshdwivedi I will be needing some help regarding the marshmallow schema validation. As to how to determine if the new combination is present or not

app/api/ticket_fees.py
shreyanshdwivedi: @uds5501 I think here @iamareebjamal simply means that use `allow_none=False` in the schema itself. It will raise error if country or currency is missing and you can use `nullable=False` in model to disallow use of NULL values.

shreyanshdwivedi:
@uds5501 also don't validate schema where you are doing it currently. Instead use the before hooks which means functions like - `before_update_object` or `before_post_object` to raise errors which you are currenlty doing via validating schema.
Reference examples - https://github.com/fossasia/open-event-server/blob/ad322e3c97cf80312a71fad76d5cda84c971c0d7/app/api/events.py#L143
See how `validate_event` function is called before posting. If there is any discrepancies error is raised. Similarly you can find checks before updating the object. Implement something like this.
Hope you understood :sweat_smile:

app/api/ticket_fees.py
uds5501: @shreyanshdwivedi Ooooh! I understand now. I will update it soon

uds5501:
@mrsaicharan1 @shreyanshdwivedi @iamareebjamal  @CosmicCoder96 
I want the model to not take any `NULL` values, but for that, I would need the migration file to ensure that old entries of ticket_fees country and currency fields are filled up with some `server_default` values. 
can I set them to be 'United States' and 'USD' respectively?

iamareebjamal:
You can just define those in marshmallow schema for now

uds5501:
@iamareebjamal understood, shall do that

uds5501:
@iamareebjamal @shreyanshdwivedi Please review

app/api/schema/ticket_fees.py
```diff
@@ -20,7 +20,7 @@ class Meta:
         inflect = dasherize
 
     id = fields.Integer(dump_only=True)
-    currency = fields.Str(validate=validate.OneOf(choices=PAYMENT_CURRENCY_CHOICES), allow_none=True)
-    country = fields.String(allow_none=True)
+    currency = fields.Str(validate=validate.OneOf(choices=PAYMENT_CURRENCY_CHOICES), allow_none=False, default='USD')
```
iamareebjamal: Don't default to USD. We need the user to provide the values. There is no UI showing the user that it will default to USD if no value is provided

uds5501: @iamareebjamal So I should use no default value?

iamareebjamal: Yes

uds5501:
@iamareebjamal I have removed the default values from schema, does anything else needs to be done now?

app/api/ticket_fees.py
```diff
@@ -1,15 +1,34 @@
...
+        method to check for existing country currency combination
+        :param args:
+        :param kwargs:
+        :param data:
+        :return:
+        """
+        if (data['country'] and data['currency']):
+            try:
+                already_existing_combination = TicketFees.query.filter_by(country=data['country'],
+                                                                          currency=data['currency']).one()
```
iamareebjamal: Just add a unique key on country and currency. No need for this check

uds5501: @iamareebjamal A unique key constraint would ensure that I don't get to save similar values of the same column.

I want to be able to save :`USA - USD` and `USA-INR` combination simultaneously. I don't think adding unique constraints would allow that

iamareebjamal: You are simulating unique key constraint on TicketFee, if unique key won't work, this shouldn't work too

iamareebjamal: You are ensuring that country and currency pair is unique and not repeated. That's the work on unique key

shreyanshdwivedi:
@uds5501 similar to my PR, this is failing due to multiple heads. Please fix it

uds5501:
@shreyanshdwivedi do i need to change the `revises` in migration file here?

migrations/versions/cb4ccda76a2b_.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+# revision identifiers, used by Alembic.
+revision = 'cb4ccda76a2b'
+down_revision = '0e80c49a6e28'
```
shreyanshdwivedi: @uds5501 change `down_revision` and `revises` to `Revision ID` of migration in my latest merged PR

uds5501:
@shreyanshdwivedi this PR won't work until the dredd hook request is fixed. I am opening an issue to fix this request : 
```
{
  "data": {
    "type": "ticket-fee",
    "attributes": {
      "currency": "USD",
      "service-fee": "3.14",
      "maximum-fee": "5.14"
    }
  }
}
```

uds5501:
@iamareebjamal you are right, I will look for corresponding resources and update this soon

kushthedude:
Closed in favour of #6017

PR #5659, opened 2019-03-04
make discount code case insensitive.

### LINKED ISSUE #5653: Make discount/promotional code case-insensitive

Users get invalid discount/promotional code error because it is currently case-sensitive.

It will be a better user experience if we make the discount/promotional code case-insensitive.

**Additional context**

Anupam-dagar:
Fixes #5653 

#### Checklist

This makes case insensitive application of discount codes possible.

- use ilike to perform case insensitive search over db for discount code.

### DIFF
--- a/migrations/versions/f21e128ea3ef_create_case_insensitive_discount_codes.py
+++ b/migrations/versions/f21e128ea3ef_create_case_insensitive_discount_codes.py
@@ -1,0 +1,30 @@
+"""Create case insensitive discount codes
+
+Revision ID: f21e128ea3ef
+Revises: 7bb5891a9f2e
+Create Date: 2019-04-18 16:30:17.668956
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+# revision identifiers, used by Alembic.
+revision = 'f21e128ea3ef'
+down_revision = '7bb5891a9f2e'
+
+
+def upgrade():
+    op.execute("UPDATE discount_codes SET code = concat(code, '_') where code not in (SELECT DISTINCT ON (upper(code)) code FROM discount_codes GROUP BY event_id);",
+               execution_options=None)
+    op.execute("alter table discount_codes alter column code type citext;",
+               execution_options=None)
+
+
+def downgrade():
+    op.execute("alter table discount_codes alter column code type text;",
+               execution_options=None)
+    op.execute("drop extension citext CASCADE;",
+               execution_options=None)


app/api/discount_codes.py
```diff
@@ -207,7 +207,8 @@ def before_get(self, args, kwargs):
         if kwargs.get('code'):
             # filter on deleted_at is required to catch the id of a
             # discount code which has not been deleted.
-            discount = db.session.query(DiscountCode).filter_by(code=kwargs.get('code'), deleted_at=None).first()
+            discount = db.session.query(DiscountCode).filter(DiscountCode.code.ilike(
+                kwargs.get('code')), DiscountCode.deleted_at == None).first()
```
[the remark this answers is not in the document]
Anupam-dagar: here `is` None doesn't works.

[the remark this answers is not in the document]
iamareebjamal: Not good, make changes at DB level and create an index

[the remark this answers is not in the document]
Anupam-dagar: @iamareebjamal does this means that there should be a single entry for the case insensitive string in the db?

iamareebjamal:
@Anupam-dagar Updates

Anupam-dagar:
@iamareebjamal I was working on a frontend issue, I'll update this after that.

Anupam-dagar:
@iamareebjamal Can you provide me with a reference link on doing it.

iamareebjamal:
See the PR by @prateekj117

Anupam-dagar:
@iamareebjamal I did look at it once, can I write the migration directly? Or is it like I update the models and then generate the migrations for it?

iamareebjamal:
No updates required in models AFAIK

prateekj117:
@Anupam-dagar Making migrations directly is the best option here. Alembic does not auto-create migrations for create_index. Also, citext is a separate extension used in postgres directly, I don't think there would be any solution implementing it through alembic,so making migration directly is the best option here.

Anupam-dagar:
Thanks for your help @prateekj117 , I will try out creating a migration directly.

Anupam-dagar:
@iamareebjamal I have made the changes taking PR #5728 as a reference. Please review.

migrations/versions/f87fee85dc66_create_e_case_insensitive_discount_codes.py
```diff
@@ -0,0 +1,32 @@
...
+import sqlalchemy_utils
+
+
+# revision identifiers, used by Alembic.
+revision = 'f87fee85dc66'
+down_revision = '6e5c574cbfb8'
+
+
+def upgrade():
+    op.execute("UPDATE discount_codes SET code = concat(code, '_') FROM (SELECT lower(code) FROM discount_codes GROUP BY lower(code) HAVING COUNT(*) > 1) as discount_code where lower(code) != code;",
```
iamareebjamal: This is wrong as mentioned in that PR

Anupam-dagar:
@iamareebjamal I need to write a query which changes all the same type of codes to lower case and for future only insert lowercase of any code,right? What if we currently have 2 same discount codes with different case. Should I delete one? I dont think it to be a good idea. Should I rename it?

iamareebjamal:
Is there a unique constraint on discount code globally?

Anupam-dagar:
No, there is no unique constraint.
`code = db.Column(db.String, nullable=False)`

iamareebjamal:
OK, I'll check if there are any duplicates in production data. And a per-event unique constraint needs to be set

Anupam-dagar:
We can use a unique constraint together on event id and the discount code like this
```
__table_args__ = (
        UniqueConstraint('code', 'event_id'),
    )
```

Anupam-dagar:
@iamareebjamal please review.

migrations/versions/f87fee85dc66_create_e_case_insensitive_discount_codes.py
```diff
@@ -0,0 +1,31 @@
...
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+# revision identifiers, used by Alembic.
+revision = 'f87fee85dc66'
+down_revision = '6e5c574cbfb8'
+
+def upgrade():
+    op.execute("UPDATE discount_codes SET code = concat(code, '_') where code not in (SELECT DISTINCT ON (upper(code)) code FROM discount_codes);",
```
iamareebjamal: Will have to verify this query

Anupam-dagar: @iamareebjamal is the query okay?

iamareebjamal: Thanks for reminding. I'll check today

iamareebjamal: There's an issue in this, this will rename all duplicate accounts. We don't want that. We want duplicates to not happen inside same event, global duplicates may happen. So you have to group by event as well

Anupam-dagar:
@iamareebjamal Please review.

iamareebjamal:
There's an issue in this, this will rename all duplicate codes. We don't want that. We want duplicates to not happen inside same event, global duplicates may happen. So you have to group by event as well

Anupam-dagar:
@iamareebjamal Apologies, I messed up my git flow earlier due to which changes didn't got pushed.
I have pushed the required changes. Please review.

migrations/versions/f21e128ea3ef_create_case_insensitive_discount_codes.py
```diff
@@ -0,0 +1,30 @@
...
+import sqlalchemy_utils
+
+
+# revision identifiers, used by Alembic.
+revision = 'f21e128ea3ef'
+down_revision = '7bb5891a9f2e'
+
+
+def upgrade():
+    op.execute("UPDATE discount_codes SET code = concat(code, '_') where code not in (SELECT DISTINCT ON (upper(code)) code FROM discount_codes GROUP BY event_id);",
```
iamareebjamal: Query is wrong. Please test before asking for review

shreyanshdwivedi:
@Anupam-dagar can you please update this?

Anupam-dagar:
@shreyanshdwivedi Please continue on this PR. I won't be able to take it forward anytime soon.

prateekj117:
@Anupam-dagar In that case, close this PR, I will make a new one and work on it.

PR #5688, opened 2019-03-14
Make Error Message more descriptive.

### LINKED ISSUE #5684: Ticketing not enabled error message
**Describe the bug**
Ticketing not enabled error message comes when an unverified user tries to register an event or someone tries to register the event without giving a location.

**Expected behavior**
Error Message should be more clear and helpful to the user.

**Additional context**
Working on it.

prateekj117:
Fixes #5684 

#### Checklist


app/api/events.py
```diff
@@ -118,8 +118,12 @@ def before_post(self, args, kwargs, data=None):
         user = User.query.filter_by(id=kwargs['user_id']).first()
         modules = Module.query.first()
         if data.get('is_ticketing_enabled', False) and not modules.ticket_include:
-            raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
-                                     "Ticketing is not enabled in the system")
+            if user.is_verified:
```
iamareebjamal: These conditions are in wrong place. They have nothing to do with tickets module. Add them as separate conditions

app/api/events.py
```diff
@@ -452,10 +456,15 @@ def before_patch(self, args, kwargs, data=None):
         :param data:
         :return:
         """
+        user = User.query.filter_by(id=kwargs['user_id']).first()
         modules = Module.query.first()
         if data.get('is_ticketing_enabled', False) and not modules.ticket_include:
-            raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
-                                     "Ticketing is not enabled in the system")
+            if user.is_verified:
```
iamareebjamal: Remove duplication

app/api/events.py
```diff
@@ -51,6 +51,18 @@
 from app.models.stripe_authorization import StripeAuthorization
 
 
+def check_ticketing(user, modules, data=None):
+    if user.is_verified:
+        raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
+                                 "Please enter the location of the event")
```
iamareebjamal: If user is not verified, throw that error and then check properly if location is not defined. What has user being verified got to do with location of the event?

Also all attribute keys are wrong, why are they 'data/attributes/is-ticketing-enabled'?

app/api/events.py
```diff
@@ -117,9 +129,8 @@ def before_post(self, args, kwargs, data=None):
         """
         user = User.query.filter_by(id=kwargs['user_id']).first()
         modules = Module.query.first()
-        if data.get('is_ticketing_enabled', False) and not modules.ticket_include:
-            raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
-                                     "Ticketing is not enabled in the system")
+        if data.get('is_ticketing_enabled', False):
```
iamareebjamal: It shouldn't be checked 'if ticketing is enabled'. It should be checked anyway

app/api/events.py
```diff
@@ -452,10 +462,10 @@ def before_patch(self, args, kwargs, data=None):
         :param data:
         :return:
         """
+        user = User.query.filter_by(id=kwargs['user_id']).first()
         modules = Module.query.first()
-        if data.get('is_ticketing_enabled', False) and not modules.ticket_include:
-            raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
-                                     "Ticketing is not enabled in the system")
+        if data.get('is_ticketing_enabled', False):
```
iamareebjamal: Same here, and rename the function from `check_ticketing`

app/api/events.py
```diff
@@ -51,6 +51,18 @@
...
 
+def check_ticketing(user, modules):
+    if not user.is_verified:
+        raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
+                                 "Please verify your Email")
+    elif not modules.ticket_include:
+        raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
+                                 "Ticket modules not included.")
+    else:
+        raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
```
iamareebjamal: Sources are still wrong

app/api/events.py
```diff
@@ -53,13 +53,13 @@
...
-        raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
+        raise ForbiddenException({'source': '/data/attributes/is-verified'},
                                  "Please verify your Email")
     elif not modules.ticket_include:
-        raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
+        raise ForbiddenException({'source': '/data/attributes/ticket-include'},
                                  "Ticket modules not included.")
     else:
-        raise ForbiddenException({'source': '/data/attributes/is-ticketing-enabled'},
+        raise ForbiddenException({'source': '/data/attributes/location-name'},
```
iamareebjamal: This will run unconditionally

prateekj117: @iamareebjamal What source should I specify in that case. Ticketing is not enabled according to location and user-verifed condition only AFAI could find.

iamareebjamal: This is not about source. As you can see, the build has failed. You'll see the reason it failed is this line

prateekj117: @iamareebjamal As the location is been checked in the later stages of the before_post function, I think we should just discard the else statement.

iamareebjamal: If location is checked, then OK

Haider8:
@prateekj117 What will the error message if ```super-admin``` tries to ```Save draft``` with only ```event name``` ?

prateekj117:
@Haider8 I don't think we need to handle that case as this script is almost the same with some more descriptive messages. @iamareebjamal Your views on this?

iamareebjamal:
Expected Scenarios:

User is unverified: Show unverified error unconditionally. This should be handled by a decorator, but for now, let it be handled as it is

User is verified:
  Event request has tickets and tickets module is disabled: Show ticketing is disabled
  Event request does not have tickets or it has and ticketing module is enabled:
    Event request does not have location: Show location error

Nested conditions mean the above condition is pre-requisite

These flows are common and should be obvious. I don't know why the changes were made so that this message is shown to unverified user but some these validations could be handled by marshmallow. But since it has been days since the issue has been reported, let's close this PR ASAP

app/api/events.py
```diff
@@ -51,13 +51,14 @@
...
-def check_ticketing(user, modules):
+def check_ticketing(user, modules, data=None):
     if not user.is_verified:
         raise ForbiddenException({'source': '/data/attributes/is-verified'},
                                  "Please verify your Email")
-    elif not modules.ticket_include:
-        raise ForbiddenException({'source': '/data/attributes/ticket-include'},
-                                 "Ticket modules not included.")
+    else:
+        if data.get('is_ticketing_enabled', False) and not modules.ticket_include:
```
iamareebjamal: Should be elif

app/api/events.py
```diff
@@ -51,13 +51,14 @@
 from app.models.stripe_authorization import StripeAuthorization
 
 
-def check_ticketing(user, modules):
+def check_ticketing(user, modules, data=None):
     if not user.is_verified:
         raise ForbiddenException({'source': '/data/attributes/is-verified'},
```
iamareebjamal: Source is wrong, it should be empty

app/api/events.py
iamareebjamal: The condition is wrong, it should check `data.get('is_ticketing_enabled', False)`

Please check logic before pushing

app/api/events.py
```diff
@@ -461,9 +462,9 @@ def before_patch(self, args, kwargs, data=None):
         """
         user = User.query.filter_by(id=kwargs['user_id']).first()
         modules = Module.query.first()
-        check_ticketing(user, modules)
+        check_ticketing(user, modules, data)
         if data.get('can_pay_by_paypal', False) or data.get('can_pay_by_cheque', False) or \
-                data.get('can_pay_by_bank', False) or data.get('can_pay_by_stripe', False):
+            data.get('can_pay_by_bank', False) or data.get('can_pay_by_stripe', False):
```
iamareebjamal: revert

app/api/events.py
```diff
@@ -53,10 +53,10 @@
 
 def check_ticketing(user, modules, data=None):
     if not user.is_verified:
-        raise ForbiddenException({'source': '/data/attributes/is-verified'},
+        raise ForbiddenException({'source': ''},
                                  "Please verify your Email")
-    else:
-        if data.get('is_ticketing_enabled', False) and not modules.ticket_include:
+    elif user.is_verified:
+        if data.get('is_ticketing_enabled', True) and not modules.ticket_include:
```
iamareebjamal: Wrong, the if condition should be in elif

```python
if something:
  ...
else:
  if something_else:
    ...

```

is useless and same as

```python
if something:
  ...
elif something_else:
  ...

```

prateekj117:
@iamareebjamal Please review.

app/api/events.py
```diff
@@ -51,6 +51,15 @@
 from app.models.stripe_authorization import StripeAuthorization
 
 
+def check_ticketing(user, modules, data=None):
```
iamareebjamal: data is unused. Remove

iamareebjamal: data is compulsory, remove data=None

iamareebjamal: Rename appropriately, this is not just checking ticketing

app/api/events.py
```diff
@@ -126,7 +126,7 @@ def before_post(self, args, kwargs, data=None):
         """
         user = User.query.filter_by(id=kwargs['user_id']).first()
         modules = Module.query.first()
-        check_ticketing(user, modules, data)
+        check_create_event(user, modules, data)
         if data.get('can_pay_by_paypal', False) or data.get('can_pay_by_cheque', False) or \
```
iamareebjamal: Move all common conditions to the function

app/api/events.py
```diff
@@ -461,7 +461,7 @@ def before_patch(self, args, kwargs, data=None):
         """
         user = User.query.filter_by(id=current_identity.id).one()
         modules = Module.query.first()
-        check_ticketing(user, modules, data)
+        check_create_event(user, modules, data)
         if data.get('can_pay_by_paypal', False) or data.get('can_pay_by_cheque', False) or \
```
iamareebjamal: Like this one

app/api/events.py
```diff
@@ -51,7 +51,7 @@
 from app.models.stripe_authorization import StripeAuthorization
 
 
-def check_ticketing(user, modules, data=None):
+def check_create_event(user, modules, data):
```
iamareebjamal: Rename to `validate_event`

prateekj117:
@CosmicCoder96 Please review and merge.

PR #5719, opened 2019-03-27
fix : remove redundant is_verified for super admin

### LINKED ISSUE #5675: Super Admin is unverified user
**Describe the bug**

Admin/Users shows the super admin to be unverified

**Expected behavior**

It should simply be verified!

image

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Ubuntu 16.04 LTS
 - Browser [e.g. chrome, safari] Chrome

**Additional context**

Open for all!

uds5501:
Fixes #5675 

#### Checklist

Apparently the super user being created is using the `default` value of `False` (i can confirm that the first super user being created is always unverified) and removing this parameter fixes it.

- removing is_verified parameter from create_user.py

### DIFF
--- a/tests/all/integration/auth_helper.py
+++ b/tests/all/integration/auth_helper.py
@@ -14,7 +14,7 @@ def create_user(email, password, is_verified=True):
 
 
 def create_super_admin(email, password):
-    user = create_user(email, password, is_verified=True)
+    user = create_user(email, password)
     user.is_super_admin = True
     user.is_admin = True
     save_to_db(user, "User updated")


uds5501:
@iamareebjamal I had a query regarding this, why does the User model not have an `__init__` in the class definition like any other `SoftDeletionModel` ?

iamareebjamal:
I don't know

tests/all/integration/auth_helper.py
```diff
@@ -14,7 +14,7 @@ def create_user(email, password, is_verified=True):
 
 
 def create_super_admin(email, password):
-    user = create_user(email, password, is_verified=True)
```
CosmicCoder96: so what does this accomplish, Super Admin is a user, why can't it be verified by default?

CosmicCoder96: Also how is this related to the issue it solves. I'm confused, can you please elaborate.

uds5501: @CosmicCoder96 the `create_user` function has the `is_verified` argument to be true by default so passing it as a true for a super admin is more or less redundant.

As for how it solves the issue, kindly bear with me because it was more of an observation task :
I could not create a `verified` super-user on using `create_db.py` upon db initialisation but the subsequent super users being created by the same place are all verified as they should be.

On removing this `is_verified` argument from the function call, the first super admin being created is verified.

This weird behaviour is being shown in my system for quite a time now but it would be great if someone could verify it too

PR #5725, opened 2019-03-28
Fix Migration which incorrectly modified confirmed sessions

CosmicCoder96:
Any of the sessions which had a confirmed state were modified to be drafts while transitioning from legacy.

Fixes: #5726

### DIFF
--- a/migrations/versions/4defe72abed3_.py
+++ b/migrations/versions/4defe72abed3_.py
@@ -25,7 +25,7 @@ def upgrade():
     op.execute("""UPDATE events SET state = 'draft' WHERE state != 'published'""")
     op.execute("""UPDATE sessions SET state = LOWER(state)""")
     op.execute("""UPDATE sessions SET state = 'draft' WHERE state not in
-    ('accepted', 'pending', 'approved', 'rejected')""")
+    ('accepted', 'pending', 'confirmed', 'rejected')""")
     # ### end Alembic commands ###
 
 


migrations/versions/4defe72abed3_.py
```diff
@@ -25,7 +25,7 @@ def upgrade():
     op.execute("""UPDATE events SET state = 'draft' WHERE state != 'published'""")
     op.execute("""UPDATE sessions SET state = LOWER(state)""")
     op.execute("""UPDATE sessions SET state = 'draft' WHERE state not in
-    ('accepted', 'pending', 'approved', 'rejected')""")
+    ('accepted', 'pending', 'confirmed', 'rejected')""")
```
iamareebjamal: This will break the migrations of people who have already applied them, inclusing the running server on api.eventyay.com

CosmicCoder96: @iamareebjamal since this is a data OP we can’t make a new migration file to fix this as this file will already change the state to draft. I thought about Adding a new migration file which changes all the draft sessions to confirmed only to find out that legacy data has some sessions which are originally in the draft state. Not sure what their purpose is.  Is there a  correct approach to fix this old migration or should I just ignore those originally draft migrations and make them confirmed.

iamareebjamal: There's no way to change the applied migration. If the current production data is not important, this can be merged. But as mentioned earlier, this will not work on any DB which has this migration already applied. If that data is important, and DB cannot be re-migrated, then the only way is to create a new migration

CosmicCoder96: @iamareebjamal Data is not important as we are replacing the current production data in next.eventyay yet again with the legacy data from open-event-legacy.

iamareebjamal: OK

PR #5728, opened 2019-03-28
Make email ID's case insensitive

### LINKED ISSUE #5643: Email IDs are case-sensitive, allowing multiple accounts with one handle
**Describe the bug**
Currently, email id is stored in the database as raw string i.e. as entered by the user and then during auth the user is expected to enter the exact case-sensitive email id. This can lead to creation of multiple accounts with one email id and is a major bug.

**To Reproduce**
1. Go to login
2. Go to create account
3. Enter email id and password
4. Click Register
5. Again go to Create Account
6. Enter the same email id and password, however this time 
    change the casing of some letters of email
7. Click Register
8. You would have successfully created two different accounts with the same email id

**Expected behavior**
When we try to create the second account, it should give an error: Email already registered.

**Additional context**
I am working on it.

prateekj117:
Fixes #5643 

#### Checklist

Email IDs are case-sensitive, allowing multiple accounts with one handle

### DIFF
--- a/migrations/versions/7bb5891a9f2e_lowercase_email.py
+++ b/migrations/versions/7bb5891a9f2e_lowercase_email.py
@@ -1,0 +1,29 @@
+"""lowercase_email
+
+Revision ID: 7bb5891a9f2e
+Revises: 6e5c574cbfb8
+Create Date: 2019-03-30 12:51:48.134800
+
+"""
+
+from alembic import op
+from sqlalchemy import func
+import sqlalchemy as sa
+import sqlalchemy_utils
+
+
+# revision identifiers, used by Alembic.
+revision = '7bb5891a9f2e'
+down_revision = '6e5c574cbfb8'
+
+
+def upgrade():
+    op.execute("delete from users where lower(_email) <> _email;",
+               execution_options=None)
+    op.execute("create unique index users_lower_email on users (lower(_email));",
+               execution_options=None)
+
+
+def downgrade():
+    op.execute("drop index users_lower_email;",
+               execution_options=None)
--- a/migrations/versions/7bb5891a9f2e_lowercase_email.py
+++ b/migrations/versions/7bb5891a9f2e_lowercase_email.py
@@ -18,9 +18,20 @@ down_revision = '6e5c574cbfb8'
 
 
 def upgrade():
-    op.execute(" UPDATE users SET deleted_at = current_timestamp FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+    op.execute("UPDATE users SET deleted_at = current_timestamp FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+               execution_options=None)
+    op.execute("UPDATE users SET _email = concat(_email, '_') FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+               execution_options=None)
+    op.execute("update users set _email = lower(_email);",
+               execution_options=None)
+    op.execute("alter table users add check ( lower(_email) = _email );",
                execution_options=None)
 
 
 def downgrade():
-    pass
+    op.execute("UPDATE users SET deleted_at = null FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+               execution_options=None)
+    op.execute("UPDATE users SET _email = replace(_email, right(_email, 1), '') FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+               execution_options=None)
+    op.execute("alter table users add check ( lower(_email) = _email );",
+               execution_options=None)
--- a/migrations/versions/7bb5891a9f2e_lowercase_email.py
+++ b/migrations/versions/7bb5891a9f2e_lowercase_email.py
@@ -18,20 +18,16 @@ down_revision = '6e5c574cbfb8'
 
 
 def upgrade():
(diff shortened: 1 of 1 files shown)


prateekj117:
@iamareebjamal Getting this error when trying to create migrations: 
Screenshot from 2019-03-29 02-13-44

prateekj117:
@iamareebjamal Please review the changes made till now. 
After this, one problem after making index is that user is still allowed to register with email address not totally in lowercase. So, for that, I will see if there is any way to make it as a constraint.

iamareebjamal:
There's no lower case unique index

prateekj117:
@iamareebjamal Using something like check(translate) will do the job I think.

prateekj117:
https://stackoverflow.com/questions/37847618/postgres-text-column-that-only-allows-converts-to-lowercase-and-no-special-chara

prateekj117:
@iamareebjamal 
`ALTER TABLE users ADD CONSTRAINT users_email_lowercase_ck CHECK (_email = lower(_email));`

I think this will do.

iamareebjamal:
Yes, so add that

iamareebjamal:
Use citext for email column

https://stackoverflow.com/a/25438489

iamareebjamal:
Also, add migration to convert all values to lowercase after creating index

prateekj117:
@iamareebjamal Please review. Tested. Everything works as required.

migrations/versions/7bb5891a9f2e_lowercase_email.py
```diff
@@ -18,9 +18,20 @@
 
 
 def upgrade():
-    op.execute(" UPDATE users SET deleted_at = current_timestamp FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+    op.execute("UPDATE users SET deleted_at = current_timestamp FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+               execution_options=None)
+    op.execute("UPDATE users SET _email = concat(_email, '_') FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
```
iamareebjamal: This can be done in above statement only, no need to run the query again

migrations/versions/7bb5891a9f2e_lowercase_email.py
```diff
@@ -18,9 +18,20 @@
...
+               execution_options=None)
+    op.execute("update users set _email = lower(_email);",
+               execution_options=None)
+    op.execute("alter table users add check ( lower(_email) = _email );",
```
iamareebjamal: No need to add check

Again, there is no index here, which makes this migration useless. Use citext extension to add a case insensitive column

migrations/versions/7bb5891a9f2e_lowercase_email.py
```diff
@@ -18,9 +18,20 @@
...
                execution_options=None)
 
 
 def downgrade():
-    pass
+    op.execute("UPDATE users SET deleted_at = null FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
```
iamareebjamal: Won't work

migrations/versions/7bb5891a9f2e_lowercase_email.py
```diff
@@ -18,9 +18,20 @@
...
+               execution_options=None)
+    op.execute("UPDATE users SET _email = replace(_email, right(_email, 1), '') FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
```
iamareebjamal: Won't work

prateekj117:
@iamareebjamal Please review.

migrations/versions/7bb5891a9f2e_lowercase_email.py
```diff
@@ -18,7 +18,7 @@
 
 
 def upgrade():
-    op.execute("UPDATE users SET deleted_at = current_timestamp, _email = concat(_email, '_') FROM (SELECT lower(_email) FROM users GROUP BY lower(_email) HAVING COUNT(*) > 1) as email where lower(_email) != _email;",
+    op.execute("UPDATE users SET deleted_at = current_timestamp, _email = concat(_email, '_') where _email not in (SELECT DISTINCT ON (upper(_email)) _email FROM users);",
```
iamareebjamal: Query is correct but it deletes the users with lowercase names rather than uppercase

prateekj117: @iamareebjamal I tried to do it using Order by as you suggested, but in the distinct on query, I can only order by using upper(_email), which does not solve this issue.

iamareebjamal: Can you order by _email?

prateekj117: @iamareebjamal It shows error to order it by using upper(_email) only. Distinct on and order by are somewhat related.

iamareebjamal: OK, will see what can be done. Merging for now, query can be improved till when it will run on production DB

migrations/versions/7bb5891a9f2e_lowercase_email.py
```diff
@@ -0,0 +1,33 @@
...
+               execution_options=None)
+    op.execute("create extension citext;",
+               execution_options=None)
+    op.execute("alter table users alter column _email type citext;",
+               execution_options=None)
+
+
+def downgrade():
+    op.execute("alter table users alter column _email type text;",
```
iamareebjamal: Add an option of undeleting users and removing underscore from their emails

migrations/versions/7bb5891a9f2e_lowercase_email.py
prateekj117: @iamareebjamal Ok.

prateekj117:
@iamareebjamal Please review.

prateekj117:
@CosmicCoder96 Please review and merge.

PR #5812, opened 2019-04-17
fix: Create checks for event identifiers being passed as ids

### LINKED ISSUE #2811: Event does not show up on "manage events" page when it is a copy
When the user copies an event and edits it, it does not show up on the event management page.

screenshot from 2017-01-01 14-01-06

### LINKED ISSUE #5811: Co-organizer access required error on leaving sponsor step
**Describe the bug**

If I make an event using non-admin user, it can't leave sponsor form or go away from session form using previous and forward buttons .

**To Reproduce**
1. Create an event using non-admin user
2. try to move forward from sponsor step
3. See error

ezgif com-video-to-gif

**Expected behavior**

It should've moved forward without any error

**Stacktrace**

```cmd

INFO:werkzeug:127.0.0.1 - - [17/Apr/2019 11:33:47] "OPTIONS /v1/events/00784287 HTTP/1.1" 200 -
127.0.0.1 - - [17/Apr/2019 11:33:47] "PATCH /v1/events/00784287 HTTP/1.1" 403 -
INFO:werkzeug:127.0.0.1 - - [17/Apr/2019 11:33:47] "PATCH /v1/events/00784287 HTTP/1.1" 403 -

```
**Additional details (please complete the following information):**
 - OS: [e.g. MacOS, Ubuntu, CentOS] Ubuntu
 - Python Version [e.g. `3.5`, `3.6`] 3.5

**Additional context**

Working on it

uds5501:
Fixes #5811

#### Checklist

In the current situation, if an event identifier is completely integer, it is being passed as `{id : <identifier integer>` instead of normal `{identifier : <identifier string>}` in the kwargs.

So, in case an event has a numeric identifier, let's modify the incoming kwarg request and modify it into the usual format.

- Checking if the incoming kwargs are usual
- Reformatting them in case of integers

### DIFF
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -444,7 +444,7 @@ class EventDetail(ResourceDetail):
         :param view_kwargs:
         :return:
         """
-        kwargs = get_id(kwargs)
+        get_id(view_kwargs)
 
         if view_kwargs.get('order_identifier') is not None:
             order = safe_query(self, Order, 'identifier', view_kwargs['order_identifier'], 'order_identifier')
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -426,7 +426,13 @@ class EventDetail(ResourceDetail):
         :param kwargs:
         :return:
         """
+        if 'id' in kwargs:
+            # condition to check if id is actually numerical identifier
+            if len(str(kwargs['id'])) >= 5:
+                new_kwarg = {'identifier': str(kwargs['id'])}
+                kwargs = new_kwarg
         kwargs = get_id(kwargs)
+
         if 'Authorization' in request.headers and has_access('is_coorganizer', event_id=kwargs['id']):
             self.schema = EventSchema
         else:
@@ -438,11 +444,6 @@ class EventDetail(ResourceDetail):
         :param view_kwargs:
         :return:
         """
-        if 'id' in kwargs:
-            # condition to check if id is actually numerical identifier
-            if len(str(kwargs['id'])) >= 5:
-                new_kwarg = {'identifier': str(kwargs['id'])}
-                kwargs = new_kwarg
         kwargs = get_id(kwargs)
 
         if view_kwargs.get('order_identifier') is not None:
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -438,7 +438,12 @@ class EventDetail(ResourceDetail):
         :param view_kwargs:
         :return:
         """
-        get_id(view_kwargs)
+        if 'id' in kwargs:
+            # condition to check if id is actually numerical identifier
+            if len(str(kwargs['id'])) >= 5:
+                new_kwarg = {'identifier': str(kwargs['id'])}
+                kwargs = new_kwarg
+        kwargs = get_id(kwargs)
 
         if view_kwargs.get('order_identifier') is not None:
             order = safe_query(self, Order, 'identifier', view_kwargs['order_identifier'], 'order_identifier')


app/api/events.py
```diff
@@ -426,7 +426,13 @@ def before_get(self, args, kwargs):
         :param kwargs:
         :return:
         """
+        if 'id' in kwargs:
+            # condition to check if id is actually numerical identifier
+            if len(str(kwargs['id'])) >= 5:
```
iamareebjamal: Why and how?

app/api/events.py
```diff
@@ -426,7 +426,13 @@ def before_get(self, args, kwargs):
...
+                new_kwarg = {'identifier': str(kwargs['id'])}
```
iamareebjamal: This will strip all other arguments passed in kwargs

app/api/events.py
uds5501: Because when we are encountering the `event_id` transition from the frontend, it usually sends an identifier. But in case it's a complete integer, our api is being passed an `id` in the kwargs instead of `identifier` which messes up the `get_id` function.

So, if an id in the kwarg is a number and has a length of >=5 we can assume that it's actually an identifier being passed as an ID.
(at least that's what on my mind)

app/api/events.py
uds5501: @iamareebjamal Okay, that's true, I will fix this up

app/api/events.py
iamareebjamal: The current mechanism is very brittle. The whole moving required params through kwargs is a bad idea and source of bugs. In future, the ID itself may be of more than 5 digits, then this will create a new bug. Can this be fixed in an earlier layer?

uds5501: @iamareebjamal I agree.
 
I had a plan, to make sure our api assign only non numeric identifiers. `1b3452` kind? This would ensure that we don't get the error at all but I am not sure if it would be a valid fix.

iamareebjamal: This would be an OK workaround, but it won't solve errors for already generated identifiers

uds5501: @iamareebjamal Okay. I just want to know where should I look to figure out why is the api behaving this way and not taking only identifier kwargs.

iamareebjamal: I don't know where it's making those conversions but most likely it's flask-json-restapi

uds5501: Understood, I shall have a look!

uds5501:
@iamareebjamal I looked into it deeper and I found some disturbing sequences which I would like to report.

### Explicit Typecasting at relationship level is problematic

Currently the relationship parameters (the kwarg keys) are being determined by the `api/__init__.py` .

```python

api.route(EventDetail, 'event_detail', '/events/<int:id>', '/events/<identifier>'

```
In one trial run I decided to change this relationship to `/events/<string: id>` effectively typecasting the id and taking identifier out of the kwargs. and to use the existing mechanism, I used this segment to check if the passed kwarg is indeed an id or actually an identifier:

```python
try:
    identifiedEvent = Event.query.filter_by(identifier=kwargs['id']).first()
        flag = True
    except NoResultFound:
        flag = False
    if flag is True:
        kwargs['identifier'] = kwargs['id']
```

Now, the Event Detail permission check is working alright but immediately, about 67 other relationships break because the permission check is still using identifiers as id in all other cases.

```cmd
{'fetch': 'event_id', 'fetch_as': 'event_id', 'event_id': 42, 'model': <class 'app.models.ticket.Ticket'>} ('is_coorganizer',)
127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/tickets/4/order-statistics HTTP/1.1" 200 -
INFO:werkzeug:127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/tickets/4/order-statistics HTTP/1.1" 200 -
in check
{'fetch': 'id', 'fetch_as': 'event_id', 'event_id': 111121, 'model': <class 'app.models.event.Event'>} ('is_coorganizer',)
127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/events/111121/general-statistics HTTP/1.1" 403 -
INFO:werkzeug:127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/events/111121/general-statistics HTTP/1.1" 403 -
in check
{'fetch': 'id', 'fetch_as': 'event_id', 'event_id': 111111, 'model': <class 'app.models.event.Event'>} ('is_coorganizer',)
127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/events/111111/general-statistics HTTP/1.1" 403 -
INFO:werkzeug:127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/events/111111/general-statistics HTTP/1.1" 403 -
in check
{'fetch': 'id', 'fetch_as': 'event_id', 'event_id': 123456, 'model': <class 'app.models.event.Event'>} ('is_coorganizer',)
127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/events/123456/general-statistics HTTP/1.1" 403 -
INFO:werkzeug:127.0.0.1 - - [18/Apr/2019 11:32:05] "GET /v1/events/123456/general-statistics HTTP/1.1" 403 -
```
Like over here, 123456, 111111 and 111121 are identifiers being passed as ids.

iamareebjamal:
I'm curious as to why we need id and identifier both?

uds5501:
Good question. I think we need it because in front end we are handling the views with identifiers and not exposing the event is. `/e/<identifier>` . Also, the action buttons re-route using identifiers only

iamareebjamal:
Then we should provide a different endpoint for the identifier in the API, or it is inevitably going to clash. Let's think about this, add a separate endpoint for identifiers, slowly migrate frontend to using that endpoint and fixing bugs in the process and finally remove the clashing endpoint from the API. This is actual solution which will work long term and with existing identifiers

@CosmicCoder96 Please see

PR #5841, opened 2019-04-24
chore: Check if flask-json-rest master branch passes all tests

uds5501:
It currently resolves nothing, this is just a test to check if the miLibris/flask-rest-jsonapi -- master passes all the tests.

If it does, then we can go ahead and rebase the fossasia's branch @iamareebjamal

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -42,7 +42,7 @@ pytz
 diff-match-patch
 blinker
 envparse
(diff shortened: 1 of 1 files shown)


iamareebjamal:
It won't pass it. You have to take `shubhamp_master` branch of the fork and rebase it with upstream master and use that to see if it passes all tests

PR #5882, opened 2019-05-14
feat: ability to delete user's account by himself/herself

### LINKED ISSUE #5805: The ability to delete my account.
I had this app mistaken for something else, and I would like to delete my account. However, there doesn't seem to be an easy way to do that, and the only way to contact you according to your website (aside from social media, which I don't have) is Github.

Allow somebody to erase their account via a button on their Profile. Or better yet, just delete my account so I can forget about this already.

I don't know what else to do, this seems like a very basic function for an app to have.

**Additional context**

Thank you.

shreyanshdwivedi:
Fixes #5805 

#### Checklist

Currenlty, any user cannot delete his/her account

We allow someone who have not created any event or is not associated with any order to delete his/her account.

### DIFF
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -216,11 +216,9 @@ class UserDetail(ResourceDetail):
 
         if has_access('is_user_itself', user_id=user.id) and data.get('deleted_at') != user.deleted_at:
             if len(user.events) != 0:
-                raise ForbiddenException({'source': ''},
-                                        "Users associated with events cannot be deleted")
+                raise ForbiddenException({'source': ''}, "Users associated with events cannot be deleted")
             elif len(user.orders) != 0:
-                raise ForbiddenException({'source': ''},
-                                        "Users associated with orders cannot be deleted")
+                raise ForbiddenException({'source': ''}, "Users associated with orders cannot be deleted")
             else:
                 user.deleted_at = data.get('deleted_at')
 
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -214,6 +214,16 @@ class UserDetail(ResourceDetail):
         if has_access('is_admin') and data.get('deleted_at') != user.deleted_at:
             user.deleted_at = data.get('deleted_at')
 
+        if has_access('is_user_itself', user_id=user.id) and data.get('deleted_at') != user.deleted_at:
+            if len(user.events) != 0:
+                raise ForbiddenException({'source': ''},
+                                        "Users associated with events cannot be deleted")
+            elif len(user.orders) != 0:
+                raise ForbiddenException({'source': ''},
+                                        "Users associated with orders cannot be deleted")
+            else:
+                user.deleted_at = data.get('deleted_at')
+
         if users_email is not None and users_email != user.email:
             try:
                 db.session.query(User).filter_by(email=users_email).one()


shreyanshdwivedi:
@uds5501 @mrsaicharan1 @iamareebjamal please review

app/api/users.py
```diff
@@ -214,6 +214,16 @@ def before_update_object(self, user, data, view_kwargs):
         if has_access('is_admin') and data.get('deleted_at') != user.deleted_at:
             user.deleted_at = data.get('deleted_at')
 
+        if has_access('is_user_itself', user_id=user.id) and data.get('deleted_at') != user.deleted_at:
```
uds5501: @shreyanshdwivedi What's the second condition for? If a user is going to be deleted, he simply will have a `deleted_at=None` right?

shreyanshdwivedi: @uds5501 I've just followed the simple check which was used on deleting a user's account by admin. Just for consistency.
Though value of `deleted_at` attribute will be `None`.

uds5501: Okay!

uds5501: @shreyanshdwivedi Fix the hound violations though

shreyanshdwivedi: @uds5501 I don't see any violation in my updated file. I already made changes related to last 2 violations.
I also see 2 violations written beside Hound in checks but I don't see any now. Please point me if you see any violations

app/api/users.py
```diff
@@ -214,6 +214,16 @@ def before_update_object(self, user, data, view_kwargs):
...
+            if len(user.events) != 0:
+                raise ForbiddenException({'source': ''},
+                                        "Users associated with events cannot be deleted")
```
uds5501: This could be a violation @shreyanshdwivedi

app/api/users.py
```diff
@@ -214,6 +214,16 @@ def before_update_object(self, user, data, view_kwargs):
...
+            elif len(user.orders) != 0:
+                raise ForbiddenException({'source': ''},
+                                        "Users associated with orders cannot be deleted")
```
uds5501: Also this

app/api/users.py
```diff
@@ -214,6 +214,14 @@ def before_update_object(self, user, data, view_kwargs):
         if has_access('is_admin') and data.get('deleted_at') != user.deleted_at:
             user.deleted_at = data.get('deleted_at')
 
+        if has_access('is_user_itself', user_id=user.id) and data.get('deleted_at') != user.deleted_at:
+            if len(user.events) != 0:
```
iamareebjamal: This is wrong. User associated with active events can't be deleted. Same with orders

shreyanshdwivedi: @iamareebjamal when we discussed this earlier, we were trying to fathom the scams which an individual can do. We also concluded these possibilities -
 - User can create an event and do not show up on the venue and deletes his/her account just after that. This should not be the case 
 - User buys free tickets for an event and does not show up. Similarly deletes his/her account just after the event.

As we were discussing this we concluded to restrict this functionality only for users who have no events or orders associated as a first step. If we change the condition and only restrict users with active events/orders to delete their account, how can we avoid above menti scams?

iamareebjamal: An event cannot be deleted if it has active orders - and hence a user can't delete its account. I'm approving this. But need to evaluate and ensure that an order can never be deleted by a user

app/api/users.py
iamareebjamal: `data.get('deleted_at') != user.deleted_at`

I don't understand this

shreyanshdwivedi: @iamareebjamal When a user's account is deleted, the `deleted_at` property is updated. So, this check ensures that the code-block below this condition is executed only when `deleted_at` is updated. Otherwise the code-block will run every-time we update any detail of the user.

app/api/users.py
shreyanshdwivedi: Currently, we are not allowing any user to delete his/her orders. @iamareebjamal

app/api/users.py
iamareebjamal: > When a user's account is deleted, the deleted_at property is updated

From where. I send a DELETE request on User endpoint, then what happens?

shreyanshdwivedi: Then request to update the user's details is sent. Idk exactly from where but then the code-base in `before_update_object` in `api/users.py` is called and we update the `deleted_at` attribute of user there. We've been doing same when user was deleted from `admin` too.

iamareebjamal: `data.get('deleted_at')` This means it is being sent from user if I'm not wrong? Where does the deleted_at in data come from. Not that your implementation is wrong, but I'm trying to understand the process here

shreyanshdwivedi: The user has a `deleted_at` attribute. When the user is not deleted, `deleted_at` is `None` by default.
When we send a DELETE request by `destroyRecord` function on front-end, `deleted_at` data is updated to current timestamp and sent on server to update it in db. This is where this code-block comes into work. This is what I understood :sweat_smile:

iamareebjamal: That is really really bad if it is sent from frontend. Please create an issue to mitigate it ASAP

iamareebjamal: I'll approve and merge once this is created

shreyanshdwivedi: Maybe @CosmicCoder96 can help us in understanding what exactly is going on at the Front-end as I'm not really sure about this. Once confirmed by him, I'll open an issue regarding the same

CosmicCoder96: @shreyanshdwivedi Not sure if I fully get the issue, but frontend doesn't supply deleted at upon deletion server assigns that.

> When we send a DELETE request by destroyRecord function on front-end, deleted_at data is updated to current timestamp and sent on the server to update it in DB

The second part is not true, deleted_at is not sent by the frontend, rather at the server itself.

iamareebjamal: Yes, I wanted to understand when is deleted_at assigned in data? Because `data.get('deleted_at') != user.deleted_at` seems dubious. This will be true even when undeleting the object as well. Why is there no hook of delete request and why are we doing every check in `before_update_object`, this breaks single responsibility principle, but anyway, from where does `deleted_at` come in data?

CosmicCoder96: However @iamareebjamal @shreyanshdwivedi when a user is restored, frontend does sets the value of `deleted_at` to null again.

iamareebjamal: That is fine @CosmicCoder96

CosmicCoder96: > Why is there no hook of delete request
> from where does deleted_at come in data?

I think its handled in the library itself hence no delete hook exists.
```python
@check_method_requirements
    def delete(self, *args, **kwargs):
        """Delete an object
        """
        self.before_delete(args, kwargs)
        obj = self._data_layer.get_object(kwargs, get_trashed=(request.args.get('permanent') == 'true'))
        if obj is None:
            raise ObjectNotFound({'pointer': ''}, 'Object Not Found')
        if 'deleted_at' not in self.schema._declared_fields or request.args.get('permanent') == 'true' or current_app.config['SOFT_DELETE'] is False:
            self._data_layer.delete_object(obj, kwargs)
        else:
            data = {'deleted_at': str(datetime.now(pytz.utc))}
            self._data_layer.update_object(obj, data, kwargs)

        result = {'meta': {'message': 'Object successfully deleted'}}
        self.after_delete(result)
        return result
```
In `flask-rest-jsonapi/flask_rest_jsonapi/resource.py`.

iamareebjamal: OK. Thanks @CosmicCoder96. Let's merge this and move discussion to a separate issue

shreyanshdwivedi:
@iareebjamal can you please merge this. I've to work on other issue related to this 😅

PR #5884, opened 2019-05-15
fix: Send order email post completion

### LINKED ISSUE #5883: Email not send for Ticket Orders
Users don't receive an email with their order and PDF.

Expected: After a user orders a ticket, they should receive a PDF. If the attendee is a different person with a different email, the attendee should also receive an email.

CosmicCoder96:
Fixes #5883 

The ticketing mails were not being sent out because frontend initially creates the order in the pending state, and then after payment, order is `completed`, hence the instructions to mail should be in the `after_create` hook.

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -307,6 +307,15 @@ class OrderDetail(ResourceDetail):
             # delete the attendees so that the tickets are unlocked.
             delete_related_attendees_for_order(order)
 
+        elif order.status == 'completed':
+            send_email_to_attendees(order, current_user.id)
+            send_notif_to_attendees(order, current_user.id)
+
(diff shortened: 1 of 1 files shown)


app/api/orders.py
```diff
@@ -307,6 +307,15 @@ def after_update_object(self, order, data, view_kwargs):
             # delete the attendees so that the tickets are unlocked.
             delete_related_attendees_for_order(order)
 
+        if order.status == 'completed':
```
uds5501: @CosmicCoder96 make this an `elif` condition maybe? Rest all looks great

iamareebjamal:
```suggestion
        elif order.status == 'completed':
```

PR #5892, opened 2019-05-16
fix: Multiple redirections for tickets

### LINKED ISSUE #5879: Tickets endpoint results in too many redirects in production.
```python
            try:
                response = make_response(send_file(file_path))
                response.headers['Content-Disposition'] = 'attachment; filename=ticket-%s.pdf' % order_identifier
                return response
            except FileNotFoundError:
                create_pdf_tickets_for_holder(order)
                return redirect(url_for('ticket_blueprint.ticket_attendee_authorized', order_identifier=order_identifier))
```

The redirect call seems to be the problem, removing it and manually sending the response seems to fix it.

mrsaicharan1:
Fixes #5879 

#### Checklist

- Added function call to mitigate the multiple redirections issue in production

- created a `return_tickets` function to return tickets


app/api/auth.py
```diff
@@ -294,12 +299,10 @@ def ticket_attendee_authorized(order_identifier):
...
             file_path = '../generated/tickets/{}/{}/'.format(key, generate_hash(key)) + order_identifier + '.pdf'
             try:
-                response = make_response(send_file(file_path))
-                response.headers['Content-Disposition'] = 'attachment; filename=ticket-%s.zip' % order_identifier
-                return response
+                return return_tickets(file_path, order_identifier)
             except FileNotFoundError:
                 create_pdf_tickets_for_holder(order)
-                return redirect(url_for('ticket_blueprint.ticket_attendee_authorized', order_identifier=order_identifier))
+                return return_tickets(file_path, order_identifier)
```
iamareebjamal: Wouldn't this still result in an infinite loop

mrsaicharan1: If the tickets are found, the pdf `response` is returned.
If the tickets are not found, the tickets are generated and the pdf `response` is returned.
Previously, the redirection which was triggered had to hit the whole route again, which resulted in re-execution of the whole function associated with the route.

iamareebjamal: Sorry, I missed this line - `create_pdf_tickets_for_holder(order)`

PR #5908, opened 2019-05-16
feat: implement organiser transfer endpoint

### LINKED ISSUE #5890: Transfer the ownership of your events to another organizer.

We are implementing feature to provide ability to a user to delete his/her account. If the user is associated with any event, he/she must be given choice to transfer the events to other organizers associated with the event. (Just like in GitHub)

Something similar to `Github's Danger Zone` 
danger

**Additional context**

Frontend issue: https://github.com/fossasia/open-event-frontend/issues/3012

uds5501:
Fixes #5890

#### Checklist

- Delete original user organiser role
- Promote co-organiser to organiser
- Return a success message

### DIFF
--- a/app/api/roles.py
+++ b/app/api/roles.py
@@ -8,7 +8,14 @@ from app.models.role import Role
 from app.models.role_invite import RoleInvite
 from app.models.users_events_role import UsersEventsRoles
 from app.api.helpers.exceptions import UnprocessableEntity
+from app.api.helpers.errors import NotFoundError, ServerError
+from sqlalchemy.orm.exc import NoResultFound
+from flask import Blueprint, request, jsonify
+from app.models.users_events_role import UsersEventsRoles as UER
+from app.models.role import Role
+from app.api.helpers.db import save_to_db
 
+role_misc_routes = Blueprint('role_misc', __name__, url_prefix='/v1')
 
 class RoleList(ResourceList):
     """
@@ -74,3 +81,45 @@ class RoleDetail(ResourceDetail):
                   'methods': {
                       'before_get_object': before_get_object
                   }}
+
+
+@role_misc_routes.route('/change-organiser', methods=['POST'])
+def change_organiser():
+    current_organiser_id = request.json.get('current_org_id', None)
+    new_organiser_id = request.json.get('new_org_id', None)
+    event_id = request.json.get('event_id', None)
+
+    # Check if the user is actually an organiser of the event
+    # Delete the User Event Role
+    try:
+        role_entry_organiser_to_delete = UER.query.filter_by(user_id=current_organiser_id,
+                                                            event_id=event_id,
+                                                            role_id=1).one()
+        db.session.delete(role_entry_organiser_to_delete)
+
+    except NoResultFound:
+        return NotFoundError({'source': ''}, 'Organiser Role Entry not found').respond()
+
+    except Exception as e:
+        return ServerError({'source': ''}, e).respond()
+
+    # Check if the coorganiser is not found but the earlier entry has been deleted
+    # Rollback the last commit in case of exception to restore organiser status
+    try:
+        role_entry_coorg_to_organiser = UER.query.filter_by(user_id=new_organiser_id,
+                                                            event_id=event_id,
+                                                            role_id=2).one()
+        role_entry_coorg_to_organiser.role_id = 1
+        save_to_db(role_entry_coorg_to_organiser)
+
+    except NoResultFound:
+        db.session.rollback()
+        return NotFoundError({'source': ''}, 'Coorganiser Role Entry not found').respond()
+
+    # Handle Unexpected Errors
+    except Exception as e:
+        return ServerError({'source': ''}, e).respond()
+
+    return jsonify({
+        "message": 'Success'
+    })
--- a/app/api/roles.py
+++ b/app/api/roles.py
@@ -12,7 +12,6 @@ from app.api.helpers.errors import NotFoundError, ServerError
 from sqlalchemy.orm.exc import NoResultFound
 from flask import Blueprint, request, jsonify
 from app.models.users_events_role import UsersEventsRoles as UER
-from app.models.role import Role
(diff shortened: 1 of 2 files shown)


uds5501:
@shreyanshdwivedi @iamareebjamal 
Currently I have implemented this endpoint straightaway using organiser and co organiser id's. In case the user will send the email id of the co organiser, I will modify this PR accordingly. Otherwise, Please review

app/api/roles.py
```diff
@@ -74,3 +80,45 @@ def before_delete_object(self, obj, kwargs):
                   'methods': {
                       'before_get_object': before_get_object
                   }}
+
+
+@role_misc_routes.route('/change-organiser', methods=['POST'])
+def change_organiser():
+    current_organiser_id = request.json.get('current_org_id', None)
+    new_organiser_id = request.json.get('new_org_id', None)
+    event_id = request.json.get('event_id', None)
```
iamareebjamal: Please use proper marshmallow schema and handle it as other endpoints are handled. This route is not following JSON API spec

uds5501:
@iamareebjamal Please review now?

app/api/roles.py
```diff
@@ -74,3 +80,45 @@ def before_delete_object(self, obj, kwargs):
...
+
+    except NoResultFound:
+        db.session.rollback()
+        return NotFoundError({'source': ''}, 'Coorganiser Role Entry not found').respond()
+
+    # Handle Unexpected Errors
+    except Exception as e:
+        return ServerError({'source': ''}, e).respond()
+
+    return jsonify({
```
uds5501: @shreyanshdwivedi Could you suggest how would a successful transaction be handled on the frontend?
Also, we need to open a new issue on front end to implement an option to transfer organiser

app/api/roles.py
```diff
@@ -74,3 +80,45 @@ def before_delete_object(self, obj, kwargs):
                   'methods': {
                       'before_get_object': before_get_object
                   }}
+
+
+@role_misc_routes.route('/change-organiser', methods=['POST'])
+def change_organiser():
+    current_organiser_id = request.json['data']['current_org_id']
+    new_organiser_id = request.json['data']['new_org_id']
+    event_id = request.json['data']['event_id']
```
iamareebjamal: You're still not using marshmallow schema. There's a reason we use it

app/api/roles.py
```diff
@@ -74,3 +80,45 @@ def before_delete_object(self, obj, kwargs):
                   'methods': {
                       'before_get_object': before_get_object
                   }}
+
+
+@role_misc_routes.route('/change-organiser', methods=['POST'])
```
mrsaicharan1: Instead of creating a separate route/blueprint, to follow the JSON-api spec, please use the marshmallow schema like this:
`def after_update_object(self, role, data, view_kwargs):
        """
        Method to update object
        :param role:
        :param data:
        :param view_kwargs:
        :return:
        """
        try:
	        role_entry_organiser_to_delete = UER.query.filter_by(user_id=current_organiser_id,
	                                                             event_id=event_id,
	                                                             role_id=1).one()
	        db.session.delete(role_entry_organiser_to_delete)
	
	    except NoResultFound:
	        return NotFoundError({'source': ''}, 'Organiser Role Entry not found').respond()
	
	    except Exception as e:
	        return ServerError({'source': ''}, e).respond()
	`

app/api/roles.py
shreyanshdwivedi: Okay. So for frontend what we'll do is we'll send a update request to change organizer email, which will retrieve user ID of new organizer and update the db and send him/her an email regarding the change to both the new and old organizer. We'll have email messages too.
We'll have to discuss this extensively.

app/api/roles.py
iamareebjamal:
```python
except Exception as e:
    return ServerError({'source': ''}, e).respond()
```

We should stop doing that ASAP. Bare exceptions should never be caught. They should be thrown

Because of this, a 500 error returns from API and does not get logged or tracked in Sentry

uds5501:
@iamareebjamal @shreyanshdwivedi now that frontend loader works, there are certain issues which I need to address here. We can't seem to get the `new-owner` attribute in our before_path data.

The data being sent is : 
```js
data: {
  attributes: {
    'new-owner': data.newOrganizerEmail,
    'starts-at'    : event.startsAt,
    'ends-at'      : event.endsAt,
    'timezone'     : event.timezone
  },
  id: event.originalId, 
  type: 'event',
  identifier: event.id
}
```
and being sent as a `PATCH` request to `events/${event.id}`
In the before_path or before_update_object hook, the data I can retrieve is everything but `new owner`.

There was one more approach @shreyanshdwivedi  suggested, straightaway modifying the organiser by patching `events/id/relationships/organiser` but for that we need the user_id of this new organiser which is again, bringing us to the same loop.

I was working with marshmallow schema validations for a while and learnt that the orgiinal data can be retrieved there. Will it be fine to implement it there if not a new custom endpoint?

iamareebjamal:
You are using type: rvent and the data you are sending is not of type event

You have to create a new schema and use that. We will either go with JSONAPI route or plain JSON route. We can't be doing plain first and then go to JSONAPI for it. We don't have time. 

Let's discuss it with the team and finalize. I have certain pros and cons for both.

shreyanshdwivedi:
@iamareebjamal new schema, as in? You mean I've to add new fields to the existing schema? Such as `new-owner` field?

iamareebjamal:
No, create a new Marshmallow schema and new jsonapi route

shreyanshdwivedi:
@iamareebjamal I and @uds5501 are stuck here. Can you please help?

iamareebjamal:
Did you read the docs?

https://flask-rest-jsonapi.readthedocs.io/en/latest/quickstart.html

Add a schema for JSON structure you want to take as input

https://flask-rest-jsonapi.readthedocs.io/en/latest/logical_data_abstraction.html

And handle the request in post handler and return appropriate response

PR #5931, opened 2019-05-19
feat: Implement locking a session feature by organizer

### LINKED ISSUE #5930: Implement locking a session feature by organizer

This feature will lock a session and associated speaker which will prevent her/him from further editing a session. It'll involve adding a new attribute `isLocked` to session which will be initially set to false. And handling a PATCH request to update it alonwith new dedicated error messages

**Additional context**

Parent issue - https://github.com/fossasia/open-event-frontend/issues/2958
Working on it

shreyanshdwivedi:
Fixes #5930 
Related PR : https://github.com/fossasia/open-event-frontend/pull/2964

#### Checklist

This feature will lock a session and associated speaker which will prevent her/him from further editing a session. It'll involve adding a new attribute isLocked to session which will be initially set to fals

 - Adds `is_locked` to session schema
 - Adds a migration file to update db
 -  Adds a check to raise error if locked session is edited


shreyanshdwivedi:
@uds5501 @CosmicCoder96 @pradeepgangwar @kushthedude @mrsaicharan1  please review

app/api/sessions.py
```diff
@@ -130,6 +130,17 @@ def before_get_object(self, view_kwargs):
...
 
+    def before_update_object(self, session, data, view_kwargs):
+        """
+        before update method to verify if session is locked before updating session object
+        :param event:
+        :param data:
+        :param view_kwargs:
+        :return:
+        """
+        if session.is_locked and data.get('is_locked') == session.is_locked:
```
uds5501: I am a bit sceptical about this condition, does an already unlocked session gets locked after this?

shreyanshdwivedi: No. This condition is to raise error if a locked session is edited anyhow.
I added this condition `data.get('is_locked') == session.is_locked` to `session.is_locked` because if I used `session.is_locked` alone, error was raised even when I changed state of session's is_locked, which makes sense.

uds5501: @shreyanshdwivedi Okay, makes sense. But wouldn't toggling the lock button go through this and hence raise a forbidden error upon unlocking?

shreyanshdwivedi: That's why I used `data.get('is_locked') == session.is_locked` because when we'll be toggling the is_locked property, this condition will never be satisfied @uds5501

uds5501: @shreyanshdwivedi Understood.

app/api/sessions.py
```diff
@@ -130,6 +130,17 @@ def before_get_object(self, view_kwargs):
...
+            raise ForbiddenException({'source': ''}, "Locked sessions cannot be edited")
```
CosmicCoder96: Point the source to the `isLocked` property like it is done for marshmallow validations.

shreyanshdwivedi: @CosmicCoder96 I've updated it. Please review

shreyanshdwivedi:
@CosmicCoder96 @iamareebjamal please review

iamareebjamal:
But this only adds the attribute. Where's the logic?

shreyanshdwivedi:
@iamareebjamal we don't actually need any logic here. Front-end PR is handling other implementation - https://github.com/fossasia/open-event-frontend/pull/2964.
We only need to add the `is_locked` attribute and raise error if a locked session is being edited here.

app/api/sessions.py
iamareebjamal: If `data.get('is_locked')` is None, this will never be true and error won't be raised. Logic is wrong

Also, what's the point of locking a session if you can unlock it and then edit. Shouldn't the unlocking only be allowed for an admin?

shreyanshdwivedi:
@iamareebjamal @CosmicCoder96  I think `data.get(is_locked)` will never be `None` as I've specifically added a check on the schema that it cannot be nullable and by default, the `is_locked` attribute will be false.

Also, I've just added a check which will ensure that `is_locked` can be updated only by admins or organizers. Please review

app/api/sessions.py
```diff
@@ -130,6 +131,21 @@ def before_get_object(self, view_kwargs):
...
+        """
+        before update method to verify if session is locked before updating session object
+        :param event:
+        :param data:
+        :param view_kwargs:
+        :return:
+        """
+        if data.get('is_locked') != session.is_locked:
+            if not (has_access('is_admin') or has_access('is_organizer')):
+                raise ForbiddenException({'source': ''}, "Access Forbidden")
```
iamareebjamal: Add proper source and message should be, 'You don't have enough permissions to change this property'

shreyanshdwivedi:
@CosmicCoder96 please review

shreyanshdwivedi:
@CosmicCoder96 please review this. Corresponding frontend PR is already merged

PR #5948, opened 2019-05-22
fix: Use SQLAlchemy in-built timestamp function

### LINKED ISSUE #5943: Timestamp saved in Database is not correct
**Describe the bug**
Timestamp saved in `created_at` column of users field is not correct.

Screenshot from 2019-05-21 20-40-02

Here, all five users were created at different times.

**Expected behavior**
Timestamp should be correct.

**Additional context**
Working on it.

prateekj117:
Fixes #5943 

#### Checklist

Pytz function used to save same timestamp in Database in case of subsequent requests. 

Use Inbuild SqlAlchemy Function for saving Timestamp


app/models/user.py
```diff
@@ -77,7 +78,7 @@ class User(SoftDeletionModel):
     is_verified = db.Column(db.Boolean, default=False)
     was_registered_with_order = db.Column(db.Boolean, default=False)
     last_accessed_at = db.Column(db.DateTime(timezone=True))
-    created_at = db.Column(db.DateTime(timezone=True), default=datetime.now(pytz.utc))
+    created_at = db.Column(db.DateTime(timezone=True), default=func.now())
```
iamareebjamal: Wow. If I understand it correctly, this is really bad. Because it will always use the same value for created at, which will be the datetime when the server started and this value got initialised. Please confirm if that is the case.

Is this present everywhere else as well

prateekj117: @iamareebjamal No, this is not the case for sure. I tested it several times myself. Plus, this is not what this function does at all. It's used when you want a default timestamp with (timezone=true). Other way was pytz, but it was showing that absurd behaviour.

iamareebjamal: I'm saying that it was doing this behaviour previously. Please revert and confirm that each time created_at will be same and only change when you restart the server

prateekj117: @iamareebjamal This is not true because last_accessed_at uses the same old function, and it's working fine.

iamareebjamal: It's not using the same function `last_accessed_at = db.Column(db.DateTime(timezone=True))`

prateekj117: @iamareebjamal So, utcnow is saving timestamp adding 1 hour.(https://stackoverflow.com/questions/28573873/python-datetime-utcnow-returning-incorrect-datetime) I think it should be replaced there as well.

iamareebjamal: Replace where? Where else is it used?

prateekj117: @iamareebjamal Used in session table as well. I am looking into other places as well.

iamareebjamal: Replace all places where default is set to datetime.now

prateekj117: @iamareebjamal Ok. That's all.

app/models/session.py
```diff
@@ -46,7 +47,7 @@ class Session(SoftDeletionModel):
     event_id = db.Column(db.Integer, db.ForeignKey('events.id', ondelete='CASCADE'))
     creator_id = db.Column(db.Integer, db.ForeignKey('users.id', ondelete='CASCADE'))
     state = db.Column(db.String, default="pending")
-    created_at = db.Column(db.DateTime(timezone=True), default=datetime.datetime.utcnow)
+    created_at = db.Column(db.DateTime(timezone=True), default=dateandtime.now())
```
iamareebjamal: Why different from func.now()?

prateekj117: Because func imported from sqlAlchemy is been used in the file already. And  I had to import func from sqlAlchemy.sql, but I imported it using `as`.

app/models/session.py
```diff
@@ -2,6 +2,7 @@
 
 import pytz
 from sqlalchemy import event, func
+from sqlalchemy.sql import func as dateandtime
```
iamareebjamal:
```suggestion
from sqlalchemy.sql import func as sql_func
```

iamareebjamal:
It isn't anywhere else?

prateekj117:
@iamareebjamal Yes.

PR #5997, opened 2019-06-04
fix: Correct 'days' attribute not found in expire tickets

### LINKED ISSUE #5996: No attribute 'days' error in expiring pending tickets after 3 days function
**Describe the bug**
After 3 days of a pending order, when trying to expire it throws error.

**Expected behavior**
Should not throw any error.

**Error Screenshot**
Screenshot from 2019-06-04 10-34-19

**Additional context**
Working on it.

prateekj117:
Fixes #5996 

#### Checklist

Corrects 'days' attribute not found in scheduled_jobs.py

### DIFF
--- a/app/api/helpers/scheduled_jobs.py
+++ b/app/api/helpers/scheduled_jobs.py
@@ -146,6 +146,6 @@ def expire_pending_tickets_after_three_days():
     from app import current_app as app
     with app.app_context():
         db.session.query(Order).filter(Order.status == 'pending',
-                                       (Order.created_at + datetime.timedelta(days=3)) > datetime.datetime.now()).\
+                                       (Order.created_at + datetime.timedelta(days=3)) <= datetime.datetime.now()).\
                                        update({'status': 'expired'})
         db.session.commit()
--- a/app/api/helpers/scheduled_jobs.py
+++ b/app/api/helpers/scheduled_jobs.py
@@ -12,9 +12,9 @@ from app.api.helpers.query import get_upcoming_events, get_user_event_roles_by_r
 from app.api.helpers.utilities import monthdelta
 from app.models import db
 from app.models.event import Event
-from app.models.session import Session
 from app.models.event_invoice import EventInvoice
 from app.models.order import Order
+from app.models.session import Session
 from app.models.ticket import Ticket
 from app.models.ticket_fee import get_fee
 from app.settings import get_settings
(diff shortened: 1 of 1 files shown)


iamareebjamal:
@uds5501 @shreyanshdwivedi @mrsaicharan1 Please review

app/api/helpers/scheduled_jobs.py
```diff
@@ -146,6 +146,6 @@ def expire_pending_tickets_after_three_days():
     from app import current_app as app
     with app.app_context():
         db.session.query(Order).filter(Order.status == 'pending',
-                                       (datetime.datetime.today() - Order.created_at).days > 3).\
+                                       (Order.created_at + datetime.timedelta(days=3)) > datetime.datetime.now()).\
```
mrsaicharan1: Could you explain what this does here? @prateekj117

uds5501: @prateekj117  just wanna ask, wouldn't `order-today` be actually negative? Also, please make it `>=` it will be a better choice

mrsaicharan1: @prateekj117 `(datetime.datetime.today() - Order.created_at).days > datetime.timedelta(days=3)` would've sufficed.

prateekj117: @mrsaicharan1 the main problem was with the days attribute. I don't get it how this condition solves it. 
Refer [this](https://stackoverflow.com/questions/28858791/sqlalchemy-filter-by-comparing-datetime-now-and-column-default-date) for more info on this.

prateekj117: Also, using .today() method will take time to be 12:00 of that particular day, we should use .now() instead.

mrsaicharan1: https://docs.python.org/3/library/datetime.html#timedelta-objects
I meant to say this 
```
datetime.datetime.today().day - Order.created_at.day > 3
```

mrsaicharan1: Also, your condition is wrong.
Suppose I placed an order yesterday. Won't it be expired?

mrsaicharan1: Also, your condition seems to be wrong.
Suppose I placed an order yesterday. Won't it be expired?

mrsaicharan1: https://docs.python.org/3/library/datetime.html#timedelta-objects
This would be a better solution
```
datetime.datetime.now().day - Order.created_at.day > 3
```

prateekj117: @mrsaicharan1 I read on stackoverflow that there might be a case that python datetime isn't compatible with Sqlalchemy datetime object. So I implemented it this way which shows on stackoverflow that it works. Silly mistake on the opposite equality sign part, will correct it.

mrsaicharan1: yes @prateekj117, there might be cases , but we are comparing `days`.Difference will never change. But anyways, if you change the sign, your fix works too.

PR #6000, opened 2019-06-05
fix: adds unique constraint on role_invite

### LINKED ISSUE #5986: Multiple role invitations to the same email are allowed and are stored multiple times
**Describe the bug**
Currently, multiple role invitations can be sent to the same email id and even invitations with the same email and same role can be sent more than once and it is being stored multiple times in the list of roles. 

**Expected behavior**
There can be an option to resend a pending invitation and that should not be stored as a duplicate entry.

shreyanshdwivedi:
Fixes #5986 

#### Checklist

Invitations with the same email and same role can be sent more than once and it is being stored multiple times in the list of roles.

 - Raises error when duplicate role invite is created


shreyanshdwivedi:
@iamareebjamal you mean this -
`  __table_args__ = (UniqueConstraint('email', 'role_id', 'event_id',name='email_role_event_uc'), )` ?
I tested it locally but the postman didn't raise error. However will check again.
Maybe @uds5501 @kushthedude  can test locally too?

iamareebjamal:
You are checking uniqueness on 3 attributes, email, role and event_id whereas the Unique Constraint has 4 attributes, that's why it's failing

shreyanshdwivedi:
@iamareebjamal actually I followed the implementation from here -
https://github.com/fossasia/open-event-server/blob/3b6632114d0214b10b4467cb79d99ed55704aab0/app/models/custom_form.py#L71

Maybe this needs to be fixed too.

iamareebjamal:
I'm sorry, I was wrong, the unique constraint is correct. You just haven't created the migration. How is the DB going to know about the constraint?

shreyanshdwivedi:
@iamareebjamal thanks a lot for help. I've fixed it.
Please review :)

migrations/versions/89a5b836b406_.py
```diff
@@ -0,0 +1,37 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute(
+        """DELETE FROM role_invites R1
+        USING role_invites R2
+        WHERE R1.id < R2.id
+        AND R1.email = R2.email
+        AND R1.role_id = R2.role_id
+        AND R1.event_id = R2.event_id
```
iamareebjamal: Uhh, what?

iamareebjamal: Removing duplicates, right? 

So are there any foreign keys in role_invites?

shreyanshdwivedi: I'm deleting all the past duplicate entries here and tested on my local too. When I didn't add this delete condition the upgradation raised error due to presence of duplicate entries. Is the delete statement wrong? @iamareebjamal

shreyanshdwivedi: Yes @iamareebjamal. `event_id` and `role_id` are foreignKeys
https://github.com/fossasia/open-event-server/blob/3b6632114d0214b10b4467cb79d99ed55704aab0/app/models/role_invite.py#L23

iamareebjamal: No, which table has role_invites as foreign key?

iamareebjamal: > I'm deleting all the past duplicate entries here and tested on my local too

Problem is about data loss in production DB, if one of the deleted role invite is linked to other table

shreyanshdwivedi: @iamareebjamal I understand your concern. I've cross-checked and didn't find any model which uses `role_invite` as a foreign key.

iamareebjamal: OK, so reverse the condition `R1.id < R2.id`, we want to delete all but first

shreyanshdwivedi: @iamareebjamal changed the condition.

PR #6011, opened 2019-06-08
fix: throw HTTP 400 on short password during signup

### LINKED ISSUE #5989: Wrong error throwing during sign up
**Describe the bug**
For short length of the password, the error is `HTTP 422`, It should be `HTTP 400`.

shreyanshdwivedi:
Fixes #5989 

#### Checklist

For short length of the password, the error is HTTP 422, It should be HTTP 400

Adds a new exception - `BadRequest` to raise HTTP 400 error

### DIFF
--- a/app/api/helpers/exceptions.py
+++ b/app/api/helpers/exceptions.py
@@ -31,3 +31,11 @@ class MethodNotAllowed(JsonApiException):
     """
     title = "Method Not Allowed"
     status = 405
+
+
+class BadRequest(JsonApiException):
+    """
+    Default class to throw HTTP 400 Exception
+    """
+    title = 'Bad Request'
+    status = 400
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -9,7 +9,7 @@ import urllib.error
 from app import get_settings
 from app.api.bootstrap import api
 from app.api.helpers.db import safe_query, get_count
-from app.api.helpers.exceptions import ConflictException, UnprocessableEntity, ForbiddenException
+from app.api.helpers.exceptions import ConflictException, ForbiddenException, BadRequest
 from app.api.helpers.files import create_save_image_sizes, make_frontend_url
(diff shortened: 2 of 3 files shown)


shreyanshdwivedi:
@mrsaicharan1 as we discussed, I've added a new exception because `return BadRequestError(..).respond` was not raising any error even when the `if(pass.length<8)` condition was satisfied. Do you think this is the correct approach @uds5501 @kushthedude ?
Once the implementation is finalized, I'll fix the hound and other checks.

mrsaicharan1:
> @mrsaicharan1 as we discussed, I've added a new exception because `return BadRequestError(..).respond()` was not raising any error even when the `if(pass.length<8)` condition was satisfied. Do you think this is the correct approach @uds5501 @kushthedude ?
> Once the implementation is finalized, I'll fix the hound and other checks.

That's how you should do I believe because we need exceptions which are compliant with the  `JSONAPI` structure.

mrsaicharan1:
@shreyanshdwivedi Just add a test here. Otherwise, LGTM
https://github.com/fossasia/open-event-server/blob/development/tests/all/integration/api/helpers/test_exceptions.py

shreyanshdwivedi:
@mrsaicharan1 @uds5501 @kushthedude @iamareebjamal please review

app/api/admin_translations.py
```diff
@@ -10,7 +10,6 @@
 translations_dir = 'app/translations'
 
 @admin_blueprint.route('/', methods=['GET'])
-@is_admin
```
kushthedude: Why did you remove the following decorator ?

shreyanshdwivedi: Ah! I was testing the translation thing on my local about which @mrsaicharan1 just opened the issue. Forgot to fix it. Fixing it now

kushthedude:
@shreyanshdwivedi ?

shreyanshdwivedi:
@kushthedude please refer to the issue. It is closed now

PR #6016, opened 2019-06-08
fix: Placed stats should be included in Completed to avoid Chaos

kushthedude:
Fixes https://github.com/fossasia/open-event-frontend/issues/3070

#### Checklist

Inclusion of Placed stats in completed for offline tickets to avoid misleading stats

### DIFF
--- a/app/api/order_statistics/events.py
+++ b/app/api/order_statistics/events.py
@@ -47,7 +47,7 @@ class OrderStatisticsEventSchema(Schema):
         placed = db.session.query(func.sum(OrderTicket.quantity.label('sum'))).join(Order.order_tickets).filter(
             Order.event_id == obj_id, Order.status == 'placed').scalar()
         completed = db.session.query(func.sum(OrderTicket.quantity.label('sum'))).join(Order.order_tickets).filter(
-            Order.event_id == obj_id, Order.status == 'completed' or Order.status == 'placed' ).scalar()
+            Order.event_id == obj_id, Order.status == 'completed' or Order.status == 'placed').scalar()
         result = {
             'total': total or 0,
             'draft': draft or 0,
--- a/app/api/order_statistics/events.py
+++ b/app/api/order_statistics/events.py
@@ -47,7 +47,7 @@ class OrderStatisticsEventSchema(Schema):
         placed = db.session.query(func.sum(OrderTicket.quantity.label('sum'))).join(Order.order_tickets).filter(
             Order.event_id == obj_id, Order.status == 'placed').scalar()
         completed = db.session.query(func.sum(OrderTicket.quantity.label('sum'))).join(Order.order_tickets).filter(
-            Order.event_id == obj_id, Order.status == 'completed').scalar()
+            Order.event_id == obj_id, Order.status == 'completed' or Order.status == 'placed' ).scalar()
         result = {
             'total': total or 0,
             'draft': draft or 0,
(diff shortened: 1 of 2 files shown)


kushthedude:
@shreyanshdwivedi @uds5501 Please have a look

app/api/order_statistics/events.py
```diff
@@ -47,7 +47,7 @@ def tickets_count(self, obj):
         placed = db.session.query(func.sum(OrderTicket.quantity.label('sum'))).join(Order.order_tickets).filter(
             Order.event_id == obj_id, Order.status == 'placed').scalar()
         completed = db.session.query(func.sum(OrderTicket.quantity.label('sum'))).join(Order.order_tickets).filter(
-            Order.event_id == obj_id, Order.status == 'completed').scalar()
+            Order.event_id == obj_id, Order.status == 'completed' or Order.status == 'placed').scalar()
```
uds5501: This implementation is wrong.
If the placed order had to be included in the completed schema, it would've been done like that in the first place.
 - A simple addition of completed and placed in the event stats on FE would do.
- To make sure no extra tickets are being bought in the confusion, you should just handle the checks.

kushthedude: @uds5501 
I tried it in first place but that would be creating a lot of confusion : 
Let's say i just include a check on ticket buying to be not less than the number of placed and completed tickets bought .
The sales would be affected , The number of orders , All the status would be affected .
Regarding your point it should have in completed schema from start , AFAIK All the offline tickets were already getting included in completed , But due to recent FE PR we made offline tickets into the` placed tag` and we didnt modify the server for the following

uds5501: @kushthedude the point I am trying to make is, don't add `placed` tickets in the schema fields which is designated for `completed` ones, **that** would be a serious confusion.

"Let's say i just include a check on ticket buying to be not less than the number of placed and completed tickets bought ." - I don't see the problem there, isn't that what we want to do? Not overshooting the tickets when they have placed orders and completed orders already?

kushthedude: `Not overshooting the tickets when they have placed orders and completed orders already?`
What about the Sales, Number of orders, Ticket Count,Attendee made by offline tickets , And if i do this all manually on FE , It would increase redundancy on the code . 
And regarding the placed and completed ticket, As by the present system , The only difference i can see between the placed and completed order is `mode of payment` and If a ticket is placed we need to collect the amount by means of offline . Is there any more difference ?
@uds5501

app/api/order_statistics/tickets.py
```diff
@@ -99,7 +99,7 @@ def sales_count(self, obj):
         placed = db.session.query(func.sum(Order.amount.label('sum'))).join(Order.order_tickets).filter(
             OrderTicket.ticket_id == obj_id, Order.status == 'placed').scalar()
         completed = db.session.query(func.sum(Order.amount.label('sum'))).join(Order.order_tickets).filter(
-            OrderTicket.ticket_id == obj_id, Order.status == 'completed').scalar()
+            OrderTicket.ticket_id == obj_id, Order.status == 'completed' or Order.status == 'placed').scalar()
         result = {
             'total': total or 0,
             'draft': draft or 0,
```
shreyanshdwivedi: @kushthedude I agree to @uds5501 
We should apply simple addition on FE to show the stats. I don't see any difficulty or problem here.
Why exactly will this create a confusion?
We should leave the `Completed` and `Placed` stats as it is on server IMO

PR #6035, opened 2019-06-12
feat: implement threshold for resetting password

### LINKED ISSUE #6033: Limit the number of emails an account can sent out in a given timeframe
One account yesterday tried a password reset requesting a reset 1000 times.

Please implement a threshhold for account resets and propose a threshhold for other emails in a certain time frame. Open relevant issues for this.

A threshhold for an account reset email should be 3 times per hour. After that the user should receive a message: "You have reached the threshhold for this action. Please try again in one hour."

shreyanshdwivedi:
Fixes #6033 

#### Checklist

A threshold for an account reset email should be 3 times per hour. After that, the user receives a message: "You have reached the threshold for this action. Please try again in one hour."

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -13,6 +13,7 @@ from app.api.helpers.order import create_pdf_tickets_for_holder
 from app.api.helpers.storage import generate_hash
 
 from app import get_settings
+from app import limiter
 from app.api.helpers.db import save_to_db, get_count
 from app.api.helpers.errors import ForbiddenError, UnprocessableEntityError, NotFoundError, BadRequestError
 from app.api.helpers.files import make_frontend_url
@@ -28,7 +29,6 @@ from app.models.mail import PASSWORD_RESET, PASSWORD_CHANGE, \
 from app.models.notification import PASSWORD_CHANGE as PASSWORD_CHANGE_NOTIF
 from app.models.user import User
 from app.api.helpers.storage import UPLOAD_PATHS
-
 
 authorised_blueprint = Blueprint('authorised_blueprint', __name__, url_prefix='/')
 ticket_blueprint = Blueprint('ticket_blueprint', __name__, url_prefix='/v1')
@@ -207,6 +207,9 @@ def resend_verification_email():
 
 
 @auth_routes.route('/reset-password', methods=['POST'])
+@limiter.limit(
+    '3/hour', key_func=lambda: request.json['data']['email'], error_message='Limit for this action exceeded'
+)
 def reset_password_post():
     try:
         email = request.json['data']['email']
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -8,6 +8,7 @@ from healthcheck import EnvironmentDump
 from functools import wraps
 from flask import request, jsonify, make_response, Blueprint, send_file, url_for, redirect
 from flask_jwt import current_identity as current_user, jwt_required
+from flask_limiter.util import get_remote_address
 from sqlalchemy.orm.exc import NoResultFound
 from app.api.helpers.order import create_pdf_tickets_for_holder
 from app.api.helpers.storage import generate_hash
@@ -209,6 +210,9 @@ def resend_verification_email():
 @auth_routes.route('/reset-password', methods=['POST'])
 @limiter.limit(
     '3/hour', key_func=lambda: request.json['data']['email'], error_message='Limit for this action exceeded'
+)
+@limiter.limit(
+    '1/minute', key_func=get_remote_address, error_message='Limit for this action exceeded'
 )
 def reset_password_post():
     try:
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -11,6 +11,7 @@ from flask_migrate import Migrate, MigrateCommand
 from flask_script import Manager
 from flask_login import current_user
 from flask_jwt import JWT
+from flask_limiter import Limiter
 from datetime import timedelta
 from flask_cors import CORS
 from flask_rest_jsonapi.errors import jsonapi_errors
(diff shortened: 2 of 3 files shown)


shreyanshdwivedi:
@uds5501 @mrsaicharan1 @kushthedude I'm not getting the time in the same format as it is getting stored in the db. For example, the time getting stored in db is - `2019-06-12 13:36:29.233271+05:30` and one I'm getting is `2019-06-12 13:36:29.227708+00:00`. Can you guys help me out to find what am I missing here?

kushthedude:
@shreyanshdwivedi We are not using pytz for time now , As recent change made by @prateekj117 we are now using `SQLalchemy` inbuilt time stamp function, Check there if you find a way to retireve time from db

app/api/auth.py
```diff
@@ -222,7 +226,14 @@ def reset_password_post():
         if user.was_registered_with_order:
             send_email_with_action(user, PASSWORD_RESET_AND_VERIFY, app_name=get_settings()['app_name'], link=link)
         else:
-            send_email_with_action(user, PASSWORD_RESET, app_name=get_settings()['app_name'], link=link)
+            previous_hour = datetime.now(pytz.utc) - timedelta(hours=1)
```
prateekj117: @shreyanshdwivedi Try datetime.utcnow

shreyanshdwivedi: @prateekj117 when I used `datetime.utcnow()`, I'm getting datetime without timezone - `2019-06-12 12:54:13.035101` - neither `00:00` nor `05:30`

prateekj117: The now() call returns the correct time for local location, while the utcnow() call returns the time in the UTC time zone.

shreyanshdwivedi: The time returned by both are absolutely correct. I just want to include my current timezone

shreyanshdwivedi:
@niranjan94 so should I use flask_limiter here rather than this approach for `password-reset`?

niranjan94:
> so should I use flask_limiter here rather than this approach for `password-reset`?

Yep. Correct. And the frontend can show an appropriate message to the user if it gets a `429` response code.

shreyanshdwivedi:
@niranjan94 ok cool. But one thing, the issue mentioned -
>A threshhold for an account reset email should be 3 times per hour. After that the user should receive a message: "You have reached the threshhold for this action. Please try again in one hour."

But I don't think this type of message can be achieved by flask_limiter.

niranjan94:
> But I don't think this type of message can be achieved by flask_limiter.

@shreyanshdwivedi 

https://github.com/fossasia/open-event-server/pull/6035#issuecomment-501293990

app/api/auth.py
prateekj117: Try datetime.utcnow().isoformat()

app/api/auth.py
```diff
@@ -207,6 +208,7 @@ def resend_verification_email():
 
 
 @auth_routes.route('/reset-password', methods=['POST'])
+@limiter.limit("10/s", exempt_when=lambda: current_user.is_admin, error_message='Limit for this action exceeded')
```
iamareebjamal: Use 3/hour

app/api/auth.py
```diff
@@ -29,10 +30,10 @@
 from app.models.user import User
 from app.api.helpers.storage import UPLOAD_PATHS
 
-
 authorised_blueprint = Blueprint('authorised_blueprint', __name__, url_prefix='/')
 ticket_blueprint = Blueprint('ticket_blueprint', __name__, url_prefix='/v1')
 auth_routes = Blueprint('auth', __name__, url_prefix='/v1/auth')
+limiter = Limiter()
```
iamareebjamal: Initialize in root __init__.py so all nested modules can use

shreyanshdwivedi:
@iamareebjamal @niranjan94 I've updated the `key_func` and finalized the PR. I've checked it on local and it works for me. The error message produced -
```
429 Too Many Requests
Too Many Requests
Limit for this action exceeded
```
Please review

app/api/auth.py
```diff
@@ -207,6 +207,9 @@ def resend_verification_email():
 
 
 @auth_routes.route('/reset-password', methods=['POST'])
+@limiter.limit(
+    '3/hour', key_func=lambda: request.json['data']['email'], error_message='Limit for this action exceeded'
```
iamareebjamal: jamal.areeb@gmail.com
jamalareeb@gmail.com
ja.mal.are.eb@gmail.com

Are all same email addresses and will be allowed by this method

shreyanshdwivedi: @iamareebjamal I just tested the above three email addresses.
Each one is distinguished perfectly and failed when the email if entered fourth time. I jumbled the order and tested, the reset request for one email is not affecting other in any way

iamareebjamal: It should

iamareebjamal: The point is that I can keep adding dots and sending emails at a rate of more than 3/hour

I can loop through a list of emails and send password reset request on them. Essentially brute forcing

shreyanshdwivedi: But there is a check which raises error if user is not registered and so no mail is sent out. If a user loops by adding dots, it doesn't mean the email is registered and so error will be raised. I hope it makes sense to you.

iamareebjamal: A hacker doesn't care if the error is raised. He'll just continue looping through emails, and our resources will get exhausted. And users will still get spammed.

Also, raising error if user is not registered is a pretty big security issue, it should be fixed ASAP - #6069

shreyanshdwivedi: @iamareebjamal from what I understood, I should create a function which treats similar emails as same email. Right?
Also can you please guide me which pattern should I consider ?

iamareebjamal: Instead, you should add one more limiter which prevents same IP to access the unprotected resource at a lower rate, like '1/minute'

shreyanshdwivedi: @iamareebjamal I've updated the PR. Please review

PR #6041, opened 2019-06-12
feat: Add an admin-only endpoint for test mail

### LINKED ISSUE #6040: Add an end point for sending test mails
The endpoint should only be available to the admin, and can be used to send a test mail to any address.

CosmicCoder96:
Fixes #6040 

- Adds an endpoint which allows an admin or super admin to send test emails.


app/api/settings.py
```diff
@@ -51,3 +55,21 @@ def before_get(self, args, kwargs):
     def after_patch(self, result):
         # Update settings cache after PATCH
         refresh_settings()
+
+
+@admin_misc_routes.route('/test-mail', methods=['POST'])
+def test_email_setup():
+    if 'Authorization' in request.headers:
+        _jwt_required(app.config['JWT_DEFAULT_REALM'])
```
iamareebjamal: Why not using decorator?

CosmicCoder96:
@iamareebjamal I completely forgot about the custom decorators have updated the PR.

app/api/settings.py
```diff
@@ -51,3 +57,15 @@ def before_get(self, args, kwargs):
...
+
+
+@admin_misc_routes.route('/test-mail', methods=['POST'])
+@is_admin
+def test_email_setup():
+    try:
+        recipient = request.json['recipient']
+    except KeyError:
+        return UnprocessableEntityError(
+            {'source': ''}, 'Invalid request payload').respond()
```
iamareebjamal: Better to only use exception handling in exceptional cases

```python
recipient = request.json.get('recipient')
if not recipient:
    return UnprocessableEntityError({'source': 'recipient'}, 'Required parameter recipient not found').respond() // explicit error message

# continue with happy path
```

CosmicCoder96:
@iamareebjamal Thanks for the review, I have made the changes.

PR #6042, opened 2019-06-13
feat: Adds event owner role to the system

### LINKED ISSUE #6001: Introduce event owner user role

Currently, owners of the event cannot be differentiated. So `organizer` event role is to be changed to `owner`, to differentiate and make it clear.

**Additional context**

On it.
Reference https://github.com/fossasia/open-event-frontend/issues/3012#issuecomment-498873640

shreyanshdwivedi:
Fixes #6001 

#### Checklist

Currently, owners of the event cannot be differentiated. So organizer event role is to be changed to owner, to differentiate and make it clear

Adds event `owner` role to the system


shreyanshdwivedi:
@iamareebjamal I wanted to ask on thing. How are we storing roles in db? I see 6 roles in db but couldn't find any migration which adds those roles.
I want to update the roles table and insert `owner` at `id=1`. Can you please give me a little background?

CosmicCoder96:
Why do you want to store it at ID 1? We don’t manually set the ID of the
records. Role is searched by its name everywhere not ID

shreyanshdwivedi:
Actually there are a few tables where id 1 is being used to associate event and role. So I thought I could add a role at id 1 which will automatically make the user the owner. I thought I could do it in a single go.
However now I feel that there might be more than 1 organizer in production and there can be only one owner. So this will not be good.
I will add role and use its id to make 1st organizer the event owner.

CosmicCoder96:
Okay, as to the answer to your original question I believe the roles are
created in the create DB script.

iamareebjamal:
> I believe the roles are created in the create DB script

This is problematic 

Make a non conflicting migration for it

CosmicCoder96:
@shreyanshdwivedi here,
https://github.com/fossasia/open-event-server/blob/f192ad56f630383a36c3eec47a3140a821968b0f/populate_db.py#L60

shreyanshdwivedi:
@CosmicCoder96 yes I got that earlier but how will it reflect in my db because changing it is not affecting my roles table.
I think I'll have to make a migration file to add roles in db as mentioned by @iamareebjamal above. But if the order of the roles will change, their ids will changes and other tables which stores id of roles will be affected.
I'll check what I can do to avoid any conflicts

CosmicCoder96:
@shreyanshdwivedi @iamareebjamal I don't think looking up for roles using their ID is ever a good idea, their name or slug is their distinguishing factor, shouldn't they be referred to using that, in which case this task can be merely reduced to adding the owner role in the populate script, and writing a migration which adds the role of owner, IF no other role with the name owner exists.  

@shreyanshdwivedi Can you point me to where 

> Actually there are a few tables where id 1 is being used to associate event and role

I think they should be replaced with 
```
role = Role.query.filter_by(name=ORGANIZER).first()
```
Like it has been used at some places in the app. (This example is from the events API).

shreyanshdwivedi:
@CosmicCoder96 you can refer to `users_events_roles` table in db for instance where `event_id`, `user_id` and `role_id` are the columns.

CosmicCoder96:
@shreyanshdwivedi For the legacy data, for each event, in the user events table the record which was inserted first (with an organizer relationship), denotes the creator of the event. You can leave the original data as is, create a migration to add owner role if it does not exist in roles table, and also add it to the end of populate db file. Next, add a migration to find the implicit creator of the event, (using the above logic), find the id of the entry in roles table whose name is 'OWNER' and insert into user_events_roles table the user_id, role_id(for the owner, whatever it may be ) and the event ID. What do you think?

shreyanshdwivedi:
I totally agree to the steps you suggested. Just one thing, when we find out the owner of the event, I'll suggest to update the role_id of the entry in users_events_roles table rather than inserting a new one as it may lead to confusion later. If we update the role_id only, it'll be better. Does it sound correct?

CosmicCoder96:
But an event owner is still an organiser yes ? Or no?

shreyanshdwivedi:
@CosmicCoder96 yes. An event owner will have all the rights of an organizer just like an organizer has all the rights of a coorganizer but the roles will still be different. Later we will provide more rights to the owner

bhaveshAn:
@shreyanshdwivedi Areeb is right. Migration should also be taken care of if modifying the ID. Else if we create owner role after all the pre existing roles in `create_roles` that's also not an issue. As there is no hard and fast rule that the roles should be created in decreasing order of their degree.

shreyanshdwivedi:
@CosmicCoder96 @iamareebjamal @bhaveshAn I've added a migration file that renames the column name too and introduce a new role in server. It also updates the first organizer to owner. 
The codacy is failing because of argument named `f` but there are other functions in same file which follow same thing. That's why I followed this convention.
Please review

shreyanshdwivedi:
@iamareebjamal @CosmicCoder96 @niranjan94 please review

app/api/__init__.py
```diff
@@ -315,6 +315,8 @@
 api.route(EventRelationship, 'event_stripe_authorization', '/events/<int:id>/relationships/stripe-authorization',
           '/events/<identifier>/relationships/stripe-authorization')
 # Events -> roles:
+api.route(EventRelationship, 'event_owner', '/events/<int:id>/relationships/owner',
+          '/events/<identifier>/relationships/owner')
```
CosmicCoder96: what about the corresponding user relationship?

shreyanshdwivedi: @CosmicCoder96 I've added owner to event relationship -
`api.route(UserRelationship, 'user_owner_event', '/users/<int:id>/relationships/owner-events')`

app/api/schema/events.py
```diff
@@ -283,6 +283,13 @@ def validate_timezone(self, data, original_data):
                                 schema='CustomFormSchema',
                                 many=True,
                                 type_='custom-form')
+    owner = Relationship(attribute='owner',
+                         self_view='v1.event_owner',
+                         self_view_kwargs={'id': '<id>'},
+                         related_view='v1.user_list',
```
CosmicCoder96: how can an event have multiple owners, wasn't the whole point of the owner was to differentiate between creator and organizers?

shreyanshdwivedi: @CosmicCoder96 missed it. Fixed now

app/models/event.py
```diff
@@ -125,6 +125,13 @@ class Event(SoftDeletionModel):
         'EventSubTopic', backref='event', foreign_keys=[event_sub_topic_id])
     events_orga = db.relationship(
         'EventOrgaModel', backref='event', foreign_keys=[events_orga_id])
+    owner = db.relationship('User',
```
CosmicCoder96: this will return an array/list with a single user inside it, check docs of alchemy to restrict it to a single result.

shreyanshdwivedi: @CosmicCoder96 I've used - `owner = db.relationship('User', backref='owner_events')` now

shreyanshdwivedi:
@CosmicCoder96 @iamareebjamal I've updated the PR and made required changes. Please review

app/api/helpers/permissions.py
```diff
@@ -103,6 +103,27 @@ def decorated_function(*args, **kwargs):
...
+    Allows only Owner to access the event resources.
+    :param f:
+    :return:
+    """
+
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        user = current_identity
+
+        if user.is_staff:
```
mrsaicharan1: Do we need this as this is only for an owner?

app/api/events.py
```diff
@@ -154,19 +154,26 @@ def before_post(self, args, kwargs, data=None):
...
-        after create method to save roles for users and add the user as an accepted role(organizer)
+        after create method to save roles for users and add the user as an accepted role(owner and organizer)
         :param event:
         :param data:
         :param view_kwargs:
         :return:
         """
-        role = Role.query.filter_by(name=ORGANIZER).first()
         user = User.query.filter_by(id=view_kwargs['user_id']).first()
+        role = Role.query.filter_by(name=OWNER).first()
```
uds5501: @shreyanshdwivedi after event creation, shouldn't there be a role for `organiser` too for the owner?

app/api/helpers/export_helpers.py
```diff
@@ -29,8 +29,8 @@
 FIELD_ORDER = {
     'event': [
         'id', 'name', 'latitude', 'longitude', 'location_name', 'starts_at', 'ends_at',
-        'timezone', 'description', 'original_image_url', 'logo_url', 'organizer_name',
-        'organizer_description', 'external_event_url', 'ticket_url', 'privacy', 'event_type_id',
+        'timezone', 'description', 'original_image_url', 'logo_url', 'owner_name',
```
uds5501: @shreyanshdwivedi IMO, organizer name should be preserved too for the sake of upcoming roles or we are using only owner for now?

app/api/helpers/permissions.py
shreyanshdwivedi: Actually this is needed. It gives admin, superadmin access to as owner of the event. An admin and superadmin should have all the rights. That's why included this like other permissions

app/api/helpers/export_helpers.py
shreyanshdwivedi: @uds5501 we can get details of organizers via Relationship we have. Only owner info should suffice here as it can be used for filtering too.

app/api/events.py
shreyanshdwivedi: @uds5501 when event is created the user creating the event will be the owner. 
He can then send role-invites to user to become organizer.

iamareebjamal: Will the owner have same/more permissions than an organizer? If not, fix it

shreyanshdwivedi: For now owner will have same permissions as that of an organizer. Later we can give more permissions to owner

shreyanshdwivedi:
@iamareebjamal @mrsaicharan1 @uds5501  few updates - 
I ran through the whole system while I was on this branch. I successfully -
 - Created event
 - Order ticket
 - Transferred event
 - Invited users to different roles
 - Created and Updated sessions
 - Used postman to check the routes I introduced. 
I encountered 2 bugs that were there for a long time and reported them.
Please review this PR so that we can move to other linked PRs. Thank you

uds5501:
@shreyanshdwivedi Understood. I am pulling the changes and testing in my system.

iamareebjamal:
Other than the changes requested, looks good

app/api/events.py
```diff
@@ -154,19 +154,26 @@ def before_post(self, args, kwargs, data=None):
...
                                  status='accepted')
-        save_to_db(role_invite, 'Organiser Role Invite Added')
+        save_to_db(role_invite, 'Owner Role Invite Added')
+
+        # role = Role.query.filter_by(name=ORGANIZER).first()
+        # uer = UsersEventsRoles(user, event, role)
+        # save_to_db(uer, 'Event Saved')
+        # role_invite = RoleInvite(user.email, role.title_name, event.id, role.id, datetime.now(pytz.utc),
+        #                          status='accepted')
+        # save_to_db(role_invite, 'Organizer Role Invite Added')
```
iamareebjamal: Remove commented lines

app/api/helpers/permission_manager.py
```diff
@@ -94,7 +109,8 @@ def is_coorganizer_endpoint_related_to_event(view, view_args, view_kwargs, *args
         _jwt_required(app.config['JWT_DEFAULT_REALM'])
         return view(*view_args, **view_kwargs)
 
-    if user.is_organizer(kwargs['event_id']) or user.is_coorganizer(kwargs['event_id']):
+    if user.is_owner(kwargs['event_id']) or user.is_organizer(kwargs['event_id']) or \
+            user.is_coorganizer(kwargs['event_id']):
```
iamareebjamal: These 3 conditions are very common and repeated. Move to user model in a function and repeat everywhere

shreyanshdwivedi:
> @shreyanshdwivedi Understood. I am pulling the changes and testing in my system.

@uds5501 do let us know, please. This is a big refactor. Your input is crucial and important to us
Thank You.

app/api/events.py
```diff
@@ -46,7 +46,7 @@
 from app.models.ticket_holder import TicketHolder
 from app.models.track import Track
 from app.models.user_favourite_event import UserFavouriteEvent
-from app.models.user import User, ATTENDEE, ORGANIZER, COORGANIZER
+from app.models.user import User, ATTENDEE, OWNER, ORGANIZER, COORGANIZER
```
uds5501: @shreyanshdwivedi The basics of event an seems to be working fine. Although you might want to create a follow up PR on the front end regarding this (for existing and new events):
image

shreyanshdwivedi: @uds5501 Yeah, a follow-up PR is required to change few more things on frontend. I'll do it once this PR is merged

uds5501: @shreyanshdwivedi Apologies for not elaborating.
There should be a role of "owner" displayed on the dashboard and the manage event table

shreyanshdwivedi: Sure @uds5501. Will open a followup PR on FE once this is merged. Thanks for the input

shreyanshdwivedi:
@uds5501 @iamareebjamal plz review

app/api/helpers/permission_manager.py
```diff
@@ -176,7 +189,7 @@ def is_speaker_itself_or_admin(view, view_args, view_kwargs, *args, **kwargs):
     if user.is_admin or user.is_super_admin:
         return view(*view_args, **view_kwargs)
 
-    if user.is_organizer(kwargs['event_id']) or user.is_coorganizer(kwargs['event_id']):
+    if user.is_event_head(kwargs['event_id']):
```
iamareebjamal: has_event_access

shreyanshdwivedi:
@iamareebjamal did the changes. Please review

shreyanshdwivedi:
@CosmicCoder96 please review

iamareebjamal:
Fix Travis Build

iamareebjamal:
Most probably due to split migration heads, and the script did not fail correctly

shreyanshdwivedi:
Yeah I'll have to update the migration to recent merged one
`Error: Multiple heads are present; please specify a single target revision` is in travis failure

shreyanshdwivedi:
@CosmicCoder96 @iamareebjamal please review

PR #6067, opened 2019-06-18
feat: Addition of Field for Email Exception in Speaker Schema

### LINKED ISSUE #6059: Addition of Field for `Override Email` for Speaker
Sub-Issue of https://github.com/fossasia/open-event-frontend/issues/2953

kushthedude:
Fixes #6059 

- Added a field `is_email_override`

#### Checklist


kushthedude:
@iamareebjamal @CosmicCoder96 @niranjan94 Please review the following

shreyanshdwivedi:
@kushthedude according to https://github.com/fossasia/open-event-frontend/issues/2953#issuecomment-500279926 we were going to add organizer's email in place of the speaker's email so that we can know which organizer added the speaker and opted to omit the speaker's email.
Then why are you adding a new field? Can you please clarify what are you trying to achieve by adding this new field?

kushthedude:
> @kushthedude according to [fossasia/open-event-frontend#2953 (comment)](https://github.com/fossasia/open-event-frontend/issues/2953#issuecomment-500279926) we were going to add organizer's email in place of the speaker's email so that we can know which organizer added the speaker and opted to omit the speaker's email.

@shreyanshdwivedi Adding the following field will help us clear the following speaker was made exceptional by the organiser , 
Make a case in which Organiser is also a speaker, it may create confusion that's why i made a separate field to `override_email`

shreyanshdwivedi:
@kushthedude do we really need to know if the speaker was added exceptionally by organizer explicitly? As I said earlier that it was discussed in meeting to add organizer's email itself.
However last call is to be made by mentors. But IMO I don't feel that adding a field explicitly for this is needed

kushthedude:
@iamareebjamal @CosmicCoder96 ?

kushthedude:
> > @kushthedude according to [fossasia/open-event-frontend#2953 (comment)](https://github.com/fossasia/open-event-frontend/issues/2953#issuecomment-500279926) we were going to add organizer's email in place of the speaker's email so that we can know which organizer added the speaker and opted to omit the speaker's email.
> 
> @shreyanshdwivedi Adding the following field will help us clear the following speaker was made exceptional by the organiser ,
> Make a case in which Organiser is also a speaker, it may create confusion that's why i made a separate field to `override_email`

@iamareebjamal @CosmicCoder96 ?

kushthedude:
@iamareebjamal @CosmicCoder96 @niranjan94 Review about the PR Please

app/models/speaker.py
```diff
@@ -24,6 +24,7 @@ class Speaker(SoftDeletionModel):
     linkedin = db.Column(db.String)
     organisation = db.Column(db.String)
     is_featured = db.Column(db.Boolean, default=False)
+    is_email_override = db.Column(db.Boolean, default=False)
```
iamareebjamal: Should be non-nullable

kushthedude: @iamareebjamal did the changes

kushthedude:
@iamareebjamal Please Review

kushthedude:
@iamareebjamal Please Review

migrations/versions/efcf6d57c41f_.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'efcf6d57c41f'
+down_revision = '43e8c59337ae'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('speaker', sa.Column('is_email_override', sa.Boolean(), nullable=False))
```
iamareebjamal: This migration will fail

kushthedude: @iamareebjamal Made the change

kushthedude:
@iamareebjamal Please check now

kushthedude:
@iamareebjamal Please check now

iamareebjamal:
Ask peers to review first

kushthedude:
> However last call is to be made by mentors. But IMO I don't feel that adding a field explicitly for this is needed

@iamareebjamal This is the contradiction we are facing , do we need the following field or not ?

shreyanshdwivedi:
@kushthedude you may follow your approach. We discussed it on a meet today again.

kushthedude:
@uds5501 @shreyanshdwivedi Please REview

kushthedude:
@niranjan94 @iamareebjamal Please REview

kushthedude:
> What about the checks? When this attribute will be true and there will be no email, will there be no errors for empty speaker email? How are you tackling it?

The default is false , When the organiser opt the option the value will be set true from FE, that's only the use of the following bool.
If Bool is true, Email of the speaker will be the same as an organiser or it can be interpreted that the following speaker is a VIP, This bool can also be helped in determining a VIP speaker and a general speaker.
So email is not going to be empty.

shreyanshdwivedi:
>Email of the speaker will be the same as an organizer

@kushthedude here comes the problem :sweat_smile: 
Actually, when we discussed this issue in yesterday's meeting, @mariobehling made a point that if we use the organizer's email then he will also receive unwanted mail which is meant for the VIP user.

kushthedude:
> > Email of the speaker will be the same as an organizer
> 
> @kushthedude here comes the problem 😅
> Actually, when we discussed this issue in yesterday's meeting, @mariobehling made a point that if we use the organizer's email then he will also receive unwanted mail which is meant for the VIP user.

That's fascinating, So email should be null? , One more thing what about the email updates related to the event ? , Who is gonna notify the speaker  ? @shreyanshdwivedi

uds5501:
@shreyanshdwivedi @kushthedude 
One point . In this case the organiser is the one setting up the speaker and the session anyway, so this should not be a problem.

kushthedude:
> @shreyanshdwivedi @kushthedude
> One point . In this case the organiser is the one setting up the speaker and the session anyway, so this should not be a problem.

If thats the case , Than I will add checks for `null` email

shreyanshdwivedi:
>That's fascinating, So email should be null? , One more thing what about the email updates related to the event ? , Who is gonna notify the speaker ? @shreyanshdwivedi

@kushthedude @uds5501 I think we should move forward by adding organizer's email to the db. But while sending email and notification, we can check if user's email is not overridden then only send the notification or email. Your thoughts?

kushthedude:
> > That's fascinating, So email should be null? , One more thing what about the email updates related to the event ? , Who is gonna notify the speaker ? @shreyanshdwivedi
> 
> @kushthedude @uds5501 I think we should move forward by adding organizer's email to the db. But while sending email and notification, we can check if user's email is not overridden then only send the notification or email. Your thoughts?

@shreyanshdwivedi @uds5501 I am getting a bit confused now, I am still confused about the following , If we added an organiser email, Than we need to refactor the whole notification system to check if the speaker has the following boolean checked or not, I don't think this would be a nice way for the following, I am closing the PR for now, Lets follow up in next meeting with this for a more brighter approach, If in the meantime anyone wants to catch up, I have no objection.

PR #6068, opened 2019-06-18
fix: Showing STR error in case of DEVELOPMENT app config

### LINKED ISSUE #4903: Show debug error information on exception for development instance
For development instance of application, especially on obscure errors like 500, debug stack trace information should be shown to the user in the response JSON

This will severely cut down the reliance of client developers on server developers and enable them to create issues with proper data

kushthedude:
Fixes #4903 

#### Checklist

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -45,6 +45,7 @@ import sentry_sdk
 from sentry_sdk.integrations.flask import FlaskIntegration
 
 
+logger = logging.getLogger(__name__)
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 
 static_dir = os.path.dirname(os.path.dirname(__file__)) + "/static"
@@ -245,9 +246,10 @@ scheduler.start()
 
 @app.errorhandler(500)
 def internal_server_error(error):
-    if current_app.config['DEVELOPMENT'] is True:
-        exc = JsonApiException({'pointer': ''}, str(error))
+    if current_app.config['PROPOGATE_ERROR'] is True:
+        raise error
     else:
+        logger.exception(error)
         exc = JsonApiException({'pointer': ''}, 'Unknown error')
     return make_response(json.dumps(jsonapi_errors([exc.to_dict()])), exc.status,
                          {'Content-Type': 'application/vnd.api+json'})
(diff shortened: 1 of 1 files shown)


kushthedude:
@shreyanshdwivedi Found a quick fix so pushed the PR ,
@iamareebjamal Please check if it is acceptable

app/__init__.py
```diff
@@ -245,7 +245,7 @@ def update_sent_state(sender=None, headers=None, **kwargs):
 
 @app.errorhandler(500)
 def internal_server_error(error):
-    if current_app.config['PROPOGATE_ERROR'] is True:
+    if current_app.config['DEVELOPMENT'] is True:
```
CosmicCoder96: This is redundant, propagate error is already true for development config.

https://github.com/fossasia/open-event-server/blob/8f2de81dbbfe179770f160f59fc365316ffed660/config.py#L119

iamareebjamal:
Secondly, also do logger.exception so that the exception is at least logged on the console

iamareebjamal:
Now check if it is working

iamareebjamal:
@mrsaicharan1 @shreyanshdwivedi Can you please check?

app/__init__.py
```diff
@@ -246,8 +247,9 @@ def update_sent_state(sender=None, headers=None, **kwargs):
 @app.errorhandler(500)
 def internal_server_error(error):
     if current_app.config['PROPOGATE_ERROR'] is True:
-        exc = JsonApiException({'pointer': ''}, str(error))
+        raise error
```
bhaveshAn: You have to display appropriate exception at API level and console level too.
In API as Json response and at console using `logger.exception`.

iamareebjamal: Nope, JSON will not be returned in case PROPAGATE_ERROR is True

kushthedude:
> Even though better than 'Unknown error', str(error) is not enough. We want full stack strace of the error

Sorry @iamareebjamal I was AFK, That's why could not reply, Regarding the Previous change which i made did show the JSON error, But full stack trace was not visible

iamareebjamal:
I have made certain changes. Please check

kushthedude:
> @kushthedude Can you replicate an error and upload a screenshot of the stacktrace being shown?

@mrsaicharan1 I didn't try to check with the latest change, As the `ENV` was set to `production` in deployments, Which had been fixed by @iamareebjamal.

PR #6062, opened 2019-06-18
feat: Model & schema validations for donations

mrsaicharan1:
Fixes #6061

#### Checklist

- [ ] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- [ ] My branch is up-to-date with the Upstream `development` branch.
- [ ] The unit tests pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

- [ ] All the functions created/modified in this PR contain relevant docstrings.

#### Short description of what this resolves:

- Adds schemas & fields for donations
#### Changes proposed in this pull request:

- Migration file
- Schema & model additions
- Added validation checks

PR #6071, opened 2019-06-19
fix: Correct server default type in migration

iamareebjamal:
Fixes bug in #6062


iamareebjamal:
@mrsaicharan1 Please always test migrations by running them

mrsaicharan1:
@iamareebjamal Yes will keep that in mind.

PR #6079, opened 2019-06-20
fix: updates message on reset password if user isn't registered

### LINKED ISSUE #6069: Error raised on reset password if user is not registered
Raising an error on password reset if the user is not registered is like gifting hackers a list of users present in the database.

Instead, regardless of user being present in the system, it should return 

"If your email was registered with us, you'll get an email with reset link shortly"

shreyanshdwivedi:
Fixes #6069 

Raising an error on password reset if the user is not registered is like gifting hackers a list of users present in the database.

 - Updates the message is user is not registered on reset password

#### Checklist

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -1,5 +1,6 @@
 import os
 import base64
+import logging
 import random
 import string
 
@@ -32,6 +33,7 @@ from app.models.user import User
 from app.api.helpers.storage import UPLOAD_PATHS
 from app.api.helpers.auth import AuthManager
 
+logger = logging.getLogger(__name__)
 authorised_blueprint = Blueprint('authorised_blueprint', __name__, url_prefix='/')
 ticket_blueprint = Blueprint('ticket_blueprint', __name__, url_prefix='/v1')
 auth_routes = Blueprint('auth', __name__, url_prefix='/v1/auth')
@@ -224,7 +226,7 @@ def reset_password_post():
     try:
         user = User.query.filter_by(email=email).one()
     except NoResultFound:
-        return NotFoundError({'source': ''}, 'User not found').respond()
+        logger.info('Tried to reset password not existing email %s', email)
     else:
         link = make_frontend_url('/reset-password', {'token': user.reset_password})
         if user.was_registered_with_order:
@@ -232,7 +234,8 @@ def reset_password_post():
         else:
             send_email_with_action(user, PASSWORD_RESET, app_name=get_settings()['app_name'], link=link)
 
-    return make_response(jsonify(message="Email Sent"), 200)
+    return make_response(jsonify(message="If your email was registered with us, you'll get an \
+                         email with reset link shortly", email=email), 200)
 
 
 @auth_routes.route('/reset-password', methods=['PATCH'])


shreyanshdwivedi:
@iamareebjamal the server responds with message `If your email was registered with us, you'll get an email with reset link shortly` and status `200` always. Do you want to update FE message too? Currently it shows - `Please go to the link sent to your email to reset your password`

app/api/auth.py
```diff
@@ -223,15 +223,17 @@ def reset_password_post():
...
+                             email with reset link shortly"), 200)
     else:
         link = make_frontend_url('/reset-password', {'token': user.reset_password})
         if user.was_registered_with_order:
             send_email_with_action(user, PASSWORD_RESET_AND_VERIFY, app_name=get_settings()['app_name'], link=link)
         else:
             send_email_with_action(user, PASSWORD_RESET, app_name=get_settings()['app_name'], link=link)
 
-    return make_response(jsonify(message="Email Sent"), 200)
+    return make_response(jsonify(message="If your email was registered with us, you'll get an \
```
iamareebjamal: Remove duplication

app/api/auth.py
```diff
@@ -223,15 +223,17 @@ def reset_password_post():
     try:
         user = User.query.filter_by(email=email).one()
     except NoResultFound:
-        return NotFoundError({'source': ''}, 'User not found').respond()
+        return make_response(jsonify(message="If your email was registered with us, you'll get an \
```
iamareebjamal: Log the email

shreyanshdwivedi:
@iamareebjamal I've updated the PR. Please review

app/api/auth.py
```diff
@@ -220,18 +220,17 @@ def reset_password_post():
     except TypeError:
         return BadRequestError({'source': ''}, 'Bad Request Error').respond()
 
-    try:
+    user_count = get_count(User.query.filter_by(email=email))
+    if user_count:
```
iamareebjamal: Now, it hits DB 2 times instead on one

shreyanshdwivedi: Actually to omit the duplicate messages I had to discard the try-except-else block. If except block will be there then there must be a message too. Any suggestions how can I remove duplication then?

iamareebjamal: >  If except block will be there then there must be a message too

Why?

shreyanshdwivedi: If I get exception NoObjectFound then there should be some code in the block, right? It shouldn't be empty

iamareebjamal: > Log the email

And you can also use `pass`, however, in this case you have to log the email

shreyanshdwivedi:
@iamareebjamal I'm logging email for both the cases so that user cannot differentiate from response that whether email is registered or not. Please review

app/api/auth.py
shreyanshdwivedi: I've used pass for `NoResultFound` case

iamareebjamal:
Logs are for us and not the user. I want to differentiate between valid and non valid emails

shreyanshdwivedi:
@iamareebjamal I have logged email in case of `NoResultFound` exception. Please have a look

app/api/auth.py
```diff
@@ -224,15 +225,16 @@ def reset_password_post():
     try:
         user = User.query.filter_by(email=email).one()
     except NoResultFound:
-        return NotFoundError({'source': ''}, 'User not found').respond()
+        logging.info(email)
```
iamareebjamal: Create a logger. Never use root logging module.

Also, logging the email makes no sense. What will I know from that. Properly log `Tried to reset password not existing email`

shreyanshdwivedi:
@iamareebjamal created a logger and logged the proper info. Please review

PR #6092, opened 2019-06-22
chore: Use development server on heroku for development config

iamareebjamal:


iamareebjamal:
We need to be able to debug apps depoyed on heroku. @CosmicCoder96 @niranjan94 Please tell if this can create any problems

CosmicCoder96:
@iamareebjamal I don't think this will be a problem -- just a question, does running the server on development limit it's scalability? Heroku is used for travis testing, often multiple PRs send a large amount of requests to it. If it can take that without timing out, it's fine I guess.

iamareebjamal:
@CosmicCoder96 Let's check if it creates a problem. If so, we'll revert

PR #6100, opened 2019-06-24
fix: Query for speakers properly for after event cron job

### LINKED ISSUE #6099: Speakers are queried by role in after event mail cron job but no such role exists
Currently, speakers are being queried under the assumption that there is a role called speaker, no such role exists.

CosmicCoder96:
Fixes #6099 

Queries speaker model by matching event id, instead of roles, since there is no role for speaker in the system.

### DIFF
--- a/app/api/helpers/scheduled_jobs.py
+++ b/app/api/helpers/scheduled_jobs.py
@@ -14,6 +14,7 @@ from app.models import db
 from app.models.event import Event
 from app.models.event_invoice import EventInvoice
 from app.models.order import Order
+from app.models.speaker import Speaker
 from app.models.session import Session
 from app.models.ticket import Ticket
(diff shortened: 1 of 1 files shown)


app/api/helpers/scheduled_jobs.py
```diff
@@ -33,7 +34,7 @@ def send_after_event_mail():
         upcoming_event_links += "</ul>"
         for event in events:
             organizers = get_user_event_roles_by_role_name(event.id, 'organizer')
-            speakers = get_user_event_roles_by_role_name(event.id, 'speaker')
+            speakers = Speaker.query.filter(Speaker.event_id == event.id).all()
```
iamareebjamal: deleted_at=None

PR #6136, opened 2019-07-03
fix: set constraints on creation/updation of order status

### LINKED ISSUE #6134: Client can create wrong status of an order
https://github.com/fossasia/open-event-attendee-android/pull/2032

As a client, I should never be able to create a wrong status on an order. 

For example, if the starting state of an order is 'initializing', client should not be able to set it as 'pending', or 'confirmed'. In fact, client should have minimal control over the order status, it should be automatically changed on backend depending on if the order is created, updated, free paid, etc. Giving this control to client is asking for errors and bugs to happen and as the system will be used by more users, it'll create security and monetary problems

Ensure that error is thrown in case the status of order is wrong and ensure that a status transition of the order is enforced.

For example, it should not be possible for an order to go from status 'pending' to 'intializing', or 'expired' to 'initializing'.

Ensure this behaviour with tests

shreyanshdwivedi:
Fixes #6134 

Currently, there is no check is status of order being created/updated is correct or not

 - Adds check on order status while creation/updation of the order

#### Checklist


app/api/orders.py
```diff
@@ -313,6 +318,14 @@ def before_update_object(self, order, data, view_kwargs):
             if order.order_notes and data['order_notes'] not in order.order_notes.split(","):
                 data['order_notes'] = '{},{}'.format(order.order_notes, data['order_notes'])
 
+        status = data.get('status', None)
+        if status and ((order.status == 'completed' and status in ['expired', 'initializing', 'pending', 'placed']) or
+            (order.status == 'expired' and status in ['completed', 'initializing', 'placed', 'pending']) or
```
mrsaicharan1: I think that this check can be optimised by making it into `if (status and order.status != status)`

app/api/orders.py
```diff
@@ -313,6 +318,14 @@ def before_update_object(self, order, data, view_kwargs):
...
+            (order.status == 'placed' and status in ['completed', 'expired', 'initializing', 'pending']) or
+                (order.status == 'pending' and status in ['initializing', 'placed'])):
```
mrsaicharan1: I think that this whole check can be optimized by making it into `if (status and order.status != status)`

app/api/orders.py
```diff
@@ -74,6 +74,10 @@ def before_create_object(self, data, view_kwargs):
         :param view_kwargs:
         :return:
         """
+        if data.get('status') == 'cancelled' or data.get('status') == 'expired':
```
iamareebjamal: OK, what if I set it to placed? Or hoola_boola?

shreyanshdwivedi: @iamareebjamal got your point.
I've a doubt, whenever any order is created on FE it is created with `initializing` status and when the user fills up the order form and hit on `Proceed` button a PATCH request is sent to update the status to either - `completed`, `placed` or `pending` and if user fails to fill it up in 15 minutes, it is changed to `expired`. 
So I just wanted to know should I only allow order with `initializing` status in `before_create_object`. Or should I allow status from - `'completed', 'initializing', 'pending', 'placed'`?

iamareebjamal: Ask yourself, can the client set an order to be completed or placed before it is actually placed or completed. If so, how can I prevent that?

Then ask, can a client fake an order to be filled and just send the PATCH request before actually filling required details. If so, how do I prevent that?

In the last step, ask yourself. Can a client set an expired/pending/completed/placed order to be initializing state or any previous state. If so, how do I prevent that?

In short you have to ensure that an order follows a model of [Deterministic Finite State Automata](https://en.wikipedia.org/wiki/Deterministic_finite_automaton)

Order should have a starting state and final state and can only go from one state to another state with a precondition and not change after it is in final state

shreyanshdwivedi: I'm planning to implement this -
 - Allow only `initializing` status in `before_create_object` (It'll also block any creation of order object by making direct API call) (POST request)
 - Valid status transitions (PATCH request) :
```
1. initializing  ->  completed, expired, placed, pending
2. completed  ->  cancelled (only non-paid orders)
3. expired  ->  None
4. placed  ->  cancelled 
5. pending  ->  completed, expired
6. cancelled  ->  None
```

iamareebjamal: Even then the transitions should be validated

shreyanshdwivedi: @iamareebjamal please review the checks applied on the transitions

shreyanshdwivedi:
@iamareebjamal can you please review this PR.

shreyanshdwivedi:
@iamareebjamal please review

app/api/orders.py
```diff
@@ -324,6 +328,21 @@ def before_update_object(self, order, data, view_kwargs):
             if order.order_notes and data['order_notes'] not in order.order_notes.split(","):
                 data['order_notes'] = '{},{}'.format(order.order_notes, data['order_notes'])
 
+        status = data.get('status', None)
+        if status and status not in ['cancelled', 'completed', 'expired', 'initializing', 'pending', 'placed']:
+            raise ConflictException({'pointer': '/data/attributes/status'},
+                                    "status '{}' is invalid".format(status))
```
mrsaicharan1: please remove inner quotes for `'{}'`

shreyanshdwivedi: Actually I wanted to have error like this - `status 'abcd' is invalid?`
However I think this isn't necessary. Updating it right away :)

iamareebjamal:
@mrsaicharan1 @uds5501 @kushthedude @prateekj117 Please test this PR. This is very important that this works correctly

kushthedude:
@shreyanshdwivedi Just came with a bug with this implementation : 
- Create a Placed Order.
- Mark it as completed on the dashboard, It does not work

shreyanshdwivedi:
@kushthedude I didn't realize that we are allowing update of placed orders to completed. Thanks for pointing it out. I've updated the checks
Please have a look again

shreyanshdwivedi:
@kushthedude @uds5501 @mrsaicharan1 please review this PR. It needs proper testing

iamareebjamal:
@kushthedude @mrsaicharan1 @uds5501 @prateekj117 Test this on priority. I want to merge this by tomorrow. So test each and every possible scenario using FE and postman

It has been a month

uds5501:
@iamareebjamal Tested the FE functionality, each transition is working fine.

iamareebjamal:
@kushthedude @prateekj117 @mrsaicharan1 Please test on priority

CosmicCoder96:
> Implement a check if the final state of order and the initial state is the same. Like for cancelling a cancelled order, I should get 500(Invalid Transition Sort of thing).

Under no circumstance do we intentionally generate a 500 @kushthedude

kushthedude:
> > Implement a check if the final state of order and the initial state is the same. Like for cancelling a cancelled order, I should get 500(Invalid Transition Sort of thing).
> 
> Under no circumstance do we intentionally generate a 500 @kushthedude

405 can be better for the same transition, In the same, I will implement checks on FE that delete or cancel button don't get shown for expired or cancelled orders.

shreyanshdwivedi:
> Made multiple transitions and manipulated multiple states of order.
> Some notes :
> 
>     * Implement a check if the final state of order and the initial state is the same. Like for cancelling a cancelled order, I should get 500(Invalid Transition Sort of thing).
> 
>     * There should not be any final state for deleted order If any transition is getting performed on `Deleted Order` throw an appropriate error except this.
> 
>     * Expired Order should not be allowed to change state like stated above, Eg: If I cancel an expired order I should get the invalid transition.
>       If above-stated points are ok, Its good to be merged.
>       @iamareebjamal @shreyanshdwivedi

@iamareebjamal can you please share your views on these points?

iamareebjamal:
Last 2 points are valid. 1st is not needed if all correct transitions are implemented

app/api/orders.py
```diff
@@ -385,6 +389,22 @@ def before_update_object(self, order, data, view_kwargs):
...
+        status = data.get('status', None)
+        if status and status not in ['cancelled', 'completed', 'expired', 'initializing', 'pending', 'placed']:
+            raise ConflictException({'pointer': '/data/attributes/status'},
+                                    "status {} is invalid".format(status))
+        elif status and status != 'cancelled':
+            if (order.status in ['completed', 'expired', 'cancelled'] and order.status != status) or \
+                    (order.status == 'placed' and status not in ['completed', 'placed']) or \
+                    (order.status == 'pending' and status not in ['completed', 'expired', 'pending']) or \
+                    (order.status == 'initializing' and status not in ['completed', 'expired', 'pending', 'placed']):
+                raise ConflictException({'pointer': '/data/attributes/status'},
```
shreyanshdwivedi: Regarding 3rd point, we cannot update the status of expired order to any other status. Neither cancelled nor any other @iamareebjamal @kushthedude 
However keeping second point in my mind, I'll add checks to avoid status transition for deleted orders

shreyanshdwivedi: @kushthedude were you able to change status of expired orders too because I think there is a check to avoid that

shreyanshdwivedi:
@kushthedude @iamareebjamal please have a look.

shreyanshdwivedi:
@uds5501 @mrsaicharan1 @iamareebjamal please review this PR

iamareebjamal:
There is a release pending, so we can't add this to the release since we won't have enough time to test it after merging and it is a breaking change. 

But we need reviews ASAP @kushthedude @uds5501 @mrsaicharan1 @prateekj117

iamareebjamal:
Don't just review the implementation. Pull the changes and try to break the site

kushthedude:
@shreyanshdwivedi I am still getting this when I cancel my already cancelled order, Shouldnt it throw 422 ?
Screenshot 2019-08-22 at 11 56 19 AM

kushthedude:
> * Expired Order should not be allowed to change state like stated above, Eg: If I cancel an expired order I should get the invalid transition.

Also this is still present @shreyanshdwivedi

shreyanshdwivedi:
@kushthedude did you checked by pulling the changes? I don't think so because I'm getting the expected results.


> @shreyanshdwivedi I am still getting this when I cancel my already cancelled order, Shouldnt it throw 422 ?
> Screenshot 2019-08-22 at 11 56 19 AM

When I cancelled an already cancelled order, proper error popped up -
Screenshot from 2019-08-30 16-02-47



> > * Expired Order should not be allowed to change state like stated above, Eg: If I cancel an expired order I should get the invalid transition.
> 
> Also this is still present @shreyanshdwivedi

Also, the expired orders aren't allowed to change their state. The error when I sent a PATCH request using postman to change its state to cancelled/completed -
```
{
    "errors": [
        {
            "status": 409,
            "source": {
                "pointer": "/data/attributes/status"
            },
            "title": "Conflict",
            "detail": "status of expired order cannot be changed to cancelled"
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}
```

Please check again and let me know if you face any other discrepancies. This PR needs to be finalized now. It has been long 
@uds5501 @mrsaicharan1 please test it out on local

kushthedude:
@iamareebjamal I think this PR can get merged now its more than 2 months for this PR.

iamareebjamal:
@prateekj117 Test this please and report before EOD

prateekj117:
@iamareebjamal Ok. Will do.

app/api/orders.py
```diff
@@ -115,6 +115,9 @@ def before_create_object(self, data, view_kwargs):
             data['is_billing_enabled'] = True
 
         free_ticket_quantity = 0
+        if data.get('status') != 'initializing':
+            raise ConflictException({'pointer': '/data/attributes/status'},
+                                    "status cannot be {} while creating an order".format(data.get('status')))
```
prateekj117: @iamareebjamal Not sure if we need this check again as we are already making sure that status of the order is `initializing` in `before_post` function above in line 104.

prateekj117:
@shreyanshdwivedi Was successfully able to change the status of an unpaid order from `pending`/`initializing` to `completed`.

shreyanshdwivedi:
@prateekj117 the state change from pending to completed is a valid transition, that's why you were able to change it.

prateekj117:
@shreyanshdwivedi "Unpaid order (of a paid ticket)".

shreyanshdwivedi:
I don't think you fully understand the issue being solved here. This PR just adds checks that prevents any wrong updation of status. For unpaid ticket i.e. with pending status, completed status is a correct transition. Why should it not allow this transition? What exactly is your concern here?

prateekj117:
@shreyanshdwivedi An `initial`/`pending` order for a paid ticket, should not be updated to `completed` unless the amount has been paid. Got it? If yes, read my above statement again.

prateekj117:
You wrote "For unpaid ticket, i.e., for pending status", you need to understand that these are two different things.

prateekj117:
For example, say someone initiates a paid order, can simply update the status to `completed` using a simple API request.

shreyanshdwivedi:
I think you still need to understand this thing, "I'm just adding checks to prevent wrong status transition". And this issue was never to check payment status before making transition.
Just a small check to ensure that no wrong updation is made e.g. expired order being marked as completed etc.
I think you'll understand the issue and PR now.

iamareebjamal:
@shreyanshdwivedi This is a serious bug if I can create a paid order and just mark it completed without paying anything.

shreyanshdwivedi:
@prateekj117 did you check without admin/organizer rights? Are u able to update status even then? If yes, then it needs to be fixed. But please check first and report

prateekj117:
@shreyanshdwivedi Yes.

iamareebjamal:
> For example, if the starting state of an order is 'initializing', client should not be able to set it as 'pending', or 'confirmed'. In fact, client should have minimal control over the order status, it should be automatically changed on backend depending on if the order is created, updated, free paid, etc. Giving this control to client is asking for errors and bugs to happen and as the system will be used by more users, it'll create security and monetary problems

This is what I wrote in original issue. Read completely, especially the last line. I was simply never aware that there is no check to see if the order has been paid for. This is exactly why this issue was created!

iamareebjamal:
@prateekj117 Check via another user if you are able to do so. Create a new user and report

shreyanshdwivedi:
Even I'm unaware that this is a flaw in the order API. I need to confirm once that we allow user without admin/organizer access to update the order status.

prateekj117:
@iamareebjamal @shreyanshdwivedi The user I tried with did not have admin/organizer/co-organizer access.

shreyanshdwivedi:
Ok, so the checks are missing the has_access condition. I think all the transition involving payment status update or cancelling tickets should involve a has_access part to allow co-organizers only.
But I'm a bit busy with my academics currently and can work on this after a week maybe.

shreyanshdwivedi:
@prateekj117 if you are able to work on this, I can close this PR and you can take over and add the remaining conditions and open a new PR.
Let me know if you want to take this up

prateekj117:
@shreyanshdwivedi Ok sure.

shreyanshdwivedi:
Closing this because of my unavailability due to exams. Will be worked upon by @prateekj117 in a new PR.

PR #6138, opened 2019-07-03
feat: implement admin restrictions on deleted orders tab 

uds5501:
TBD : Fixes #

Ensure that `events/<event_identifier>/orders?get_trashed=true` (the queries requesting deleted order objects) do not pass without admin authentication

#### Checklist

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -194,6 +194,8 @@ class OrdersList(ResourceList):
         :param kwargs:
         :return:
         """
+        if 'get_trashed=true' in str(request.query_string) and not has_access('is_admin'):
+            raise ForbiddenException({'source': ''}, "Admin Access is Required")
         if kwargs.get('event_id') and not has_access('is_coorganizer', event_id=kwargs['event_id']):
             raise ForbiddenException({'source': ''}, "Co-Organizer Access Required")
 


app/api/orders.py
```diff
@@ -194,6 +194,8 @@ def before_get(self, args, kwargs):
         :param kwargs:
         :return:
         """
+        if 'get_trashed=true' in str(request.query_string) and not has_access('is_admin'):
```
iamareebjamal: That's a terrible way to check this

uds5501: @iamareebjamal I was actually looking for something which could indicate access for deleted objects. Any suggestion would be helpful

@mrsaicharan1 @shreyanshdwivedi please see

mrsaicharan1: @uds5501 I guess it's better to use `parse_qs` from `requests` library

uds5501: @mrsaicharan1 okay, taking a look at it

iamareebjamal: Why can't you use kwargs?

uds5501: @iamareebjamal because the kwargs are only passing this : `{'event_id': 1}`

iamareebjamal: Where does flask access get_trashed=true?

uds5501: It doesn't. The `get_trashed=true` is supposed to work on flask-rest-json api fork I think.
In this `before_get` hook, args is empty and the kwargs is being passed as I mentioned above

PR #6145, opened 2019-07-05
feat: Serving event invoices through a protected route

### LINKED ISSUE #6144: Implementation of a route to serve Event Invoices
An event invoice should be served over a route which allows only the organizer to access their respective event invoices. 

Similar to tickets & order invoice serving.

**Additional context**
Working on it.

mrsaicharan1:
Fixes #6144 

This route serves event invoices in the same way the order invoices and tickets are served. It allows event organizers to access their respective event invoices. 
Allows staff members to download/view them.

- Added route to serve event invoices
- Added security checks to enable access-control for these invoices & only allow authenticated event organisers & staff members to access them.

#### Checklist

- [ ] All the functions created/modified in this PR contain relevant docstrings

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -10,6 +10,7 @@ from flask import request, jsonify, make_response, Blueprint, send_file
 from flask_jwt import current_identity as current_user, jwt_required
 from flask_limiter.util import get_remote_address
 from healthcheck import EnvironmentDump
+from flask_rest_jsonapi.exceptions import ObjectNotFound
 from sqlalchemy.orm.exc import NoResultFound
 
 from app import get_settings
@@ -32,6 +33,8 @@ from app.models.mail import PASSWORD_RESET, PASSWORD_CHANGE, \
 from app.models.notification import PASSWORD_CHANGE as PASSWORD_CHANGE_NOTIF
 from app.models.order import Order
 from app.models.user import User
+from app.models.event_invoice import EventInvoice
+
 
 logger = logging.getLogger(__name__)
 authorised_blueprint = Blueprint('authorised_blueprint', __name__, url_prefix='/')
@@ -295,9 +298,9 @@ def change_password():
     })
 
 
-def return_file(file_name_prefix, file_path, order_identifier):
+def return_file(file_name_prefix, file_path, identifier):
     response = make_response(send_file(file_path))
-    response.headers['Content-Disposition'] = 'attachment; filename=%s-%s.pdf' % (file_name_prefix, order_identifier)
+    response.headers['Content-Disposition'] = 'attachment; filename=%s-%s.pdf' % (file_name_prefix, identifier)
     return response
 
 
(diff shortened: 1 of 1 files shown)


mrsaicharan1:
@iamareebjamal @uds5501 @shreyanshdwivedi Please have a look at this as it is ready now. 
Also, @shreyanshdwivedi , I think you can test your basic design layout for the event invoice once this is merged.

poush:
The title of this PR seems misleading xD. Instead of "secure", it can be simply "protected" or "restricted"

app/api/auth.py
```diff
@@ -345,6 +348,32 @@ def order_invoices(order_identifier):
         return ForbiddenError({'source': ''}, 'Authentication Required to access Invoice').respond()
 
 
+@ticket_blueprint.route('/events/invoices/<string:invoice_identifier>')
+@jwt_required()
+def event_invoices(invoice_identifier):
+    if current_user:
+        try:
+            event_invoice = EventInvoice.query.filter_by(identifier=invoice_identifier).first()
```
poush: Can we merge this and line below as 
`.first().event_id`

as NoResultFound will be be raise in any case. Let me know if that may not work or you think it will be a bad idea.

app/api/auth.py
```diff
@@ -345,6 +348,32 @@ def order_invoices(order_identifier):
         return ForbiddenError({'source': ''}, 'Authentication Required to access Invoice').respond()
 
 
+@ticket_blueprint.route('/events/invoices/<string:invoice_identifier>')
+@jwt_required()
+def event_invoices(invoice_identifier):
+    if current_user:
```
poush: There are better design methods that you can consider. For eg:

```
if not current_user:
    raise Forbidden error
```
There will be no unnecessary if-else ladder

app/api/auth.py
mrsaicharan1: @poush This will throw an `AttributeError` as the SQLAlchemy attempts to find `event_id` on the `NoneType` object. Ergo, I wouldn't recommend merging the two :)

app/api/auth.py
mrsaicharan1: @poush This if-else ladder is required because at each stage, we are validating the user & the corresponding permissions. If there is any violation at any point, we throw the appropriate exception.

The example given by you is more or less the equivalent of my logic. If we use 
```
if not current_user:
    raise Forbidden error
```,
we would then have to proceed with the checks  in the else block similarly. 
```
else if current__user.is_staff and current_user.is_verified:
                ...
```

Comment (mrsaicharan1) [reply app/api/auth.py]: @poush You can check the implementation given for tickets & order invoices which is akin to what I've proposed here :D

Comment (iamareebjamal) [reply app/api/auth.py]: > we would then have to proceed with the checks in the else block

No you won't 
You can assume from the next line that current_user is present, removing the need of nested conditions

Comment (mrsaicharan1) [reply app/api/auth.py]: > > we would then have to proceed with the checks in the else block
> 
> No you won't
> You can assume from the next line that current_user is present, removing the need of nested conditions

Won't the nesting still be present to check for current_user.is_verified & other cases?

Comment (iamareebjamal) [reply app/api/auth.py]:
```python
if current_user:
  # 1 level nesting
  if current_user.is_verified:
    # 2 level
  else:
    # 2 level
    throw
else:
  # 1 level
  throw
```
Max Level 2
Conditions 4

```python
if not current_user:
  throw

if current_user.is_verified:
  throw

# continue
```

Max level nesting: 0
Conditions 2

mrsaicharan1: @iamareebjamal @poush Understood! Will be updating this.

mrsaicharan1:
@poush @iamareebjamal Removed the additional nesting.

app/api/auth.py
```diff
@@ -345,6 +348,28 @@ def order_invoices(order_identifier):
...
+@jwt_required()
+def event_invoices(invoice_identifier):
+    if not current_user:
+        return ForbiddenError({'source': ''}, 'Authentication Required to access Invoice').respond()
+    try:
+        event_invoice = EventInvoice.query.filter_by(identifier=invoice_identifier).first()
+        event_id = event_invoice.event_id
+    except NoResultFound:
+        return NotFoundError({'source': ''}, 'Event Invoice not found').respond()
+    if not current_user.is_organizer(event_id) or not current_user.is_staff:
```
poush: The logic is incorrect here. the whole if statement should be true if user is both not the organizer and not from staff.
Instead of "or" it should be "and"

mrsaicharan1: @poush Yeah, you're right. I seem to have overlooked that. I've made the change!

mrsaicharan1:
@iamareebjamal @poush THis is ready for another review

PR #6154, opened 2019-07-07
fix: allows hard delete for feedbacks

### LINKED ISSUE #6153: No hard delete for feedback
**Describe the bug**

Currently there is hard delete for feedback. When a user sets his/her rating to 0 again, the feedback of the user should be removed completely from the db.
This issue needs to be solved to merge https://github.com/fossasia/open-event-frontend/pull/3198

**Expected behavior**

When the user clears his/her rating (i.e. sets it to 0 again), the feedback must be removed.

**Additional context**

On it

shreyanshdwivedi:
Fixes #6153 

Currently there is hard delete for feedback. 

- Hard deletes the feedback if deleted_at is not NULL
- Changes range from `0-5` to `1-5` for rating

#### Checklist

### DIFF
--- a/app/api/feedbacks.py
+++ b/app/api/feedbacks.py
@@ -5,6 +5,7 @@ from flask_jwt import current_identity as current_user
 from app.api.bootstrap import api
 from app.api.helpers.db import safe_query
 from app.api.helpers.exceptions import UnprocessableEntity, ForbiddenException
+from app.api.helpers.feedback import delete_feedback
 from app.api.helpers.permission_manager import has_access
 from app.api.helpers.permissions import jwt_required
 from app.api.helpers.query import event_query
@@ -133,6 +134,12 @@ class FeedbackDetail(ResourceDetail):
             if session and not has_access('is_coorganizer', event_id=session.event_id):
                 raise ForbiddenException({'source': ''},
                                          "Event co-organizer access required")
+        if feedback and data.get('deleted_at'):
+            if has_access('is_user_itself', user_id=feedback.user_id):
+                delete_feedback(feedback)
+            else:
+                raise ForbiddenException({'source': ''},
+                                         "Feedback can be deleted only by user himself")
 
     decorators = (api.has_permission('is_user_itself', fetch='user_id',
                                      fetch_as="user_id", model=Feedback, methods="PATCH,DELETE"),)
--- a/app/api/helpers/feedback.py
+++ b/app/api/helpers/feedback.py
@@ -1,0 +1,18 @@
+import logging
+from app.models import db
+
+logger = logging.getLogger(__name__)
+
+
+def delete_feedback(feedback):
+    """
+    Delete the feedback if rating is zero
+    :param feedback: Feedback to be deleted
+    :return:
+    """
+    db.session.delete(feedback)
+    try:
+        db.session.commit()
+    except Exception:
+        logger.exception('Feedback delete exception!')
+        db.session.rollback()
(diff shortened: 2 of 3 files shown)


app/api/helpers/feedback.py
```diff
@@ -0,0 +1,16 @@
...
+from app.models import db
+
+
+def delete_feedback(feedback):
+    """
+    Delete the feedback if rating is zero
+    :param feedback: Feedback to be deleted
+    :return:
+    """
+    db.session.delete(feedback)
```
mrsaicharan1: Just a question, Where are we checking if the feedback is 0?

shreyanshdwivedi: @mrsaicharan1 when rating is set to 0 on FE, `destroyRecord()` is called which sets `deleted_at` to non null value and then the rest is handled by the checks I introduced -
```
if feedback and 'deleted_at' in data and data['deleted_at']:
            if has_access('is_user_itself', user_id=feedback.user_id):
                delete_feedback(feedback)
            else:
                raise ForbiddenException({'source': ''},
                                         "Feedback can be deleted only by user himself")
```

app/api/feedbacks.py
```diff
@@ -133,6 +134,12 @@ def before_update_object(self, feedback, data, view_kwargs):
             if session and not has_access('is_coorganizer', event_id=session.event_id):
                 raise ForbiddenException({'source': ''},
                                          "Event co-organizer access required")
+        if feedback and 'deleted_at' in data and data['deleted_at']:
```
iamareebjamal:
```suggestion
        if feedback and data.get('deleted_at'):
```

app/api/helpers/feedback.py
```diff
@@ -0,0 +1,16 @@
...
+    try:
+        db.session.commit()
+    except Exception as e:
+        logging.error('DB Exception! %s' % e)
```
iamareebjamal: Never ever use root logging, create a logger

iamareebjamal: logger.exception

app/api/schema/feedbacks.py
```diff
@@ -23,7 +23,7 @@ class Meta:
         inflect = dasherize
 
     id = fields.Str(dump_only=True)
-    rating = fields.Float(required=True, validate=Range(min=0, max=5))
+    rating = fields.Float(required=True, validate=Range(min=1, max=5))
```
iamareebjamal: This behaviour of frontend deleting the record when setting the rating to 0 is wrong IMO. Delete action should be very explicit. Either show a dialog on frontend that setting rating to 0 will delete the feedback

shreyanshdwivedi: FE is using ui-rating from semantic UI. If a user selects rating 2 suppose and wants to clear out the rating, he/she needs to click on 2 stars again and the rating is cleared (set to 0). I think showing a dialog box will be better.
@CosmicCoder96 your opinion?

app/api/helpers/feedback.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    Delete the feedback if rating is zero
+    :param feedback: Feedback to be deleted
+    :return:
+    """
+    db.session.delete(feedback)
+    try:
+        db.session.commit()
+    except Exception as e:
+        logging.exception('DB Exception! %s' % e)
```
iamareebjamal: Still using root logging module.

Also, DB Exception is useless. It doesn't tell where the error happened.

Also, no need to include e in the log, stacktrace is automatically printed

shreyanshdwivedi: Ah, missed it. Fixed it now

app/api/schema/feedbacks.py
shreyanshdwivedi: @iamareebjamal please review this PR. The FE PR is still WIP so I can include your suggestions there

PR #6184, opened 2019-07-12
fix: order deletion triggers emails

uds5501:
Fixes fossasia/open-event-frontend#3269

Ensures that deletion of orders doesn't trigger notifications and emails

#### Checklist

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -323,15 +323,14 @@ class OrderDetail(ResourceDetail):
         # create pdf tickets.
         create_pdf_tickets_for_holder(order)
 
-        if order.status == 'cancelled':
+        if order.status == 'cancelled' and order.deleted_at is None:
             send_order_cancel_email(order)
             send_notif_ticket_cancel(order)
 
             # delete the attendees so that the tickets are unlocked.
             delete_related_attendees_for_order(order)
 
-        elif order.status == 'completed' or order.status == 'placed':
-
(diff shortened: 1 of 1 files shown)


uds5501:
@mrsaicharan1 @shreyanshdwivedi @prateekj117 Please take a look

kushthedude:
@uds5501 , There should be an email to attendee perhaps

uds5501:
@kushthedude deletion of an order doesn't change any status so, IMO there should be none. 
Deletion of the order is organizer activity and only unpaid or cancelled tickets can be deleted (AFAIK) so attendee discretion isn't really necessary here.

kushthedude:
> @kushthedude deletion of an order doesn't change any status so, IMO there should be none.
> Deletion of the order is organizer activity and only unpaid or cancelled tickets can be deleted (AFAIK) so attendee discretion isn't really necessary here.

Let us take this case, Some user has ordered a bunch of `Offline Tickets`, Due to some reason `Organiser` deletes there order, How will the user know his order does not exist now. I totally agree with you that `Cancelled or Expired or Pending` order should deliver no notification upon deletion .

uds5501:
@kushthedude as far as notifying them that their order has been deleted, it'll take a new notification and email.
In this particular issue however, what I am trying to point out is that if we allow this behaviour to persist, the user simply gets notified that that their ticket has been placed instead of deleted. Maybe you can open up a new issue for the one you pointed out.

PR #6192, opened 2019-07-14
fix: make billing info mandatory for paid orders

### LINKED ISSUE #6137: Make billing information mandatory for paid orders

shreyanshdwivedi:
Fixes #6137 

Make billing information mandatory for paid orders

- Adds checks in `before_create_object` and `before_update_object` in `api/orders.py`

#### Checklist

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -63,6 +63,16 @@ def is_payment_valid(order, mode):
         return (order.paid_via == 'paypal') and order.transaction_id
 
 
+def check_billing_info(data):
+    if data.get('amount') > 0 and not data.get('is_billing_enabled'):
+        raise UnprocessableEntity({'pointer': '/data/attributes/is_billing_enabled'},
+                                  "Billing information is mandatory for paid orders")
+    if data.get('is_billing_enabled') and not (data.get('company') and data.get('address') and data.get('city') and
+                                               data.get('zip') and data.get('country')):
+        raise UnprocessableEntity({'pointer': '/data/attributes/is_billing_enabled'},
+                                  "Billing information incomplete")
+
+
 class OrdersListPost(ResourceList):
     """
     OrderListPost class for OrderSchema
@@ -96,6 +106,7 @@ class OrdersListPost(ResourceList):
         :param view_kwargs:
         :return:
         """
+        check_billing_info(data)
 
         free_ticket_quantity = 0
 
@@ -299,6 +310,7 @@ class OrderDetail(ResourceDetail):
         :param view_kwargs:
         :return:
         """
+        check_billing_info(data)
         if (not has_access('is_coorganizer', event_id=order.event_id)) and (not current_user.id == order.user_id):
             raise ForbiddenException({'pointer': ''}, "Access Forbidden")
 


app/api/orders.py
```diff
@@ -104,6 +104,14 @@ def before_create_object(self, data, view_kwargs):
             raise ConflictException({'pointer': '/data/attributes/amount'},
                                     "Amount cannot be null for a paid order")
 
+        if data.get('payment_mode') in ['stripe', 'paypal', 'alipay', 'omise'] and not data.get('is_billing_enabled'):
+            raise ConflictException({'pointer': '/data/attributes/is_billing_enabled'},
+                                    "Billing information is mandatory for paid orders")
+        if data.get('is_billing_enabled') and not (data.get('company') and data.get('address') and data.get('city') and
+                data.get('zip') and data.get('country')):
+            raise ConflictException({'pointer': '/data/attributes/is_billing_enabled'},
+                                    "Billing information incomplete")
```
kushthedude: Use the same above message for this too

shreyanshdwivedi: Actually, both the checks imply different thing and so they have different messages.
In the above case - The user hasn't ticked the checkbox `Enable Billing Information` on the order page while placing a paid order.
While in the second one, the user have checked the checkbox but one or more information is missing. @kushthedude

iamareebjamal:
I think you made a method which checks the payment mode, use that. If there's no method, create one to avoid repetition

shreyanshdwivedi:
Yeah that function was to check if any of the payment option is enabled in `event`, both online and offline form of payments.
But here I need to check if the order is paid via our platform i.e. stripe, omise, alipay or paypal. When order is crrated, I don't think that we can check it via a method. However while updating, we might check via a method.

shreyanshdwivedi:
@iamareebjamal should I add a new function just to check if payment options are from stripe, omise, alipay or paypal?

iamareebjamal:
I see repeated code. There should be one function called from 2 places instead of repetition

shreyanshdwivedi:
@iamareebjamal made a function `check_billing_info` to remove the code duplication. Please have a look

iamareebjamal:
Have a peer review

shreyanshdwivedi:
@mrsaicharan1 @uds5501 @kushthedude please review

kushthedude:
@shreyanshdwivedi There is one hound violation.

app/api/orders.py
```diff
@@ -40,6 +40,16 @@
 alipay_blueprint = Blueprint('alipay_blueprint', __name__, url_prefix='/v1/alipay')
 
 
+def check_billing_info(data):
+    if data.get('payment_mode') in ['stripe', 'paypal', 'alipay', 'omise'] and not data.get('is_billing_enabled'):
```
uds5501: wouldn't `data.get('amount') > 0` be a better criteria to check for paid orders? @shreyanshdwivedi

shreyanshdwivedi: I think u are right. I thought that amount can be zero even due to discount code and got confused so didn't use this 
However u're right here. Will update it asap

app/api/orders.py
```diff
@@ -39,6 +39,15 @@
 order_misc_routes = Blueprint('order_misc', __name__, url_prefix='/v1')
 alipay_blueprint = Blueprint('alipay_blueprint', __name__, url_prefix='/v1/alipay')
 
+def check_billing_info(data):
+    if data.get('payment_mode') in ['stripe', 'paypal', 'alipay', 'omise'] and not data.get('is_billing_enabled'):
+        raise ConflictException({'pointer': '/data/attributes/is_billing_enabled'},
+                                "Billing information is mandatory for paid orders")
+    if data.get('is_billing_enabled') and not (data.get('company') and data.get('address') and data.get('city') and
+            data.get('zip') and data.get('country')):
```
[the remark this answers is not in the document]
shreyanshdwivedi: @uds5501 @mrsaicharan1 I'm unable to fix this. Can u guys help?

app/api/orders.py
shreyanshdwivedi: @uds5501 I thought about your suggestion but I think we should not use amount here. Placed tickets also have amount > 0 but we only need billing Info for the tickets paid via our platform. What do u think about this?

uds5501: In that case, I think this should work just fine.

shreyanshdwivedi:
@uds5501 @mrsaicharan1 @kushthedude please review

app/api/orders.py
```diff
@@ -40,6 +40,16 @@
 alipay_blueprint = Blueprint('alipay_blueprint', __name__, url_prefix='/v1/alipay')
 
 
+def check_billing_info(data):
+    if data.get('payment_mode') in ['stripe', 'paypal', 'alipay', 'omise'] and not data.get('is_billing_enabled'):
+        raise ConflictException({'pointer': '/data/attributes/is_billing_enabled'},
```
iamareebjamal: How is this a Conflict? This is 400 Bad Request

shreyanshdwivedi: @iamareebjamal  Currently, we don't have BadRequestException with 400 status in `helpers/exceptions.py`
Should I add it or use `UnprocessableEntity` instead? However, 422 status is responded from the server if we use UnprocessableEntity which states that syntax is correct.

iamareebjamal: Yes, I meant to say 422

shreyanshdwivedi:
@iamareebjamal please review. Updated the exception

shreyanshdwivedi:
@CosmicCoder96 I have a query. The issue for which I opened this PR doesn't state whether the billing Info is required for only orders paid via our platform or even for offline paid event. Do you want it for both online as well as offline paid events?

shreyanshdwivedi:
Keep it on hold till @CosmicCoder96 replies to https://github.com/fossasia/open-event-server/pull/6192#issuecomment-515655226

shreyanshdwivedi:
@iamareebjamal I've just updated the check to include all type of paid orders - `data.get('amount') > 0` as we are following this on FE too.

PR #6203, opened 2019-07-16
fix: Making Event Name mandatory for publishing of event

### LINKED ISSUE #6202: Events without name can be published
Disable publishing of event without a name .

kushthedude:
Fixes #6202

### DIFF
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -79,6 +79,10 @@ def validate_event(user, modules, data):
         raise ConflictException({'pointer': '/data/attributes/location-name'},
                                 "Online Event does not have any locaton")
 
+    if not data.get('name', None) and data.get('state', None) == 'published':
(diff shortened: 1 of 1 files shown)


kushthedude:
@uds5501 @mrsaicharan1 @iamareebjamal Please Review

app/api/events.py
```diff
@@ -79,6 +79,10 @@ def validate_event(user, modules, data):
         raise ConflictException({'pointer': '/data/attributes/location-name'},
                                 "Online Event does not have any locaton")
 
+    if not data.get('name', None) and data.get('state', None) == 'published':
```
iamareebjamal: Why `not data.get('name', None)`?

kushthedude: Because event name should not be empty

PR #6211, opened 2019-07-18
[wip] feat: Making Tickets Mandatory for Publishing Event

### LINKED ISSUE #6210: Event Wizard: Require Tickets before Event Publishing
See https://github.com/fossasia/open-event-frontend/issues/2482 

Make sure `Event` has tickets available before ` Publishing`

kushthedude:
Fixes #6210


kushthedude:
@uds5501 @mrsaicharan1 @shreyanshdwivedi Please Review

app/api/events.py
```diff
@@ -70,6 +70,13 @@ def validate_event(user, modules, data):
         raise ForbiddenException({'source': ''},
                                  "Only verified accounts can publish events")
 
+    if data.get('state', None) == 'published':
+        event = Event.query.filter_by(name=data.get('name', None)).first()
+        ticket_available = event.tickets_available
+        if (ticket_available == 0):
+            raise ForbiddenException({'source': ''},
```
iamareebjamal: Forbidden is wrong error, it's mainly used for authorization errors. And point source correctly

iamareebjamal:
Wasn't this done already by @shreyanshdwivedi or @uds5501 ?

kushthedude:
> Wasn't this done already by @shreyanshdwivedi or @uds5501 ?

@iamareebjamal That was just an FE check, Which can be bypassed so easily by making an event draft and then publishing it from the dashboard , That's why couldn't rely on it.

app/api/events.py
kushthedude: @iamareebjamal Please Check now

iamareebjamal:
> That was just an FE check

I was talking about API check, but OK

kushthedude:
@uds5501 @mrsaicharan1 @shreyanshdwivedi Please Review

kushthedude:
> @kushthedude I tried this on my local. The server is raising 500 POST error
> `AttributeError: 'NoneType' object has no attribute 'tickets_available'`

It was perfect when I checked, can you tell the steps which made you get the following messafe?

shreyanshdwivedi:
@kushthedude I just added your check on my local and tried to create event without tickets on FE

kushthedude:
image
@shreyanshdwivedi Please check it, I just tried publishing another event without a ticket.

kushthedude:
@mrsaicharan1 @uds5501 Can you please verify if you get the correct error after publishing a draft event?

> image
> @shreyanshdwivedi Please check it, I just tried publishing another event without a ticket.

shreyanshdwivedi:
That's weird. I checked again. Please check this. @kushthedude 
Steps - I only filled event name and location and clicked on publish event

Request - `POST: http://127.0.0.1:5000/v1/events`
```
Host: 127.0.0.1:5000
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0
Accept: */*
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
Referer: http://localhost:4200/create
content-type: application/vnd.api+json
authorization: JWT *********
Content-Length: 1086
Origin: http://localhost:4200
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
```

Response - `'NoneType' object has no attribute 'tickets_available'`

Screenshot from 2019-07-20 16-19-40

kushthedude:
@shreyanshdwivedi I will check it ASAP

app/api/events.py
```diff
@@ -70,6 +70,13 @@ def validate_event(user, modules, data):
         raise ForbiddenException({'source': ''},
                                  "Only verified accounts can publish events")
 
+    if data.get('state', None) == 'published':
+        event = Event.query.filter_by(name=data.get('name', None)).first()
+        ticket_available = event.tickets_available
+        if (ticket_available == 0):
```
mrsaicharan1: You could make this as 
```
if event.tickets:
    ....
```

app/api/events.py
```diff
@@ -70,6 +70,13 @@ def validate_event(user, modules, data):
         raise ForbiddenException({'source': ''},
                                  "Only verified accounts can publish events")
 
+    if data.get('state', None) == 'published':
+        event = Event.query.filter_by(name=data.get('name', None)).first()
```
uds5501: @kushthedude this will simply break in `before_post` call when there is no event to query.

app/api/events.py
shreyanshdwivedi: I agree to @mrsaicharan1. We already face this a lot, you need to check if anything exists first then try to retrieve it

kushthedude: I will make appropriate changes

shreyanshdwivedi:
@kushthedude can you update this PR?

kushthedude:
> @kushthedude can you update this PR?

Will finalise today

app/api/events.py
kushthedude: @uds5501 Where should it be put first ?

uds5501: @kushthedude IMO it shouldn't be placed in before_post at least. Try and test it. See if it even works.

uds5501:
@kushthedude Have you thought this through? 
An event post won't be including any data about the `tickets`. This is why we handled it on the client sides because during generation of event, both `event` model and `ticket` models are different and Tickets are generated **after** the generation of events (according to the client workflows) and there data won't be fetched in the `data` object in `before_post` of Event model.

Besides your logic for event model doesn't make sense either as pointed out earlier. You cannot query   an event which hasn't even been created. The point of looking for ticket relationships is a good logic but not something for a `before_post` hook.

kushthedude:
> @kushthedude Have you thought this through?
> An event post won't be including any data about the `tickets`. This is why we handled it on the client sides because during generation of event, both `event` model and `ticket` models are different and Tickets are generated **after** the generation of events (according to the client workflows) and there data won't be fetched in the `data` object in `before_post` of Event model.
> 
> Besides your logic for event model doesn't make sense either as pointed out earlier. You cannot query an event which hasn't even been created. The point of looking for ticket relationships is a good logic but not something for a `before_post` hook.

@uds5501 Since you have already implemented a client-side check, Anyone who finds a more efficient server check may work further on this PR or make a new one. As currently, I have other PRs to fix and issue too. And also adding one more server check in the mean time can be delayed for a later time.

iamareebjamal:
The check has to be this way. The implementation may not be correct here, but frontend will have to change its implementation when the check is in place.

There is no way this check can be implemented without frontend changing its implementation, so let's stop using the excuse that frontend will break. It will in either case. Having a check on frontend is not enough. 

The issue was created to prevent spam. There are no JavaScript crawler spam bots. They will always use API to spam. So, frontend check is useless.

Frontend will have to first add a draft event, create tickets and then publish event as discussed many times before. This will have to be the way due to the limitations of JSONAPI spec that does not support nested creation of items

uds5501:
@iamareebjamal 
Draft event -> Create Tickets -> publish event. This can be implemented on the client side.

One thing which still remains is, I can post a `published` state event directly (by a simple POST call) . So should we restrict events being directly published altogether?

kushthedude:
> Frontend will have to first add a draft event, create tickets and then publish event as discussed many times before.

@uds5501

kushthedude:
Can you work on client side changes? @uds5501

uds5501:
@kushthedude I am interested in handling the server side in this case. 
maybe you and @shreyanshdwivedi can collaborate to take up the client side?

iamareebjamal:
> One thing which still remains is, I can post a published state event directly (by a simple POST call) . So should we restrict events being directly published altogether?

Yes, it should not be possible

PR #6215, opened 2019-07-19
fix: Fix event exports & implement route to serve exported ZIP

### LINKED ISSUE #6214: Event Export always has localhost in file path(any environment)
**Describe the bug**
When you export an event, the file path always contains localhost as a part of it regardless of the environment.

**Screenshot**
Screenshot 2019-07-19 at 10 49 12 AM

mrsaicharan1:
Fixes #6214 

Screenshot 2019-07-19 at 12 53 57 PM

EventExport

Related FE PR: https://github.com/fossasia/open-event-frontend/pull/3294

#### Checklist

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -374,6 +374,22 @@ def order_invoices(order_identifier):
         return ForbiddenError({'source': ''}, 'Authentication Required to access Invoice').respond()
 
 
+@ticket_blueprint.route('static/media/exports/<event_id>/zip/<event_hash>/<filename>')
+@jwt_required
+def event_export(event_id, event_hash, filename):
+    if not current_user:
+        return ForbiddenError({'source': ''}, 'Authentication Required to export events').respond()
+    if not current_user.is_organizer(event_id) and not current_user.is_staff:
+        return ForbiddenError({'source': ''}, 'Unauthorized Access').respond()
+    file_path = '../static/media/exports/{}/zip/{}/{}'.format(event_id, event_hash, filename)
+    try:
+        return make_response(send_file(file_path,  mimetype='application/zip',
+                             as_attachment=True,
+                             attachment_filename='event{}.zip'.format(event_hash)))
+    except FileNotFoundError:
+        raise ObjectNotFound({'source': ''}, "The event export has failed")
+
+
 @ticket_blueprint.route('/events/invoices/<string:invoice_identifier>')
 @jwt_required
 def event_invoices(invoice_identifier):
--- a/app/api/helpers/export_helpers.py
+++ b/app/api/helpers/export_helpers.py
@@ -184,7 +184,7 @@ def export_event_json(event_id, settings):
     Exports the event as a zip on the server and return its path
     """
     # make directory
-    exports_dir = app.config['BASE_DIR'] + '/static/uploads/exports/'
+    exports_dir = app.config['BASE_DIR'] + 'static/uploads/exports/'
     if not os.path.isdir(exports_dir):
         os.makedirs(exports_dir)
     dir_path = exports_dir + 'event%d' % int(event_id)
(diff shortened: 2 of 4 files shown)


app/api/helpers/export_helpers.py
```diff
@@ -221,13 +221,7 @@ def export_event_json(event_id, settings):
     shutil.make_archive(dir_path, 'zip', dir_path)
     dir_path = dir_path + ".zip"
 
-    storage_path = UPLOAD_PATHS['exports']['zip'].format(
-        event_id=event_id
-    )
-    uploaded_file = UploadedFile(dir_path, dir_path.rsplit('/', 1)[1])
-    storage_url = upload(uploaded_file, storage_path)
-
-    return storage_url
```
iamareebjamal: Why?

mrsaicharan1: @iamareebjamal the file path is turning out to be `http://localhost/events/exports/4/zip/aUdMcHJTSl/event4.zip`. `localhost` shouldn't be a part of the `dir_path` here. But anyway, I will refactor this so that this is served over a route, similar to how admin translations was implemented.

iamareebjamal: But you have removed parts like uploaded file and upload function. Why were they here and why won't removing them cause issues

mrsaicharan1: the port is not getting attached to localhost
instead of `http://localhost:5000/events/exports/4/zip/aUdMcHJTSl/event4.zip`, it's becoming 
`http://localhost/events/exports/4/zip/aUdMcHJTSl/event4.zip`

iamareebjamal: That is a long standing issue. But my point is that why will something not break if you have removed these lines of code

mrsaicharan1: @iamareebjamal the function `export_event_json` is only being used here so I believe that it won't break anywhere else?

mrsaicharan1:
@iamareebjamal Why is this check present in the create_url function? This is nullifying the port for a reason which I wasn't able to understand. Is this really required here? 
https://github.com/fossasia/open-event-server/blob/7d6399d8bec141a57676673e0791598742330c46/app/api/helpers/storage.py#L192

mrsaicharan1:
What do you suggest as the best workaround for this?

app/api/helpers/export_helpers.py
iamareebjamal: What about this line `storage_url = upload(uploaded_file, storage_path)`

iamareebjamal:
Right now, I am not thinking about the solution, rather than the implications of the line removal

`storage_url = upload(uploaded_file, storage_path)`

mrsaicharan1:
> Right now, I am not thinking about the solution, rather than the implications of the line removal
> 
> `storage_url = upload(uploaded_file, storage_path)`

Well, till that line, the zip archive is already made and stored at a specific place. The line `storage_url = upload(uploaded_file, storage_path)` only moves the zip file around.

app/api/helpers/export_helpers.py
mrsaicharan1: I've kept the original implementation intact and extracted the actual file path to be served so that the request is processed & goes to the newly implemented route.

app/api/auth.py
```diff
@@ -353,6 +353,22 @@ def order_invoices(order_identifier):
         return ForbiddenError({'source': ''}, 'Authentication Required to access Invoice').respond()
 
 
+@ticket_blueprint.route('static/media/exports/<event_id>/zip/<event_hash>/<filename>')
+@jwt_required()
```
iamareebjamal:
```suggestion
@jwt_required
```

app/api/helpers/export_helpers.py
```diff
@@ -226,10 +226,13 @@ def export_event_json(event_id, settings):
     )
     uploaded_file = UploadedFile(dir_path, dir_path.rsplit('/', 1)[1])
     storage_url = upload(uploaded_file, storage_path)
-
+    # to extract the raw path
+    if 'http://localhost/' in storage_url:
```
iamareebjamal: This is very brittle

mrsaicharan1: If I modify the upload function, would break a lot of other modules. should I pass an optional kwarg in the upload function, such `file_path_only` and pass it in the upload function so that it would check that?

iamareebjamal: The solution would be to never store localhost or domain name in the path and constructing the URL on the go when API request is made

mrsaicharan1: @iamareebjamal The problem is that the `upload` function is being used for tickets and order invoices. Just to export the event as a zip, If we don't store the domain in the function, it would break the current implementation of them.

mrsaicharan1:
Fixed with localhost issue of download links.

PR #6216, opened 2019-07-19
chore: Migrate to flask_jwt_extended

iamareebjamal:


app/api/helpers/ticketing.py
```diff
@@ -1,7 +1,5 @@
 from datetime import datetime
 
-from flask_jwt import current_identity as current_user
```
CosmicCoder96: @iamareebjamal This was an oversight right? I think this is causing a 500 which is causing #6284

CosmicCoder96: Current_user became undefined after this.

iamareebjamal: Yeah, my mistake. I thought this was covered by tests and did a Find/Replace. Not know how it slipped

PR #6242, opened 2019-07-27
fix: Change list url of Event and Ticket Discount Code. 

### LINKED ISSUE #4937: Separate Discount code into EventDiscountCode and TicketDiscountCode
Right now the Discount code schema/model/endpoints they handle two types of Discount codes. 
1. The Event Discount code created by the admin and applied on the service charges.
2. The Ticket discount code created by the event organizer and applied on the ticket prices of that event. 

This is IMO not following separation of concerns and a source of a lot of bugs. It is extremely confusing to the developers specially the new ones. It is well explained in the dev handbook but I am not sure how many developers would go through the dev-handbook before contributing to the project. For the client side devs of frontend and android, it is even worse. 

Consider the two endpoints:
1. `/v1/events/{event_identifier}/discount-codes` - List of ticket discount codes under an event.
2. `/v1/events/{event_identifier}/discount-code` - Event discount code related with that event.

The difference between the two endpoints is a bare `s`. It seems highly confusing to me atleast if not others. 
I propose to separate them out into two endpoints. If there are lot of common fields between them we can always create a base class and let the two classes extend them in order to obey DRY. 
I am also not sure about the priority of this issue as to whether we should do it after successful feature parity with v1 and release or before that but I think we should definitely fix this one time or the other. 

Thoughts: @iamareebjamal @niranjan94 @SaptakS @poush @mayank8318 @schedutron @bhaveshAn

prateekj117:
Fixes #4937 

#### Checklist

### DIFF
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -469,8 +469,8 @@ api.route(EventInvoiceRelationshipOptional, 'event_invoice_discount_code',
 
 # discount codes
 api.route(DiscountCodeListPost, 'discount_code_list_post', '/discount-codes')
-api.route(DiscountCodeList, 'discount_code_list', '/events/<int:event_id>/discount-codes',
-          '/events/<event_identifier>/discount-codes', '/users/<int:user_id>/discount-codes',
+api.route(DiscountCodeList, 'discount_code_list', '/events/<int:event_id>/ticket-discount-codes',
+          '/events/<event_identifier>/ticket-discount-codes', '/users/<int:user_id>/discount-codes',
           '/tickets/<int:ticket_id>/discount-codes')
 api.route(DiscountCodeDetail, 'discount_code_detail', '/discount-codes/<int:id>',
           '/events/<int:event_id>/discount-code', '/event-invoices/<int:event_invoice_id>/discount-code',
--- a/docs/api/api_blueprint.apib
+++ b/docs/api/api_blueprint.apib
@@ -17580,7 +17580,7 @@ Create a new Discount Code for event. (Only by Co-organizers)
             }
         }
 
-### List All Ticket Discount Codes [GET]
+### List All Ticket Discount Codes [GET /v1/events/{event_identifier}/ticket-discount-codes{?page%5bsize%5d,page%5bnumber%5d,sort,filter}]
 Get a list of Discount Codes.
 
 + Request
--- a/tests/hook_main.py
+++ b/tests/hook_main.py
@@ -2736,7 +2736,7 @@ def ticket_discount_code_post(transaction):
 @hooks.before("Discount Codes > Ticket Discount Code Collection > List All Ticket Discount Codes")
 def ticket_discount_code_get_list(transaction):
     """
-    GET /events/1/discount-codes
+    GET /events/1/ticket-discount-codes
     :param transaction:
     :return:
     """


app/api/__init__.py
```diff
@@ -469,11 +469,11 @@
 
 # discount codes
 api.route(DiscountCodeListPost, 'discount_code_list_post', '/discount-codes')
-api.route(DiscountCodeList, 'discount_code_list', '/events/<int:event_id>/discount-codes',
+api.route(DiscountCodeList, 'discount_code_list', '/events/<int:event_id>/ticket-discount-codes',
           '/events/<event_identifier>/discount-codes', '/users/<int:user_id>/discount-codes',
           '/tickets/<int:ticket_id>/discount-codes')
 api.route(DiscountCodeDetail, 'discount_code_detail', '/discount-codes/<int:id>',
-          '/events/<int:event_id>/discount-code', '/event-invoices/<int:event_invoice_id>/discount-code',
+          '/events/<int:event_id>/event-discount-codes', '/event-invoices/<int:event_invoice_id>/discount-code',
```
iamareebjamal: Why does DiscountCodeDetail have a URL of `/events/<int:event_id>/event-discount-codes`

prateekj117: @iamareebjamal I changed the appropriate Url names which were specifically for `event-discount-codes` and `ticket-discount-codes`.

iamareebjamal: But what will `/events/<int:event_id>/event-discount-codes` show. It is DiscountCodeDetail, not a list. Show me the response

prateekj117: @iamareebjamal Ok, I am confused. See [this](https://api.eventyay.com/#discount-codes-ticket-discount-code-collection-get) and [this](https://api.eventyay.com/#discount-codes-get-discount-code-detail-of-an-event-get), we are using the same Url to retrieve a list and to retrieve Detail of a discount code.

prateekj117: @iamareebjamal Yaa, it wouldn't give a list. It gives the detail of a discount code under an event, and not the discount code list which is of the type `event`. I will make appropriate changes. I got confused.

app/api/__init__.py
```diff
@@ -469,7 +469,7 @@
 
 # discount codes
 api.route(DiscountCodeListPost, 'discount_code_list_post', '/discount-codes')
-api.route(DiscountCodeList, 'discount_code_list', '/events/<int:event_id>/discount-codes',
+api.route(DiscountCodeList, 'discount_code_list', '/events/<int:event_id>/ticket-discount-codes',
```
shreyanshdwivedi: You are using `ticket-discount-codes` and `event-discount-codes` in docs and test but I don't see `event-discount-codes` here? Also, did you open an issue to accompany these changes on FE?

prateekj117: @shreyanshdwivedi I got confused. The Url to create ticket discount code is also wrong here. I will open a separate PR for that. And correct the changes here.

prateekj117: @shreyanshdwivedi Done. Please review now.

prateekj117:
@CosmicCoder96 The issue was not about a separate endpoint, but about naming the endpoint more appropriately.

CosmicCoder96:
If the name of the model  is `discount-code`, and the url is `/discount-codes`, then why does it need to be renamed, if separate endpoints are serving no purpose?
You are indeed creating two new endpoints for discount codes in this PR.

prateekj117:
@CosmicCoder96 The issue description gives an example of how a difference of basic 's' gives two separate results. I just renamed one of the endpoints to something more appropriate. Also, the Url of creating ticket discount code ([this](https://api.eventyay.com/#discount-codes-ticket-discount-code-collection-post)) is wrong. Opening a separate issue for that.

CosmicCoder96:
Yes, but at the expense of making manual GET requests, to get the same thing. FE can handle this easily (already does via properly defined relationships), and if for android it doesn't matter if it is following the JSON API spec or not, then we don't need to break away from the JSON API convention, this issue is not valid in my opinion. @iamareebjamal Do we really need this?

prateekj117:
@CosmicCoder96 Ya, agree on this.

iamareebjamal:
@CosmicCoder96 No, that's why I commented on the issue "low priority'

prateekj117:
@CosmicCoder96 @iamareebjamal Solved it because it was discussed in the weekly meeting and was told to be solved within last week. Sorry for wasting your time. Closing this.

PR #6269, opened 2019-07-30
feat: add billing info mandatory field to event model

### LINKED ISSUE #6268: Give event organizer the ability to decide if billing info is mandatory
Introduce a field in the event model for this purpose.

CosmicCoder96:
Fixes #6268 

 Adds a billing info mandatory field to the event model


app/models/event.py
```diff
@@ -235,6 +236,7 @@ def __init__(self,
                  discount_code_id=None,
                  onsite_details=None,
                  is_tax_enabled=None,
+                 is_billing_info_mandatory=None,
```
iamareebjamal: Should be False or None?

CosmicCoder96:
@iamareebjamal Fixed, please review again. Thanks!

PR #6328, opened 2019-08-07
fix: Return only accepted session and public speaker call by default

### LINKED ISSUE #5998: Fetching featured speakers, accepted sessions, public call for speaker by default
**Describe the bug**
By default, the server should only return featured speakers, accepted sessions, public speakers call when client-side make GET requests. Currently, it also returns pending sessions or private call for speaker.

**Additional context**
Discussed in: https://github.com/fossasia/open-event-android/pull/1902 (comment)

prateekj117:
Fixes #5998 

#### Checklist


app/api/sessions.py
```diff
@@ -89,22 +89,24 @@ def query(self, view_kwargs):
         query_ = self.session.query(Session)
```
iamareebjamal: Why write it 4 times when it can be written once

app/api/speakers_calls.py
```diff
@@ -86,20 +86,29 @@ def before_get_object(self, view_kwargs):
         """
         if view_kwargs.get('event_identifier'):
             try:
-                event = self.session.query(Event).filter_by(identifier=view_kwargs['event_identifier']).one()
+                event = self.session.query(Event).filter_by(identifier=view_kwargs['event_identifier'],
+                                                            privacy='public').one()
```
iamareebjamal: public and accepted

prateekj117: Speaker calls are an invitation to submit a proposal for being a speaker at the event. They are either `private` or `public`. I didn't get the context of accepted here. Did you mean making changes for speakers to only return accepted speakers?

app/api/sessions.py
```diff
@@ -86,7 +86,7 @@ def query(self, view_kwargs):
         :param view_kwargs:
         :return:
         """
-        query_ = self.session.query(Session)
+        query_ = self.session.query(Session).filter(Session.state == "accepted")
```
iamareebjamal: deleted_at=None

prateekj117: @iamareebjamal I tested it, in this particular case as well, it doesn't return rows with `deleted_at != null`

iamareebjamal: How does it do that automatically?

prateekj117: @iamareebjamal I will debug which part of the code does that, but I tested it, it was not returning deleted sessions.

prateekj117: @iamareebjamal This is filtered by flask-rest-jsonapi by default. 

https://github.com/fossasia/flask-rest-jsonapi/blob/4b10b48e443a839bbff369c769198ede34fdbff3/flask_rest_jsonapi/data_layers/alchemy.py#L94

iamareebjamal: OK :+1:

PR #6344, opened 2019-08-10
fix: check for case where client sends null for max_quantity

### LINKED ISSUE #6343: Discount codes patch: 500 internal server error
```
 File "/data/app/app/api/schema/discount_codes.py", line 153, in validate_quantity
2019-08-10T16:48:03.322594701Z     if data['tickets_number'] < data['max_quantity']:
2019-08-10T16:48:03.322599976Z TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

CosmicCoder96:
Fixes #6343 

There is a case where the client does supply the value for maximum discount codes, by specifying that it is null, this PR handles that case to prevent validation code from breaking.

### DIFF
--- a/app/api/schema/discount_codes.py
+++ b/app/api/schema/discount_codes.py
@@ -86,7 +86,7 @@ class DiscountCodeSchemaEvent(DiscountCodeSchemaPublic):
 
         DiscountCodeSchemaEvent.quantity_validation_helper(data)
 
-        if 'tickets_number' in data and 'max_quantity' in data:
+        if data.get('tickets_number') and data.get('max_quantity'):
             if data['tickets_number'] < data['max_quantity']:
                 raise UnprocessableEntity({'pointer': '/data/attributes/tickets-number'},
                                           "tickets-number should be greater than max-quantity")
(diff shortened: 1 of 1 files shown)


app/api/schema/discount_codes.py
```diff
@@ -86,7 +86,7 @@ def validate_quantity(self, data, original_data):
 
         DiscountCodeSchemaEvent.quantity_validation_helper(data)
 
-        if 'tickets_number' in data and 'max_quantity' in data:
+        if 'tickets_number' in data and 'max_quantity' in data and data['max_quantity'] is not None:
```
iamareebjamal:
```suggestion
        if 'tickets_number' in data and 'max_quantity' in data and data.get('max_quantity'):
```

app/api/schema/discount_codes.py
```diff
@@ -149,7 +149,7 @@ def validate_quantity(self, data, original_data):
 
         DiscountCodeSchemaTicket.quantity_validation_helper(data)
 
-        if 'tickets_number' in data and 'max_quantity' in data:
+        if 'tickets_number' in data and 'max_quantity' in data and data['max_quantity'] is not None:
```
iamareebjamal: Same

iamareebjamal:
```suggestion
        if data.get('tickets_number') and data.get('max_quantity'):
```

PR #6357, opened 2019-08-13
fix: remove additional parameters for is_paytm_activated(schema)

### LINKED ISSUE #6356: is_paytm_activated shouldn't be a read only attribute
**Describe the bug**
 Alipay, Omise, and PayPal have a property to enable them(e.g. `is_paypal_activated`). They were hybrid properties(condition was like `if client_key and secret_key then true`), so they had a dump_only=True. We need to remove this as we've moved away from such a practice. The reason for this is because we need the keys to be stored, whether the property is enabled/disabled

**Additional context**
Working on this.

mrsaicharan1:
Fixes #6356 
Related: https://github.com/fossasia/open-event-frontend/pull/3385
Removes dump_only property to enable write ability to the DB.

### DIFF
--- a/app/api/schema/settings.py
+++ b/app/api/schema/settings.py
@@ -58,7 +58,7 @@ class SettingSchemaPublic(Schema):
     #
     # Online Payment Flags
     #
-    is_paytm_activated = fields.Bool(dump_only=True, allow_none=False, default=False)
+    is_paytm_activated = fields.Bool(default=False)
     is_paypal_activated = fields.Bool(dump_only=True)
     is_stripe_activated = fields.Bool(dump_only=True)
     is_omise_activated = fields.Bool(dump_only=True)


mrsaicharan1:
@iamareebjamal @uds5501 @shreyanshdwivedi Please review.

app/api/schema/settings.py
```diff
@@ -58,7 +58,7 @@ class Meta:
     #
     # Online Payment Flags
     #
-    is_paytm_activated = fields.Bool(dump_only=True, allow_none=False, default=False)
+    is_paytm_activated = fields.Bool(allow_none=False, default=False)
```
shreyanshdwivedi: allow_none is also not required.

mrsaicharan1: Okay. Updating right now

mrsaicharan1:
@shreyanshdwivedi @uds5501 @kushthedude PLease have a look. Removed the allow_none param too.

PR #6373, opened 2019-08-16
chore: Upgrade dredd and increase hooks timeout

### LINKED ISSUE #6372: Increase dredd timeout
dredd has been timing out a lot lately. Increase the timeouts

Also, dredd 4.0 is ancient. It hasn't been updated since dredd tests were added in the repo. Use latest one

kushthedude:
Fixes #6372

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before-install:
 
 install:
   - npm install -g aglio
-  - npm install -g dredd@11.0.0
+  - npm install -g dredd
   - pip3 install -r requirements/tests.txt
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before-install:
 
 install:
   - npm install -g aglio
-  - npm install -g dredd@4.0.0
+  - npm install -g dredd@11.0.0
   - pip3 install -r requirements/tests.txt
 
 env:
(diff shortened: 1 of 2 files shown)


iamareebjamal:
Did you just increase the version without even verifying that it works?

kushthedude:
> Did you just increase the version without even verifying that it works?

I was checking if it would directly, ALAS I was wrong, There a lot of depreciation and some removals too, I will fix on local

iamareebjamal:
We don't want silent mode.

kushthedude:
@iamareebjamal Its ready !

dredd.yml
```diff
@@ -1,7 +1,7 @@
 dry-run: null
 hookfiles: ./tests/hook_main.py
 language: python
-sandbox: false
+# sandbox: false
```
iamareebjamal: Remove
```suggestion
```

kushthedude:
> We don't want silent mode.

Info mode has been removed, So I went with the `warning` as `debug` was extending the log far Travis limits

.travis.yml
```diff
@@ -20,7 +20,7 @@ before-install:
 
 install:
   - npm install -g aglio
-  - npm install -g dredd@4.0.0
+  - npm install -g dredd@11.0.0
```
iamareebjamal: This is still old. Use latest

```suggestion
  - npm install -g dredd
```

kushthedude: Its LTS, I used the stable one .

kushthedude: Done !

PR #6375, opened 2019-08-17
chore: Addition of npm cache and using npx

### LINKED ISSUE #6374: Add cache for npm
There is no cache for npm in Travis

Also, avoid global installation and use npx instead

kushthedude:
Fixes #6374

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,11 @@ cache: pip
 
 before-install:
   - npm -g install npm@latest
+  - cache: npm
 
 install:
-  - npm install -g aglio
-  - npm install -g dredd
+  - npx install -g aglio
+  - npx install -g dredd
   - pip3 install -r requirements/tests.txt
 
 env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -17,10 +17,11 @@ cache: pip
 
 before-install:
   - npm -g install npm@latest
+  - cache: npm
 
 install:
-  - npm install -g aglio
-  - npm install -g dredd
+  - npx install -g aglio
+  - npx install -g dredd
```
iamareebjamal: Not at all what I meant

.travis.yml
```diff
@@ -17,10 +17,11 @@ cache: pip
 
 before-install:
   - npm -g install npm@latest
+  - cache: npm
```
iamareebjamal: This won't work

.travis.yml
kushthedude: Do I need to install `npx` first ?

iamareebjamal:
Read about npx

PR #6376, opened 2019-08-17
chore(deps): Downgrade to dredd 11.0.0

iamareebjamal:


kushthedude:
@iamareebjamal What was the issue with the latest one ?

iamareebjamal:
See the issue

PR #6390, opened 2019-08-18
refactor: Changing Table Name

### LINKED ISSUE #6389: Rename association table to a proper name
There's a table called association

Can anyone guess what it is used for?

That's right. Linking ticket tags and tickets. So obvious...

kushthedude:
Fixes #6389


kushthedude:
@iamareebjamal There are some unwanted changes I think generated by Alembic, Should I remove them or are they required?

iamareebjamal:
There are no unwanted changes in alembic, the entire migration file generated is unwanted. It is creating a new table because obviously alembic cannot detect if you want to rename a table or create one

kushthedude:
> There are no unwanted changes in the alembic, the entire migration file generated is unwanted. It is creating a new table because obviously, alembic cannot detect if you want to rename a table or create one

Is it fine now?

kushthedude:
@iamareebjamal Please Review

migrations/versions/change_table_name_asso.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+ # revision identifiers, used by Alembic.
+ # This is a hand-written migration , Please check thouroughly .
+revision = '90d62fe3b5f4'
+down_revision = '90d62fe3b5e3'
+
+
+def upgrade():
+	op.rename('ticket_tagging', 'association')
```
iamareebjamal: Wrong

iamareebjamal:
Please read once before asking for reviews

migrations/versions/change_table_name_asso.py
kushthedude: Do I need to use 
```py
op.execute('`psql Query for Changing Table Name`')
```
@iamareebjamal

iamareebjamal: `op.rename('ticket_tagging', 'association')`

Read this again. Slowly...

prateekj117:
@kushthedude Please update the PR or close it, I will make a new one. Hardly a one day job.

PR #6412, opened 2019-08-23
fix: Comparison cast for paypal payment verification

### LINKED ISSUE #6411: Wrong type to compare when verifying PayPal order
**Describe the bug**
In verify paypal payment, the type of order amount from PayPal is a string, which make function always return False

anhanh11001:
Fixes #6411

### DIFF
--- a/app/api/helpers/payment.py
+++ b/app/api/helpers/payment.py
@@ -223,7 +223,7 @@ class PayPalPaymentsManager(object):
             currency_server = transaction.amount.currency
             sale_state = transaction.related_resources[0].sale.state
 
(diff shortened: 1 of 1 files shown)


iamareebjamal:
Why did you say it worked when it didn't?

anhanh11001:
I still have the same problem, can you check whether it is deployed yet or not? 
About the mistake above, I forgot to put change. 

Also, I think I should also tell you this: sorry for not testing properly as I couldn't set up my local database so I just create an object manually and test if it matches my postman request, it wasn't proper testing. I should have asked for proper guideline instead of doing this.

Issue #6416, opened 2019-08-25
save_to_db function is mistakenly removed in helpers/payment.py


anhanh11001:
**Describe the bug**
save_to_db function is mistakenly removed in helpers/payment.py

PR #6417, opened 2019-08-25
fix: Add save_to_db in payment.py

anhanh11001:
Fixes: #6416


prateekj117:
@anhanh11001 `get_settings` is imported twice in this file. Correct it here only.

liveHarshit:
@anhanh11001 Any updates?

PR #6427, opened 2019-08-29
fix: Ensure that event identifier is a string

### LINKED ISSUE #6423: Event on public page fails to open
On eventyay.com, 
https://eventyay.com/e/78230603

This event is listed in the public page, but server cannot fetch it using it's identifier as the identifier is missing alphabets and thus is interpreted as an ID.

Ideal behavior: identifiers should always have letters in them.

Is the rationale correct @iamareebjamal

pradeepgangwar:
Fixes #6423 

Event identifier in case of some event happened to be an integer and hence it failed to open.

- Ensure that identifier is always a string

### DIFF
--- a/app/models/event.py
+++ b/app/models/event.py
@@ -29,10 +29,15 @@ from app.models.ticket_holder import TicketHolder
 def get_new_event_identifier(length=8):
     identifier = str(binascii.b2a_hex(os.urandom(int(length / 2))), 'utf-8')
     count = get_count(Event.query.filter_by(identifier=identifier))
-    if count == 0:
-        return identifier
-    else:
+    try:
+        int(identifier)
+    except ValueError:
         return get_new_event_identifier(length)
+    else:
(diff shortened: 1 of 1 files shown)


app/models/event.py
```diff
@@ -29,10 +29,15 @@
 def get_new_event_identifier(length=8):
     identifier = str(binascii.b2a_hex(os.urandom(int(length / 2))), 'utf-8')
     count = get_count(Event.query.filter_by(identifier=identifier))
-    if count == 0:
-        return identifier
-    else:
+    try:
+        int(identifier)
+    except ValueError:
```
iamareebjamal: Wrong way to check for integer and opposite logic and in wrong place.

This will only return integer identifiers

Please test before sending PRs

Write a test for it to ensure it works correctly

pradeepgangwar:
Will change and send a fresh PR

PR #6432, opened 2019-08-30
fix: Fix max tickets allowance bug 

### LINKED ISSUE #6433: Server returns 409 : Ticket already sold out, even though tickets are remaining
image
{"errors": [{"status": 409, "source": {"pointer": "/data/attributes/ticket_id"}, "title": "Conflict", "detail": "Ticket already sold out"}], "jsonapi": {"version": "1.0"}}

Either the statistic is incorrect on frontend or it's wrong on the server.

mrsaicharan1:
Fixes #6433

Random entries of 'John Doe' pop up in the DB for every order we place! This floods the TicketHolder table with a linked ticket ID, causing the tickets to be sold out(exceeds the quantity). 
When we are retrieving the count to check for number of tickets sold, we shouldn't be including the entries with no user linked.

- Filtered out those entries with no linked user object
(The dummy John Does don't have a related user)


mrsaicharan1:
@CosmicCoder96 @shreyanshdwivedi @uds5501 @iamareebjamal Please have a look as this solves a high priority bug.

app/api/attendees.py
```diff
@@ -56,8 +56,9 @@ def before_post(self, args, kwargs, data):
                 "Ticket belongs to a different Event"
             )
         # Check if the ticket is already sold out or not.
-        if get_count(db.session.query(TicketHolder.id).
-                     filter_by(ticket_id=int(data['ticket']), deleted_at=None)) >= ticket.quantity:
+        if get_count(db.session.query(TicketHolder).
```
CosmicCoder96: you are referring to `TicketHolder.user` if I am buying a ticket for 3 others which ae not in the system then ticketHolde.user will be none for them.

mrsaicharan1: Currently, how do we go about it? When we want to book tickets for other users, do they need to be registered?

mrsaicharan1: Okay so they don't have to be registered, checked locally. So the point is valid.

mrsaicharan1:
The ticket is not generated for the Dummy entries and hence ticket pdf url is not generated. I guess we could be using this as the differentiating factor.

kushthedude:
@mrsaicharan1 Your hotfix works but the main thing is why this `John Doe` random entry is getting created that without any `Order Label`, AFAIK this is some new bug a month back no such issue or any random entry was getting created. Do you by any chance know what may be the cause for it ?

mrsaicharan1:
When I pushed this fix, couldn't find out any module/piece of code which could be the possible culprit for this. @shreyanshdwivedi @uds5501 Maybe you guys could have a look once.

mrsaicharan1:
@uds5501 Please have a look at this.

kushthedude:
@uds5501 @mrsaicharan1 @iamareebjamal This is not the server issue, While I was working on frontend, I got the cause of this. Just look at this
https://github.com/fossasia/open-event-frontend/blob/9ae3414e7a911ffe007e95dbf7fae59716fb1fe7/app/controllers/events/view/tickets/add-order.js#L62

kushthedude:
> @uds5501 @mrsaicharan1 @iamareebjamal This is not the server issue, While I was working on frontend, I got the cause of this. Just look at this
> https://github.com/fossasia/open-event-frontend/blob/9ae3414e7a911ffe007e95dbf7fae59716fb1fe7/app/controllers/events/view/tickets/add-order.js#L62

@iamareebjamal Can you please transfer the issue to FE, need to check where else this `createRecord` instance is repeated and dummy orders are created.

iamareebjamal:
PRs can't be transferred. As for the issue, let's keep it here because we need to think of a better system to avoid this from happening at the server-side as well, so that even if the client abuses the server like this, it is not rejected. 

Or else, I can just send the request to add ticket holders through Postman and everyone will have to wait for the orders to expire before acting on

kushthedude:
> PRs can't be transferred. As for the issue, let's keep it here because we need to think of a better system to avoid this from happening at the server-side as well, so that even if the client abuses the server like this, it is not rejected.
> 
> Or else, I can just send the request to add ticket holders through Postman and everyone will have to wait for the orders to expire before acting on
> 
> So, create a child PR in frontend

I acutally wanted to say to transfer the issue.
I will open a relevant isaue on FE and send PR for same

kushthedude:
@iamareebjamal Is it still needed ?

kushthedude:
@iamareebjamal Is it still needed?

PR #6435, opened 2019-09-02
feat: add custom form json column to sessions, speaker, attendees

### LINKED ISSUE #6191: Propose and Implement a schema for complex custom forms
Current custom forms are really restricted, and rigid. Only text fields are available, along with hardcoded dropdowns. Further, the user is only able to select from among the hardcoded fields and toggle them on/off for his/her event. 

We already have the framework to associate simple custom fields with individual events or orders, and an API to create them. The custom forms schema needs to be now expanded to allow more complex fields. Taking Google forms and our use case as an inspiration, the user should be able to create the following fields:

- Simple text *
- Paragraph *
- Radio Buttons (single choice)
- Checkboxes
- Dropdown
- File upload *
- Time
- Date
- Date & Time

\* Already implemented

The schema needs expansion to accommodate options for fields like dropdowns, checkboxes and radio buttons. 
Also, to store custom labels to the fields, which the user assigns. Currently, they are hardcoded by comparing the name of the field with if-else.

-- First, plan the schema refactor completely, before beginning to code it. Get @niranjan94 and @iamareebjamal to review it, and once it's finalized, proceed with it. 

-- Design the schema while writing appropriate migrations to take care of legacy data on the server.

-- use this parent issue as for all discussions, and queries.

### LINKED ISSUE #6434: Add JSON column to attendees, sessions and speakers
Parent issue: #6191

CosmicCoder96:
Fixes #6434 
Fixes #6191

Adds fields to accommodate complex custom form values in sessions, speakers and attendees.


iamareebjamal:
This will allow users to add arbritrarily nested JSON to the DB. Please add validation for only flat JSON

CosmicCoder96:
@iamareebjamal Added a validation check to only allow flattened JSON. Verified locally by passing a two-level deep  JSON object.

app/api/schema/speakers.py
```diff
@@ -12,6 +13,13 @@ class SpeakerSchema(SoftDeletionSchema):
     Speaker Schema based on Speaker Model
     """
 
+    @validates_schema(pass_original=True)
+    def validate_json(self, data, original_data):
+        if data.get('complex_field_values'):
+            if any(isinstance(i, dict) for i in data['complex_field_values'].values()):
```
iamareebjamal:
```json
{
  "a": [{... big nested data ...}, {... big nested data ...}, {... big nested data ...}] 
}
```

This will be supported

iamareebjamal: Also, let's add a limit to the fields in settings. By default, allow only 30 items. I think it is fair. If more are required, then it can be changed by the admin

CosmicCoder96: > ```
> {
>   "a": [{... big nested data ...}, {... big nested data ...}, {... big nested data ...}] 
> }
> ```
> 
> This will be supported

@iamareebjamal for this can we impose that the value **has** to be a string or number or null, that will eliminate the possibility of a list or other unsupported types.

iamareebjamal: Agree

CosmicCoder96:
@iamareebjamal Please take a look now.

PR #6438, opened 2019-09-10
[WIP] fix: Remove secret and app_environment from settings

### LINKED ISSUE #6093: Remove app_environment and secret from settings
1. app_environment is a server wide setting. Changing it while running server can lead to unpredictable results. For example, Changing from production to development won't cause the server to start showing debug errors since it requires server restart. Also, the APP_CONFIG environment variable controls the flask server mode and database app_environment controls the programmatic config settings, these 2 conflict and overlap without any single one completely overriding the other. Only APP_CONFIG should control the system wide setting of app mode, since you should definitely restart the server in order to change from development to production and vice versa. Hence, it should be removed from the db
2. Keeping the secret in DB is like leaving the key to the vault taped to its front with a label on it saying - "Hey, this is the key to the vault where all my valuables are stored. Have fun stealing them". secret is used in (potentially) cryptographic operations such as password hashing and encryption. We MAY not be using it now, but if in future, we encrypt the data we store and the database is compromised, the hackers will get the key to decrypt the data with it. I don't why it was thought as a good idea to store it in the DB. There is no use to be able to change secret through the frontend since it should never be changed once a server is deployed and ready to use. In fact, transfer from server to frontend and vice versa over insecure connection will leak it as well. It should only be configured through environment variables. Hence, remove from settings and make it a compulsory environment variable. Server should not start if the secret is not configured

prateekj117:
Fixes #6093 

#### Checklist


kushthedude:
@prateekj117 Stale?

prateekj117:
@kushthedude Will update it in a few days, exams going on right now.

prateekj117:
Will be solved after solving some other Urgent issues.

Issue #6465, opened 2019-09-24
Refactoring model definitions to data classes


mrsaicharan1:
**Feature Description**
A ton of boilerplate code is written everywhere because of the redundant `__init__`, `__repr__` and other methods. Moving to the dataclass architecture would prove to get rid of this and focus mainly on data.

**Describe the solution you'd like**
Implementation of Data Classes

Soniyanayak51:
Hi, I am having trouble with testing. Can you tell me the the process of running unit tests? 
`python3 -m unittest discover tests/all/` raises many module not found errors. Where can I find all the dependencies that need to be installed?

iamareebjamal:
`nosetests tests/all`

Soniyanayak51:
@iamareebjamal I am still getting many errors like

ERROR: Failure: InvalidRequestError (One or more mappers failed to initialize - can't proceed with initialization of other mappers. Triggering mapper: 'mapped class EventTopic->event_topics'. Original exception was: When initializing mapper mapped class EventTopic->event_topics, expression 'EventSubTopic' failed to locate a name ("name 'EventSubTopic' is not defined"). If this is a class name, consider adding this relationship() to the <class 'app.models.event_topic.EventTopic'> class after both dependent classes have been defined.)
.
.
.
sqlalchemy.exc.InvalidRequestError: One or more mappers failed to initialize - can't proceed with initialization of other mappers. Triggering mapper: 'mapped class EventTopic->event_topics'. Original exception was: When initializing mapper mapped class EventTopic->event_topics, expression 'EventSubTopic' failed to locate a name ("name 'EventSubTopic' is not defined"). If this is a class name, consider adding this relationship() to the <class 'app.models.event_topic.EventTopic'> class after both dependent classes have been defined.

iamareebjamal:
Are you running it on development branch or after some changes you made?

Soniyanayak51:
I created a different branch. Running the command after I made changes.

iamareebjamal:
Then there's a problem in the changes you made. Debug what's the issue

Soniyanayak51:
@iamareebjamal but I did not make changes to the files mentioned in the error and I created the branch after updating the development branch. The last part of the error is:
`======================================================================
ERROR: Test generation of hash for a key.
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #6466, opened 2019-09-25
refactor: Convert TicketHolder Model Def to dataclass

mrsaicharan1:
Part of #6465 

Added Initial Data Class Definition for TicketHolder. Will Proceed with the other models if this is OK to go.

- Moved to Data Classes scheme for `TicketHolder`
- Upgraded Python version to 3.7 in Travis Config file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: python
 
+dist: xenial
+
 services:
   - docker
+  - postgresql
 
 addons:
   apt:
@@ -9,11 +12,8 @@ addons:
       - python3-dev
       - libssl-dev
       - libffi-dev
-matrix:
-    include:
-        - python: 3.7
-          dist: xenial
-          sudo: true
+
+python: 3.7
 
 cache: pip
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,11 @@ addons:
       - python3-dev
       - libssl-dev
       - libffi-dev
-
-python:
-  - '3.6'
+matrix:
+    include:
+        - python: 3.7
+          dist: xenial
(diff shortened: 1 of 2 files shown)


iamareebjamal:
https://stackoverflow.com/questions/55577165/sqlalchemy-with-dataclass-default-not-populating-postgres-database

mrsaicharan1:
@iamareebjamal I'm just a bit unsure about the type hinting for the relationships defined in the class. I'll test it out by purchasing tickets on FE and see if it breaks mid-way.

iamareebjamal:
Type hinting can never break anything because it is ignored at runtime. If something gets broken, it might be due to dataclass

iamareebjamal:
You need to update the python version to 3.7 in travis CI

mrsaicharan1:
Tested this for a normal ticket purchase. Works fine.

mrsaicharan1:
Python 3.7 couldn't be configured by changing just the version number. Had to specify the xenial source explicitly.
https://travis-ci.community/t/unable-to-download-python-3-7-archive-on-travis-ci/639/10

.travis.yml
```diff
@@ -9,9 +9,11 @@ addons:
...
       - libssl-dev
       - libffi-dev
-
-python:
-  - '3.6'
+matrix:
+    include:
+        - python: 3.7
+          dist: xenial
+          sudo: true
```
iamareebjamal: No need to use matrix. Just write dist: xenial

iamareebjamal:
You have to add postgresql as service as well

https://docs.travis-ci.com/user/database-setup/#postgresql

PR #6475, opened 2019-09-30
fix: Revert dredd to 4.0.0

iamareebjamal:


kushthedude:
@iamareebjamal Revert to at least 10.0, as 4.0 would be too depreciated

iamareebjamal:
This is the last well-tested version. And it is working fine. This is not a permanent solution. We will use the latest dredd once we know what's wrong. Meanwhile, we can't let tests fail

PR #6510, opened 2019-10-06
fix: Fix and test cron job for Invoice Due Marking

### LINKED ISSUE #6494: Cron job for marking event invoices due is wrong
PR https://github.com/fossasia/open-event-server/pull/6166 is wrong. The query fails in every case

https://github.com/fossasia/open-event-server/blob/bc45c1d267b924b556d36855f29c1dd96e901209/app/api/helpers/scheduled_jobs.py#L164-L174

This line is not valid SQLAlchemy syntax:

https://github.com/fossasia/open-event-server/blob/bc45c1d267b924b556d36855f29c1dd96e901209/app/api/helpers/scheduled_jobs.py#L169

`Neither 'InstrumentedAttribute' object nor 'Comparator' object associated with EventInvoice.event has an attribute 'ends_at'`

It should be `db.session.query(EventInvoice).join(Event).filter(EventInvoice.status == 'upcoming', Event.ends_at >= datetime.datetime.now()).all()`

Each run of this task is failing

Fix it and write tests to ensure it is working

mrsaicharan1:
Fixes #6494 

Fixed event invoice marking error and added integration test for the same.

- Used Multi-Value table update logic in accordance with the ORM norms
- Added integration test for the logic

### DIFF
--- a/app/api/helpers/scheduled_jobs.py
+++ b/app/api/helpers/scheduled_jobs.py
@@ -164,14 +164,12 @@ def expire_pending_tickets():
 def event_invoices_mark_due():
     from app import current_app as app
     with app.app_context():
-        db.session.query(EventInvoice).\
-                    filter(EventInvoice.status == 'upcoming',
-                           EventInvoice.event.ends_at >= datetime.datetime.now(),
-                           (EventInvoice.created_at + datetime.timedelta(days=30) <=
-                            datetime.datetime.now())).\
-                    update({'status': 'due'})
-
-        db.session.commit()
+        db.session.query(EventInvoice).filter(
+            EventInvoice.status == 'upcoming',
+            Event.id == EventInvoice.event_id,
+            Event.ends_at >= datetime.datetime.now(),
+            (EventInvoice.created_at + datetime.timedelta(days=30) <= datetime.datetime.now())
+        ).update({EventInvoice.status: 'due'}, synchronize_session=False)
 
 
 def send_monthly_event_invoice():
--- a/tests/all/integration/api/helpers/test_scheduled_jobs.py
+++ b/tests/all/integration/api/helpers/test_scheduled_jobs.py
@@ -16,16 +16,17 @@ class TestScheduledJobs(OpenEventTestCase):
         with app.test_request_context():
             event_invoice_new = EventInvoiceFactory(event__ends_at=datetime.datetime(2019, 7, 20))
             event_invoice_paid = EventInvoiceFactory(status="paid")
-            event_invoice_future = EventInvoiceFactory(created_at=datetime.datetime(2099, 12, 14))
-            
+            event_invoice_future = EventInvoiceFactory()
+            event_invoice_future.created_at = datetime.datetime(2099, 12, 14)
+
             db.session.commit()
-            
+
             event_invoice_new_id = event_invoice_new.id
             event_invoice_paid_id = event_invoice_paid.id
             event_invoice_future_id = event_invoice_future.id
-            
+
             event_invoices_mark_due()
-            
+
             status_new = EventInvoice.query.get(event_invoice_new_id).status
             status_paid = EventInvoice.query.get(event_invoice_paid_id).status
             status_future = EventInvoice.query.get(event_invoice_future_id).status
--- a/tests/all/integration/api/helpers/test_scheduled_jobs.py
+++ b/tests/all/integration/api/helpers/test_scheduled_jobs.py
@@ -16,21 +16,16 @@ class TestScheduledJobs(OpenEventTestCase):
         with app.test_request_context():
             event_invoice_new = EventInvoiceFactory(event__ends_at=datetime.datetime(2019, 7, 20))
             event_invoice_paid = EventInvoiceFactory(status="paid")
-            event_invoice_future = EventInvoiceFactory()
-            event_invoice_future.created_at = datetime.datetime(2099, 12, 14)
 
             db.session.commit()
 
             event_invoice_new_id = event_invoice_new.id
             event_invoice_paid_id = event_invoice_paid.id
-            event_invoice_future_id = event_invoice_future.id
 
             event_invoices_mark_due()
 
             status_new = EventInvoice.query.get(event_invoice_new_id).status
             status_paid = EventInvoice.query.get(event_invoice_paid_id).status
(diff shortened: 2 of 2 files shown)


tests/all/integration/api/helpers/test_scheduled_jobs.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_event_invoices_mark_due(self):
+        """Method to test marking of event invoices as due"""
+
+        with app.test_request_context():
+            event_invoice_obj = EventInvoiceFactory()
+            event_obj = EventFactoryBasic()
+            event_obj.ends_at = datetime.date(2019, 7, 20)
+            event_invoice_obj.event = event_obj
+            save_to_db(event_obj)
+            save_to_db(event_invoice_obj)
```
iamareebjamal: Can be changed to a single line

```python
event_invoice_obj = EventInvoiceFactory(event__ends_at=datetime.date(2019, 7, 20))
```

mrsaicharan1:
Will update it asap.

tests/all/integration/api/helpers/test_scheduled_jobs.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+from datetime import datetime
```
iamareebjamal:
```suggestion
import datetime
```

iamareebjamal:
https://stackoverflow.com/questions/42543223/how-do-you-express-a-multi-table-update-update-from-in-sqlalchemy-orm

iamareebjamal:
I don't think you read the answer in SO question, you can't use update after using join.

Also, you can run the test locally before pushing, `nosetests tests/all/integration/api/helpers/test_scheduled_jobs.py`

app/api/helpers/scheduled_jobs.py
```diff
@@ -164,12 +165,12 @@ def expire_pending_tickets():
...
     with app.app_context():
-        db.session.query(EventInvoice).\
+        db.session.query(EventInvoice).join(Event).\
                     filter(EventInvoice.status == 'upcoming',
-                           EventInvoice.event.ends_at >= datetime.datetime.now(),
+                           Event.ends_at >= datetime.datetime.now(),
                            (EventInvoice.created_at + datetime.timedelta(days=30) <=
                             datetime.datetime.now())).\
-                    update({'status': 'due'})
+                    update({'status': 'due'}, synchronize_session=False)
```
iamareebjamal: `synchronize_session=False` has no use here

mrsaicharan1:
@iamareebjamal I will test it out locally once before pushing.

app/api/helpers/scheduled_jobs.py
mrsaicharan1: Was just checking something else. Yeah.

app/api/helpers/scheduled_jobs.py
```diff
@@ -164,14 +165,15 @@ def expire_pending_tickets():
...
     with app.app_context():
-        db.session.query(EventInvoice).\
-                    filter(EventInvoice.status == 'upcoming',
-                           EventInvoice.event.ends_at >= datetime.datetime.now(),
-                           (EventInvoice.created_at + datetime.timedelta(days=30) <=
-                            datetime.datetime.now())).\
-                    update({'status': 'due'})
-
-        db.session.commit()
+        event_invoices = db.session.query(EventInvoice).join(Event).\
```
iamareebjamal: I don't know how else to say this, you cannot use join in the update query...

app/api/helpers/scheduled_jobs.py
```diff
@@ -164,14 +165,15 @@ def expire_pending_tickets():
...
+                            filter(EventInvoice.status == 'upcoming',
+                            Event.ends_at >= datetime.datetime.now(),
+                            (EventInvoice.created_at + datetime.timedelta(days=30) <=
+                            datetime.datetime.now())).all()
+        for invoice in event_invoices:
+            obj = EventInvoice.query.filter_by(id=invoice.id).one()
+            obj.status = 'due'
+            save_to_db(obj)
```
iamareebjamal: This is grossly inefficient. Please see the stackoverflow answer

iamareebjamal:
Now that it's tested for an event that should be marked due. Check by creating another invoice that it should not be marked due to ensure that function is working correctly

mrsaicharan1:
@iamareebjamal Okay, I think I swayed away from the StackOverflow Answer. So basically, without a join, the related attributes can still be accessed for an update and the strategy for updation was changed with respect to synchronisation. Alright

mrsaicharan1:
I'll be adding for another test case where it shouldn't be marked as due.

tests/all/integration/api/helpers/test_scheduled_jobs.py
```diff
@@ -0,0 +1,29 @@
...
+            event_invoices_mark_due()
+            status = EventInvoice.query.get(event_invoice_id).status
+            self.assertEqual(status, "due")
+
+            event_invoice = EventInvoiceFactory(status="paid")
+            db.session.commit()
+            event_invoice_id = event_invoice.id
+            event_invoices_mark_due()
+            status = EventInvoice.query.get(event_invoice_id).status
+            self.assertNotEqual(status, "due")
```
mrsaicharan1: Added the test case. @iamareebjamal

iamareebjamal: There's one case missing for upcoming events with `(EventInvoice.created_at + datetime.timedelta(days=30) > datetime.datetime.now()`

Also, commit each one and run `event_invoices_mark_due()` once and verify the 3 invoice status

mrsaicharan1: This test seems to fail. Created a future datetime object and passed it into the `created_at` attribute.

iamareebjamal:
Seems like created_at check won't work this way. @mrsaicharan1 Can you tell me what does marking due means in terms of invoices?

mrsaicharan1:
> Seems like created_at check won't work this way. @mrsaicharan1 Can you tell me what does marking due means in terms of invoices?

So the basic workflow here is that whenever an invoice is created, after 30 days from its creation, it would be marked as due [upcoming->due]

PR #6523, opened 2019-10-15
fix: ICal Attribute error & add integration test

### LINKED ISSUE #6511: ICalExporter not working
https://github.com/fossasia/open-event-server/blob/97e94f2ce14d20341253822b5456c93bef590c71/app/api/helpers/ICalExporter.py#L48-L53

There is no email attribute on event object.

Fix the method and write tests

mrsaicharan1:
Fixes #6511 

-
-
-

#### Checklist

### DIFF
--- a/app/api/helpers/ICalExporter.py
+++ b/app/api/helpers/ICalExporter.py
@@ -47,8 +47,7 @@ class ICalExporter:
 
             for speaker in session.speakers:
                 # Ref: http://icalendar.readthedocs.io/en/latest/usage.html#file-structure
-                # can use speaker.email below but privacy reasons
-                attendee = vCalAddress('MAILTO:' + event.email if event.email else 'undefined@email.com')
+                attendee = vCalAddress('MAILTO:' + speaker.email if speaker.email else 'undefined@email.com')
                 attendee.params['cn'] = vText(speaker.name)
                 event_component.add('attendee', attendee)
 
--- a/app/models/speaker.py
+++ b/app/models/speaker.py
@@ -90,7 +90,7 @@ class Speaker(SoftDeletionModel):
         self.sponsorship_required = sponsorship_required
         self.event_id = event_id
         self.user_id = user_id
-        self.deleted_at = deleted_at,
+        self.deleted_at = deleted_at
         self.complex_field_values = complex_field_values
 
     @staticmethod
--- a/tests/all/integration/api/helpers/test_icalexporter.py
+++ b/tests/all/integration/api/helpers/test_icalexporter.py
@@ -2,17 +2,10 @@ import unittest
 import icalendar
 from app import current_app as app
 
-from tests.all.integration.setup_database import Setup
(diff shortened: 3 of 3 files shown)


app/api/helpers/ICalExporter.py
```diff
@@ -48,7 +48,7 @@ def export(event_id):
             for speaker in session.speakers:
                 # Ref: http://icalendar.readthedocs.io/en/latest/usage.html#file-structure
                 # can use speaker.email below but privacy reasons
-                attendee = vCalAddress('MAILTO:' + event.email if event.email else 'undefined@email.com')
```
iamareebjamal: That's not what we want to do. People will get emails of speakers by exporting large events and use it to create spam list

That's the privacy issue with it.

Ans also, there needs to a test

mrsaicharan1: Just to be clear, organizers are the only individuals who'd be able to export the events in ical formats right? Organizers would need the list of emails, wouldn't they?
Also, what email should be considered here? If we were to add an email attribute to the event, what should it be?

iamareebjamal: If only organizers are able to export events, then it's not an issue.

If not, then owner email should be set

mrsaicharan1: Currently users who organize events are the only individuals who can export them in `.ical` format. Therefore, there doesn't seem to be an issue here. Will test it and remove `DO-NOT-MERGE`

mrsaicharan1:
@iamareebjamal I've added tests to check the presence of subcomponents in an `str` format.

iamareebjamal:
It doesn't test for speaker, which is the failing case/

mrsaicharan1:
@iamareebjamal I couldn't link the factories to test it. It always shows up as None. 
According to factoryboy docs,
Whenever I link the speaker with the session factory in the usual fashion, the `str` version of the exported ical shows absolutely no info pertaining to the speakers.
```
test_session = SessionFactory()
test_speaker = SpeakerFactory()
test_speaker.session = test_session
save_to_db(test_session)
save_to_db(test_speaker)
test_cal_str = icalexport_object.export(test_session.event_id)
```


Here is the string repr of the ical export
```
['BEGIN:VCALENDAR', 'VERSION:2.0', 'PRODID:-//fossasia//open-event//EN', 'X-WR-CALDESC:Schedule for sessions at example', 'X-WR-CALNAME:example', 'BEGIN:VEVENT', 'SUMMARY:example', 'DTSTART;VALUE=DATE-TIME:20991213T000000', 'DTEND;VALUE=DATE-TIME:20991214T000000', 'UID:1-4a766c6a', 'DESCRIPTION:example', 'GEO:1.23456789;1.23456789', 'LOCATION:example', 'URL:http://localhost/v1/events?identifier=4a766c6a', 'END:VEVENT', 'END:VCALENDAR', '']
```

iamareebjamal:
Will test

PR #6544, opened 2019-10-21
Seperate endpoint for celery

### LINKED ISSUE #6471: Implement docker healthcheck and fix endpoint
Currently, the health check endpoint shows that celery is not up. Which is untrue
https://api.eventyay.com/health-check

Besides, the presence or absence of celery does not signify the health of the web container. Hence, only DB and migrations should be checked by default and celery should have another endpoint for itself for checking the celery container

The second part of the issue is to add healthcheck command to Dockerfile so that 
1. DockerHub can automatically test to see if it runs or not (will most likely fail as it depends on PSQL)
2. Docker Compose can restart brain dead containers - live containers no longer fit for serving requests

YashJipkate:
Fixes #6471

Separate endpoint for celery check.

- Split the combined health check for db and celery into two
- The endpoint for db and migrations check remains the same as `/health-check`
- The endpoint for celery check is now `/celery-check`

#### Checklist

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -223,13 +223,16 @@ def make_celery(app=None):
     return ext.celery
 
 
-# Health-check
+# Health-check for db and migrations
 health = HealthCheck(current_app, "/health-check")
-health.add_check(health_check_celery)
 health.add_check(health_check_db)
 with current_app.app_context():
     current_app.config['MIGRATION_STATUS'] = check_migrations()
 health.add_check(health_check_migrations)
+
+#Health-check for celery
+health = HealthCheck(current_app, "/celery-check")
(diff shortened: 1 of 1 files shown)


YashJipkate:
Hi @iamareebjamal The tests seem to be failing because we can't have the same app for two HealthCheck endpoints at the same time. Should I create a new app for the same?

iamareebjamal:
No, new endpoint is not needed

YashJipkate:
@iamareebjamal So the endpoint for celery checks should be removed altogether? If yes, then where should we have the celery check?

iamareebjamal:
No, it should be fixed. Not removed

PR #6545, opened 2019-10-21
fix: Add ticketing check for publishing an event in PATCH request

### LINKED ISSUE #6210: Event Wizard: Require Tickets before Event Publishing

prateekj117:
Partially Fixes #6210 

Implemented a server-side check for PATCH request to see if the event has some associated tickets or not before publishing an event.

#### Checklist


prateekj117:
@iamareebjamal Please review. This is just a check for `PATCH` request. As already noticed, we cannot implement a similar check for `POST` requests as the current workflow does not allow us. I will implement that part later.

app/api/events.py
```diff
@@ -577,11 +578,21 @@ def before_update_object(self, event, data, view_kwargs):
             else:
                 event.deleted_at = data.get('deleted_at')
 
+        deleted_tickets = 0
+        for ticket in event.tickets:
+            if ticket.deleted_at:
+                deleted_tickets += 1
```
iamareebjamal: ?

app/api/events.py
```diff
@@ -577,11 +578,21 @@ def before_update_object(self, event, data, view_kwargs):
...
         if 'is_event_online' not in data and event.is_event_online \
                 or 'is_event_online' in data and not data['is_event_online']:
-            if data.get('state', None) == 'published' and not data.get('location_name', None):
-                raise ConflictException({'pointer': '/data/attributes/location-name'},
-                                        "Location is required to publish the event")
+            if data.get('state', None) == 'published':
+                if not data.get('location_name', None):
+                    raise ConflictException({'pointer': '/data/attributes/location-name'},
+                                            "Location is required to publish the event")
+                if deleted_tickets == len(event.tickets):
```
iamareebjamal: How does this check if there are no tickets in the event?

Makes no sense to me

app/api/events.py
prateekj117: If the event has more than 1 ticket associated with it, we want to make sure that every one of them is deleted. The event has tickets object even when all of them are soft deleted.

app/api/events.py
prateekj117: Because when there are no tickets associated with an event, the event has an array of tickets with size 0, while the `deleted_tickets` are itself zero as well. So even when event has no tickets, `event.tickets` does not throw an error, and `len(event.tickets)` returns 0.

iamareebjamal: `event.ticket.query(deleted_at=None).count() > 1`

This is not actual code. But this is how it should be done

If there are 2 active tickets and a million deleted tickets, you'll have 1 million python objects in memory. Whereas there need to be 0

prateekj117: @iamareebjamal Will make the required changes.

prateekj117:
@iamareebjamal Done. Please review.

app/api/events.py
```diff
@@ -578,18 +578,15 @@ def before_update_object(self, event, data, view_kwargs):
             else:
                 event.deleted_at = data.get('deleted_at')
 
-        deleted_tickets = 0
-        for ticket in event.tickets:
-            if ticket.deleted_at:
-                deleted_tickets += 1
+        deleted_tickets = event.tickets.filter(Ticket.deleted_at != None).count()
```
iamareebjamal: As I said, this is not actual code. It won't work

prateekj117: @iamareebjamal This is. When we put `lazy=dynamic`, it allows us to query into relationships using this syntax. 
Reference: https://stackoverflow.com/questions/15218732/filter-on-sqlalchemy-relationship

iamareebjamal: So, is lazy_dynamic set?

prateekj117: @iamareebjamal Check all file changes.

iamareebjamal:
Ok, now create compatible PR on frontend

prateekj117:
@iamareebjamal Patch request won't break anything on the frontend. We would need a separate implementation for POST request. And that would require change of workflow of event creation as well.

iamareebjamal:
Ok

prateekj117:
@kushthedude And why do you think that the user will neither be able to move forward from the Basic Details Page nor able to save draft because of this PR.

kushthedude:
I have tested locally, please test yourself.

prateekj117:
@kushthedude Paste your logs.

prateekj117:
@kushthedude Please paste your logs so that I can correct if I missed something. This is a major fix.

PR #6572, opened 2019-11-04
fix: Check if identifier does not contain of all digits.

### LINKED ISSUE #6423: Event on public page fails to open

prateekj117:
Fixes #6423 

The public event page does not open in case the identifier of the event contains all digits. 

#### Checklist


prateekj117:
@iamareebjamal As you told on Gitter that you have appended `ref` on the identifiers of the events which contain all digits, so we don't need to handle the old cases now, right?

iamareebjamal:
Not everyone who has open event server will have done it, so we need to handle it

Make a script which goes through events which have digit identifier and updates their identifier

So, that we can add in release note to just run `python manage.py fix digit_identifier`

prateekj117:
@iamareebjamal Done. Please review.

manage.py
```diff
@@ -37,6 +37,15 @@ def add_event_identifier():
         save_to_db(event)
 
 
+@manager.command
+def fix_digit_identifier():
+    events = Event.query.all()
```
iamareebjamal: I want to complete this script before my daughter's wedding, so let's not iterate on 15411442... events when there are only 36 events with digit identifier

manage.py
```diff
@@ -37,6 +37,15 @@ def add_event_identifier():
...
+    for event in events:
+        if event.identifier.isdigit():
+            event.identifier = event.identifier + 'ref'
```
iamareebjamal: I have already done this as a workaround, then what is the need of this modification? This was a workaround, not a fix

manage.py
```diff
@@ -37,6 +37,15 @@ def add_event_identifier():
...
+            save_to_db(event)
```
iamareebjamal: No need to hit DB on each iteration

app/models/event.py
```diff
@@ -29,7 +29,7 @@
 def get_new_event_identifier(length=8):
     identifier = str(binascii.b2a_hex(os.urandom(int(length / 2))), 'utf-8')
     count = get_count(Event.query.filter_by(identifier=identifier))
-    if count == 0:
+    if count == 0 and not identifier.isdigit():
```
iamareebjamal: Why event check for count if it is a digit? Why make the DB call?

manage.py
prateekj117: @iamareebjamal I am not sure how I would write a query like that in SqlAlchemy. I need a little help there.

iamareebjamal: `select * from events where identifier ~ '^[0-9\.]+$';`

Convert this to SQLAlchemy

prateekj117:
@iamareebjamal Please review.

app/models/event.py
```diff
@@ -28,6 +28,8 @@
 
 def get_new_event_identifier(length=8):
     identifier = str(binascii.b2a_hex(os.urandom(int(length / 2))), 'utf-8')
+    if identifier.isdigit():
+        return get_new_event_identifier(length)
     count = get_count(Event.query.filter_by(identifier=identifier))
     if count == 0:
```
iamareebjamal:
```suggestion
    if not identifier.isdigit() or get_count(Event.query.filter_by(identifier=identifier)) == 0:
```

prateekj117:
@iamareebjamal Done.

iamareebjamal:
Add the command in CHANGELOG.md to be run for correcting the identifiers

prateekj117:
@iamareebjamal Done.

PR #6538, opened 2019-10-19
fix: #6522 Added organizer email in events table attributes

ranjsa:
Fixes #6522 

#### Short description of what this resolves:
- https://github.com/fossasia/open-event-attendee-android/issues/1964
- Added organizer email to contact event organizer directely through email .

#### Checklist

- [ ] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- [ ] My branch is up-to-date with the Upstream `development` branch.
- [ ] The unit tests pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

- [ ] All the functions created/modified in this PR contain relevant docstrings.

PR #6584, opened 2019-11-08
added email attribute for organizers

### LINKED ISSUE #6522: Add Organizer email in get event details attribute
Event organizer email should be added to contact organizer through email .

https://github.com/fossasia/open-event-attendee-android/issues/1964

**Additional context**

akshat0047:
Fixes #6522 

It fixes the permission manager check which was breaking due to lack of event arguments being passed in UserDetail resource manager's decorator

- added Event model and event_id attribute field in decorator kwargs

####Result Comparison
Before
Screenshot_2019-11-16_20-02-05

After
- /v3/events/{event_id}/owner(As owner/admin)
Screenshot_2019-11-16_20-04-12

- /v3/events/{event_id}?include=owner(As owner/admin)
Screenshot_2019-11-16_20-10-10

- /v3/events/{event_id}/owner(As other)
Screenshot_2019-11-23_11-10-21

-/v3/events/{event_id}?include=owner(As other)
Screenshot_2019-11-23_16-22-22

#### Checklist

### DIFF
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -108,7 +108,7 @@ class UserDetail(ResourceDetail):
     """
 
     def before_get(self, args, kwargs):
-        if current_user.is_admin or current_user.is_super_admin or current_user:
+        if current_user or (kwargs['event_id'] and has_access('is_owner', event_id=int(kwargs['event_id']))):
             self.schema = UserSchema
         else:
             self.schema = UserSchemaPublic
@@ -281,7 +281,7 @@ class UserDetail(ResourceDetail):
                                             DiscountCode, EmailNotification, Speaker, User],
                                      fetch_key_url="notification_id, feedback_id, users_events_role_id, session_id, \
                   event_invoice_id, access_code_id, discount_code_id, email_notification_id, speaker_id, id",
-                                     leave_if=lambda a: a.get('attendee_id')),)
+                                     leave_if=lambda a: a.get('attendee_id') or a.get('event_id')),)
     schema = UserSchema
     data_layer = {'session': db.session,
                   'model': User,


akshat0047:
@iamareebjamal , this is identical to the changes which were tried in #6538 but as agreed by you an email is added in organizer relationship. Even I am wrong with my approach, apologies if the time span I took was not worth it, had few dual boot issues.

akshat0047:
Do I need to just include data email in the relationship, the compound document part?

iamareebjamal:
owner relationship needs to be fixed

akshat0047:
Now owner relationship generates a user specific url like v1/users/1 and not v1/events/1/owner, I think this is not the one because its different from other related urls, let me know if this wrong. @iamareebjamal

iamareebjamal:
If I include owner of event, it should be included successfully

akshat0047:
If include means post, I need to change related view as user_list, this way api is working with the url v1/events/<int:event_id>/owner. Is this the fix? But an event should have only one owner why we want to include owner of event

iamareebjamal:
> But an event should have only one owner why we want to include owner of event

I don't understand?

> If include means post, I need to change related view as user_list

I meant `/v1/events/12?include=owner` should include the owner information

app/api/__init__.py
```diff
@@ -88,12 +88,12 @@
 from app.api.users import UserList, UserDetail, UserRelationship
 
 # users
-api.route(UserList, 'user_list', '/users', '/events/<int:event_id>/organizers')
+api.route(UserList, 'user_list', '/users', '/events/<int:event_id>/organizers', '/events/<int:event_id>/owner')
```
iamareebjamal: This shouldn't be a list, so this is wrong

app/api/schema/events.py
```diff
@@ -261,7 +261,7 @@ def validate_timezone(self, data, original_data):
     owner = Relationship(attribute='owner',
                          self_view='v1.event_owner',
                          self_view_kwargs={'id': '<id>'},
-                         related_view='v1.user_detail',
+                         related_view='v1.user_list',
```
iamareebjamal: This is wrong

iamareebjamal:
Goal of this PR:

1. /events/1/owner should work
2. /events/1/?include=owner should work with owner information returned with the event JSON

akshat0047:
I was trying to fix the first goal, I will be using debugger to further inspect it today but even everything looks perfect the request is not entering the detail resource manager and returning object not found in postman. Atleast it should enter the UserDetail but print logs are not working.

Here we have owner relationship, user_detail in related view along with the kwargs
Screenshot_2019-11-11_01-29-49

followed by route for the view
Screenshot_2019-11-11_01-30-34

None of the print log is working though its working with /users/<int:id>
Screenshot_2019-11-11_01-31-12

Getting object not found error
Screenshot_2019-11-11_01-32-43


I regret I'm asking for too much but not before my complete efforts, this issue made me understand the flow of api completely except for the factories.

iamareebjamal:
Yes, it's unfortunate that the library is not transparent and works somewhat like magic and doesn't sufficiently log as well. What you can do is enable SQL logs and see which SQL is generated when accessing the owner endpoint. It'll show which ID is being used to get which table.

Thanks for sticking with the issue and trying to resolve it. It is not as easy as it sounds. Most people give up and leave

akshat0047:
Ok, I see. I checked the logs and a weird behaviour is, for event and user related request an identical set of logs is being created due to large schemas.
Another fact I noticed is if I remove /users/<int:id> and replace event_id with id in /events/<int:event_id>/owner, the owner endpoint starts working. Which means there is something wrong with interpretation of event id, its unable to access Resource Detail of User as GET, not even the before get method. Will be solving it soon, taking up adjacent issues to be consistent with contributions.

akshat0047:
Solved it! It was too short but things were getting stuck deep down in permission manager where the custom system to fetch model according to view_kwargs for is_user_itself permission check was breaking because entries for event_id were not present.

Please merge.

iamareebjamal:
Can you give a before after JSON response in PR body for reference

akshat0047:
> Can you give a before after JSON response in PR body for reference

Done

iamareebjamal:
That's admin output, please remove authentication and paste the response

iamareebjamal:
@akshat0047 Updates?

akshat0047:
Actually, I am having two consecutive end-sems on 18 and 19 and will be active again from 19th evening. By removing authentication I get access is forbidden, for which I added it in the leave-if lambda on line 284. Now one problem needs to be solved which will resolve #6600 also.

Fixing why curent_user is None inside /app/api/users.py

akshat0047:
I've almost solved it, need to fix either we want to give access to: 
- only is_user_itself
- limited to all and full to admin/super_admin

iamareebjamal:
limited to all and full to admin/super_admin

app/api/helpers/permission_manager.py
```diff
@@ -122,6 +122,13 @@ def is_user_itself(view, view_args, view_kwargs, *args, **kwargs):
     Otherwise the user can only access his/her resource.
     """
     user = current_user
+    if kwargs['source'] and kwargs['source'] == 'users':
+        if not user.is_admin and not user.is_super_admin and user.id != kwargs['user_id']:
+            view_kwargs['schema']= 'public'
+        else:
+            view_kwargs['schema']='private'
+        return view(*view_args, **view_kwargs)
```
iamareebjamal: Not correct, see the implementation in other parts of the server

app/api/helpers/permission_manager.py
```diff
@@ -122,6 +122,13 @@ def is_user_itself(view, view_args, view_kwargs, *args, **kwargs):
...
+
```
niranjan94: Codacy Issue found: [Trailing whitespace](https://app.codacy.com/app/fossasia/open-event-orga-server/pullRequest?prid=4467218)

niranjan94:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/open-event-orga-server/open-event-orga-server/pullRequest?prid=4467218&bid=15086452)

app/api/helpers/permission_manager.py
akshat0047: I understand, but in other parts of the server the API view is written differently checks are done inside each kwargs, but the user view has a global check of is_user_itself which needs to be carried out, if not we have to implement a jwt_required in before_get to fetch current_user followed by an admin check which will still miss is_user_itself check to provide private data.

And if we by default do is_user_itself check, we'll be unable to serve data to random users, limited one as the Access Denied message shows up.

My soluton is to do an additional check in is_user_itself for user view and behave differently

akshat0047:
There is one more problem is user view, it assigns each view_kwarg to behave as 'id' view_kwarg if it's present in the check, this serves the data of user but 'self' url served is of /v1/users/{user_id} and not of the relationship, Is this an issue. If yes the whole view needs to be re-written @iamareebjamal

app/api/helpers/permission_manager.py
akshat0047: Ok my changes failed circleci this should be wrong

akshat0047:
@iamareebjamal can we merge this if its right? or the codecov test plays an important role?

iamareebjamal:
Will review and merge

app/api/users.py
```diff
@@ -108,9 +108,12 @@ class UserDetail(ResourceDetail):
     """
 
     def before_get(self, args, kwargs):
-        if current_user.is_admin or current_user.is_super_admin or current_user:
+        if current_user:
             self.schema = UserSchema
-        else:
+        elif 'Authorization' in request.headers and (has_access('is_admin') or has_access('is_super_admin')):
```
iamareebjamal: or can be used, it's the same branch

app/api/users.py
```diff
@@ -108,9 +108,12 @@ class UserDetail(ResourceDetail):
     """
 
     def before_get(self, args, kwargs):
-        if current_user.is_admin or current_user.is_super_admin or current_user:
```
iamareebjamal: I'd prefer this syntax, what's problem in this?

app/api/users.py
```diff
@@ -108,9 +108,12 @@ class UserDetail(ResourceDetail):
...
+        if current_user:
             self.schema = UserSchema
-        else:
+        elif 'Authorization' in request.headers and (has_access('is_admin') or has_access('is_super_admin')):
+            self.schema = UserSchema
+        elif 'Authorization' in request.headers and (kwargs['event_id'] and has_access('is_owner',
+                                                     event_id=int(kwargs['event_id']))):
```
iamareebjamal: We want it to be visible to all users, not just owner of the event

app/api/users.py
akshat0047: 1- current_user is only fetched if the code encounters a jwt_required, excluding event_id from the decorator makes current_user null
2- If the third option allows every current_user why are we even checking is_admin or is_super_admin

iamareebjamal: 1. Other endpoints have worked around this by conditionally loading current_user using a decorator. Check those out
2. `or current_user` This section is wrong

app/api/users.py
akshat0047: Yess It is visible to all users, I've added a screenshot above. This check is for if the user itself request for owner then we can provide the complete info because we're excluding this case from the decorator we need to check for is_user_itself somewhere, this is the place

app/api/users.py
akshat0047: ok, instead I should remove authorization check or everyone will be able to access the complete use profile

iamareebjamal: No, everyone should not be able to check complete user profile

app/api/users.py
iamareebjamal: Then `or` should be used. It's looking like 2 different branches

app/api/users.py
akshat0047: 1- In this section current_user is something differentiating between whether the decorator is_user_itself, considered or not:

- importing it conditionally on the whole if-else check will fail the system to differentiate and will provide complete user detail for all

- importing it on 2nd elif will help with current_user.is_admin or curent_user.is_super_admin check but we will need has_access to check is_user_itself (equivalent to the code I'll update considering suggestion in the next hound comments)

2- Yes it was wrong but current_user check is indeed important for the request coming from decorators to take some form of schema

app/api/users.py
akshat0047: Implementing this

app/api/users.py
akshat0047: Yess

app/api/users.py
iamareebjamal: It's done in other places without requiring the Authorization check. Take a look

akshat0047:
- No authorization check
- merged the elif's
- has_access is_user_itself will check for admin and super_admin as well
- conditionally importing current_user will be of no use, we will have to check if the current_user is admin or not

akshat0047:
I checked other implementations for 4 (orders, feedbacks, tickets, events), if you want me to fetch current_user using jwt_required decorator or verify_jwt_request(), I'll be able to seperate admins and non admins...

But what about the current_user coming from is_user_itself who's the owner and need complete access, also the current_user of event_id not coming from decorator and is the owner and need complete access, adding one more is_user check in before_get will be heavier than the present solution @iamareebjamal

Though I need to generalize this for every view_kwargs passed in 'leave_if' attribute

akshat0047:
I am sorry, I'm really late post my exams. Getting back on this issue, went through the details, it will be of great help if you review the commit on Nov 24, 2019 if it was the solution

PR #6590, opened 2019-11-12
fix: DB Migrations

### LINKED ISSUE #6516: Discrepancy between migrations and production DB
Migrations in the app are not applied to production DB because it was just stamped to the latest migration without running them. In fact, the migrations can't even be run successfully on a database created from scratch. Since we are not in a position to correct the old migrations, the only thing we can do is to create new migrations to make production DB aligned with current schema or change the schema to current production DB semantics.

See Also: #6513 #6514

iamareebjamal:
Fixes #6516 

Many migrations were re-added maybe due to not being applied to production DB. This should never have been done. This creates a problem for new DB migrations as they fail. This creates a schema difference in production and development DB. This PR fixes old migrations. This is one way to deal with the discrepancy.

To be discussed if this is the best way


prateekj117:
@iamareebjamal WOW !!! I can only think about the amount of debugging required here. All the best !!!

prateekj117:
@iamareebjamal Just a thought. How about we write an automated script which arranges the `revision` and `down_revision` of every migration according to the time they were committed on GitHub.

iamareebjamal:
Good idea, but it'll also ruin the git blame. We can do that using the dates written inside the migration files as well, to see when were they created, however it won't be accurate depiction of in what order they were applied in, but still help

prateekj117:
@iamareebjamal Will start working on the script.

PR #6628, opened 2019-11-27
fix: Allow tickets to have same name as deleted ones

### LINKED ISSUE #6621: Cannot create ticket with name "Business Ticket" if former ticket with name "Business Ticket" was deleted
Cannot create ticket with name "Business Ticket" if former ticket with name "Business Ticket" was deleted. Error message appears.

codedsun:
Fixes #6621 : Modified unique constraints in ticket model

When deleting event the name was not changed of ticket which resulted in unique key constraint while creating another ticket with same name.

- unique constraint changed to name, event and deleted_at

#### Checklist


codedsun:
@iamareebjamal I have tested this on my local with every case
1. When a non-corganizer deletes
2. deleting through the frontend and also the API.

Working fine for me 👍

app/api/tickets.py
```diff
@@ -182,6 +182,11 @@ def before_update_object(self, ticket, data, view_kwargs):
                 raise UnprocessableEntity(
                     {'event_id': ticket.event.id}, "Event having paid ticket must have a payment method")
 
+        if ticket.deleted_at != data.get('deleted_at'):
+            if not has_access('is_coorganizer', event_id=ticket.event_id):
+                raise ForbiddenException({'source': ''}, 'Co-organizer access is required.')
```
iamareebjamal: This is already checked below in decorators

app/api/tickets.py
```diff
@@ -182,6 +182,11 @@ def before_update_object(self, ticket, data, view_kwargs):
                 raise UnprocessableEntity(
                     {'event_id': ticket.event.id}, "Event having paid ticket must have a payment method")
 
+        if ticket.deleted_at != data.get('deleted_at'):
```
iamareebjamal: This will be true when restoring tickets from deleted as well

app/api/tickets.py
codedsun: Ok will remove

app/api/tickets.py
codedsun: So should I remove this? and just add the line for append?

codedsun: We should add a check when data.get('deleted_at') is null correct?

iamareebjamal: Check if there's a feature to restore deleted tickets in admin panel. If there's not, then it won't need to be changed

codedsun: No, there's nothing as such @iamareebjamal

iamareebjamal:
OK, then create a script to do the same to existing deleted tickets

codedsun:
@iamareebjam scheduled job ? Please describe a little more. Thanks

iamareebjamal:
I'm thinking if we change the unique constraint from name to (name, deleted_at), this will be solved in much better way. Please try doing that instead

codedsun:
@iamareebjamal Why dont we have a check of state in the system? If we have a state like 
1 - Active 
2- Inactive
3 - Deleted

Can't this solve the problem?

iamareebjamal:
If a patient has sore throat, giving it medicine can fix the problem as well. Or cutting the patient's throat will also end the sore throat, we recommend medicines. Adding complexities like active inactive deleted pending just_about_to_start almost_ready pinky_promise is not a solution

codedsun:
I will do this today!

codedsun:
@iamareebjamal Currently the unique constraint are applied to name, eventId. Shal this be changed to name, deletedate and eventId.?

codedsun:
@iamareebjamal While adding the constraint in the __table_args__ and applying migrations, I am not able to add new tickets. I am getting error
```INFO:werkzeug:127.0.0.1 - - [03/Dec/2019 20:49:39] "POST /v1/tickets HTTP/1.1" 405 -```

Screenshot 2019-12-03 at 8 51 01 PM

**Comment by prateekj117** [comment]
@codesun Update your branch. Check Gitter.

**Comment by codedsun** [comment]
@iamareebjamal @prateekj117 Please check this!

**Comment by prateekj117** [comment]
@codedsun Also make a migration to handle the already existing deleted tickets.

**Comment by codedsun** [comment]
@prateekj117 How do i do this? Please explain more

**Comment by prateekj117** [comment]
@codedsun The approach you were following before was correct. Unique Constraint was already there.

**Comment by codedsun** [comment]
@prateekj117 Unique constraint was only for name and event-id. Not for deleted-at, hence the issue was caused.

**Comment by prateekj117** [comment]
@codedsun Ohh, ya. I think this approach is also right. In this case, you won't even need to handle the already existing cases.

**Comment by codedsun** [comment]
> @codedsun Also make a migration to handle the case of already existing deleted tickets.

@prateekj117 Guide me for this

**Comment by iamareebjamal** [comment]
@codedsun Have you tested and confirmed that it is working?

**Comment by prateekj117** [comment]
@codedsun Not required now. Though, I was asking for something similar to this: https://github.com/fossasia/open-event-server/pull/6386

File: app/api/tickets.py [hunk 3]
```diff
@@ -17,7 +17,7 @@
 from app.models.ticket import Ticket, TicketTag, ticket_tags_table
 from app.models.event import Event
 from app.models.ticket_holder import TicketHolder
-from app.api.helpers.exceptions import ConflictException, MethodNotAllowed, UnprocessableEntity
```
iamareebjamal: Revert

iamareebjamal:
You've just added constraint in model. Generate the migration as well

codedsun:
@iamareebjamal 
Created a ticket named Paid ticket

image

Deleted and created the ticket with same name, working fine
image

app/api/tickets.py
codedsun: Added!

codedsun:
@iamareebjamal Completed, check now!

iamareebjamal:
Also check that you cannot create "Paid Ticket" if it is not deleted

codedsun:
@iamareebjamal Throws error of Ticket Already exist, working fine :)

image

image

PR #6599, opened 2019-11-15
chore: container script run through CLI args

codedsun:
Fixes #6598 

#### Short description of what this resolves:
Container script now can be run by CLI args instead of envt variables

#### Changes proposed in this pull request:

- Accessed CLI Args

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The unit tests pass locally with my changes 
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

- [x] All the functions created/modified in this PR contain relevant docstrings.

PR #6651, opened 2019-12-10
fix:changed celery health-check endpoint

### LINKED ISSUE #6471: Implement docker healthcheck and fix endpoint

codedsun:
Fixes #6471 

Removed celery check from health-check of app, added a new end point with queryparam for celery check

#### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,4 +26,5 @@ WORKDIR /data/app
 ADD . .
 
 EXPOSE 8080
+HEALTHCHECK CMD curl --fail http://localhost:5000/health-check || exit 1
 CMD ["sh", "scripts/container_start.sh"]
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -4,7 +4,7 @@ import os.path
 from envparse import env
 
 import sys
-from flask import Flask, json, make_response
+from flask import Flask, json, make_response, request
 from flask_celeryext import FlaskCeleryExt
 import sentry_sdk
 from sentry_sdk.integrations.flask import FlaskIntegration
@@ -229,12 +229,13 @@ def make_celery(app=None):
 
 
 # Health-check
-health = HealthCheck(current_app, "/health-check")
-health.add_check(health_check_celery)
+health = HealthCheck()
+health_dependencies = HealthCheck()
 health.add_check(health_check_db)
 with current_app.app_context():
     current_app.config['MIGRATION_STATUS'] = check_migrations()
 health.add_check(health_check_migrations)
+health_dependencies.add_check(health_check_celery)
 
 
 # http://stackoverflow.com/questions/9824172/find-out-whether-celery-task-exists
@@ -285,3 +286,11 @@ def internal_server_error(error):
 
 if __name__ == '__main__':
     current_app.run()
+
+
+@app.route('/health-check')
+def health_check():
+    param = request.args.get('q')
+    if param == 'celery':
+        return health_dependencies.run()
+    return health.run()
--- a/app/views/healthcheck.py
+++ b/app/views/healthcheck.py
@@ -1,6 +1,5 @@
 from errno import errorcode
 
-from celery.task.control import inspect
 from flask import current_app
 from redis.exceptions import ConnectionError
 
@@ -13,7 +12,7 @@ def health_check_celery():
     :return:
     """
     try:
-        d = inspect().stats()
+        d = current_app.celery.inspect().stats()
         if not d:
             capture_message('No running Celery workers were found.')
             return False, 'No running Celery workers were found.'


codedsun:
@iamareebjamal Currently I am getting error for celery endpoint 

```

File "/usr/local/lib/python3.7/site-packages/healthcheck/healthcheck.py", line 139, in run_check
        elapsed_time = end_time - start_time
        # Reduce to 6 decimal points to have consistency with timestamp
        elapsed_time = float('{:.6f}'.format(elapsed_time))
 
        if not passed:
            msg = 'Health check "{}" failed with output "{}"'.format(checker.__name__, output)
            logger.error(msg)
 
        timestamp = time.time()
        if passed:
            expires = timestamp + self.success_ttl
AttributeError: 'tuple' object has no attribute '__name__'

```

app/__init__.py
```diff
@@ -229,12 +229,13 @@ def make_celery(app=None):
...
 # Health-check
-health = HealthCheck(current_app, "/health-check")
-health.add_check(health_check_celery)
+health = HealthCheck()
+health_dependencies = HealthCheck()
 health.add_check(health_check_db)
 with current_app.app_context():
     current_app.config['MIGRATION_STATUS'] = check_migrations()
 health.add_check(health_check_migrations)
+health_dependencies.add_check(health_check_celery())
```
iamareebjamal: See what was written previously and what you have written

codedsun: @iamareebjamal Previously it was written 
`health.add_check(health_check_celery)` But how do we add to same instance of health-check when we want only celery result at a seperate endpoint.

codedsun: I have made a different instance to get the result of celery

iamareebjamal: Again you are looking at a completely different thing than what I am trying to point out

```diff
-health.add_check(health_check_celery)
+health_dependencies.add_check(health_check_celery())
```

iamareebjamal: Can you see any other difference than health and health_dependencies?

codedsun: @iamareebjamal Please read the comment sir. I am looking at the same line

codedsun:
```
health.add_check(health_check_celery)
```

codedsun: @iamareebjamal I am at the same point and direction you said. Also I wrote why I did this, please read.

codedsun: Yes I saw. Sorry

codedsun: But it's a function, in python dont they have ( ) braces when calling a function. Dumb me. Apologies

codedsun:
@iamareebjamal Done. thanks

app/__init__.py
codedsun: Writing kotlin and java daily had made me done this :( Will not repeat again. Thanks

codedsun:
Is any task pending in this, if not. Please check. Thankyou

codedsun:
@iamareebjamal To restart brain dead containers, we have to use docker swarm?

https://www.katacoda.com/courses/docker-orchestration/healthcheck

I have added the health check command in dockerfile

Dockerfile
```diff
@@ -26,4 +26,5 @@ WORKDIR /data/app
 ADD . .
 
 EXPOSE 8080
+HEALTHCHECK CMD curl --fail http://localhost:5000/health-check || exit 1
```
iamareebjamal: It should check proper healthcheck. Normal for API container and celery for celery container

codedsun: Okay that means I have to add for respective containers in docker-compose file. THanks

iamareebjamal: No

codedsun: then two healthcheck command in this DOCKERFILE? like this? 

```
HEALTHCHECK CMD curl --fail http://localhost:5000/health-check || exit 1
HEALTHCHECK CMD curl --fail http://localhost:5000/health-check?q=celery || exit 1
```

iamareebjamal: There can't be 2 HEALTHCHECKs

codedsun: @iamareebjamal Their can be only 1 healthcheck command in dockerfile but we have 2 different URLs for healthcheck(one for app, other for celery). Also we dont' have to add the healthcheck in docker-compose file for respective containers. Then how can I do it? Can you help me with this. 

Also, To restart brain dead containers, we have to use docker swarm?

iamareebjamal: See container start shell file

codedsun: @iamareebjamal Got it have to access the deployment variable in dockerfile and based on value of it add healthcheck? correct?

codedsun:
@iamareebjamal May be I should correct this https://github.com/fossasia/open-event-server/pull/6599 and it will get me the args directly from the docker-compose file to the dockerfile and based on it I will add the the healthcheck. Should I follow this way and also am I correct?

PR #6656, opened 2019-12-12
feat: Optimize link generation

iamareebjamal:
Related: #6666 
Upstream: https://github.com/marshmallow-code/marshmallow-jsonapi/pull/278

Before:
/v1/events/ = 2.7 seconds == 2700 ms (20 events)
/v1/events/<id> = 250 ms

After:
/v1/events/ = 150 ms (20 events)
/v1/events/<id> = 30 ms

### DIFF
--- a/app/api/schema/__init__.py
+++ b/app/api/schema/__init__.py
@@ -7,4 +7,5 @@ def serialize(self, attr, obj, accessor=None):
         return super().serialize(attr, obj, accessor)
     return self._serialize(None, attr, obj)
 
+
 Relationship.serialize = serialize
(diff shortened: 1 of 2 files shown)


app/api/schema/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Monkey Patch Marshmallow JSONAPI
+from marshmallow_jsonapi.flask import Relationship
+
+
+def serialize(self, attr, obj, accessor=None):
```
kushthedude: @iamareebjamal @prateekj117 This change has broke the user endpoint

iamareebjamal: This didn't

PR #6692, opened 2019-12-23
fix: CompletedAt attribute of Completed/Placed orders

### LINKED ISSUE #6691: Ticket Dashboard: Completed_At Property of Order is not timezone aware.
**Describe the bug**

Completed_At Property of Order is not timezone aware.

**To Reproduce**
1. Make any completed order
2. Visit order dashboard
3. See error

**Expected behavior**

Time should be correct

**Additional context**

Screenshot 2019-12-23 at 8 14 42 AM

kushthedude:
Fixes #6691

### DIFF
--- a/app/api/orders.py
+++ b/app/api/orders.py
@@ -208,7 +208,7 @@ class OrdersListPost(ResourceList):
             send_notif_to_attendees(order, current_user.id)
 
             if order.payment_mode in ['free', 'bank', 'cheque', 'onsite']:
-                order.completed_at = datetime.utcnow()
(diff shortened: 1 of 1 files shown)


iamareebjamal:
This will make it timezone unaware. Again, read about timezones, and fix the issue which is in frontend not in server

kushthedude:
No completedAt is getting assigned at Server, also we are not making any
changes at FE. I checked it. We are showing it as it is stored in db.

iamareebjamal:
Exactly, show it in the correct timezone on frontend

kushthedude:
Got it

PR #6700, opened 2019-12-29
feat: run docker container as non root user with permissions

### LINKED ISSUE #6505: Run docker container as non root user
Currently, the app and celery run as root in docker containers 

They should be run as normal users

codedsun:
Fixes #6505 

`WIP`

feat: run docker container as non root user with permissions

#### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,5 +25,13 @@ RUN apk --no-cache add postgresql-libs ca-certificates libxslt jpeg zlib file li
 WORKDIR /data/app
 ADD . .
 
+RUN adduser -S open_event_user
+# Gives permission to open_event_user
+RUN chown -R open_event_user /data/app
+RUN chown -R open_event_user ./static
+RUN chown -R open_event_user ./generated
+# Change to non-root privilege
+USER open_event_user
+
 EXPOSE 8080
 ENTRYPOINT ["sh", "scripts/container_start.sh"]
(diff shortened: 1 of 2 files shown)


codedsun:
@iamareebjamal Test please

Dockerfile
```diff
@@ -25,5 +25,13 @@ RUN apk --no-cache add postgresql-libs ca-certificates libxslt jpeg zlib file li
 WORKDIR /data/app
 ADD . .
 
+RUN adduser -S open_event_user
+# Gives permission to open_event_user
+RUN chown -R open_event_user /data/app
+RUN chown -R open_event_user ./static
+RUN chown -R open_event_user ./generated
```
iamareebjamal: This is run at creation of the docker image, so it ran on your machine only. It'll never run on server

codedsun: Then to run on the server, what should be done? Yes while building the image this was done. But in tomorrow PR how does it work?

iamareebjamal: Because that folder is inside the container and these are on host

codedsun: @iamareebjamal - Okay but can you show me the direction about how to solve this now. We want this to run without building the image.?

codedsun: Means what exactly the issue is here. We wont build the image again on server and I have to add these host files permission which are outside the container.

iamareebjamal: Searching for better solution

PR #6746, opened 2020-01-16
fix: migration for country select type

kushthedude:
Refers https://github.com/fossasia/open-event-frontend/issues/3780


iamareebjamal:
Where does it fetch the options from?

kushthedude:
> Where does it fetch the options from?

Handled on FE from `utils`

PR #6755, opened 2020-01-18
feat: Addition of is_promoted field in event

kushthedude:
Refers https://github.com/fossasia/open-event-frontend/issues/3658


kushthedude:
> What's the difference between featured and promoted here?

None in server side, On FE promoted events will be shown before the general events in the list of upcoming events as requested in the issue.

kushthedude:
@iamareebjamal Please check!

iamareebjamal:
The frontend should not just show promoted events but it should show promoted events first

kushthedude:
This field will be needed anyway.

iamareebjamal:
More changes will be needed on server if the behaviour is to be carried out

kushthedude:
Can't we just query these events before the other while rendering on FE?

iamareebjamal:
OK

PR #6824, opened 2020-02-08
chore: Pin werkzeug version

iamareebjamal:
Since we are directly depending on werkzeug, it should
be tracked in requirements file so that transitive upgrade
does not break the project


iamareebjamal:
https://travis-ci.org/fossasia/open-event-server/builds/647520319?utm_source=github_status&utm_medium=notification

https://github.com/pallets/werkzeug/issues/1714

PR #6850, opened 2020-02-19
fix: Cron entries of period tasks

### LINKED ISSUE #6826: After Event Mail: Mails are getting sent in every 5 hours.
The scheduled job for the after event mail is misbehaving. I have been getting 3 emails periodically every 5 hours once my event was over.

iamareebjamal:
Fixes #6826

### DIFF
--- a/app/api/helpers/scheduled_jobs.py
+++ b/app/api/helpers/scheduled_jobs.py
@@ -302,17 +302,27 @@ def send_monthly_event_invoice():
 def setup_scheduled_task(sender, **kwargs):
     from celery.schedules import crontab
 
-    sender.add_periodic_task(crontab(hour='*/5', minute=30), send_after_event_mail)
-    sender.add_periodic_task(crontab(minute=0, hour=0), send_event_fee_notification)
+    # Every day at 5:30
+    sender.add_periodic_task(crontab(hour=5, minute=30), send_after_event_mail)
+    # Every 1st day of month at 0:00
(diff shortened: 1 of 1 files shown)


iamareebjamal:
Almost every cron entry was wrong. And I don't blame anyone. Took me an hour to guess what the period should be, is it  actually that period or is it a bug of the author

@kushthedude @mrsaicharan1 Please see my comments and confirm this is what it should be. This is what was happening in APS before moving to celery integration.

I don't get the point of 5:00 and 5:30, anyone reading the previous code will assume that it is supposed to run every 5 hours or 5 hours 30 minutes, whereas it actually runs at 5:00/5:30 every day. That's why it caused confusion when migrating to celery. I really hope it is not to sync with +5:30 timezone of India, that would be very embarrasing

iamareebjamal:
@kushthedude Can you please tell if you know something about this so we can make it live on production?

iamareebjamal:
Merging since getting no response

kushthedude:
> @kushthedude Can you please tell if you know something about this so we can make it live on production?

@iamareebjamal I was travelling, hence could not give any response before seeing the actual code.
I don't know why 5:30/5:00 why is added. I think it is actually correct now as it makes no sense in sending the after event mail after every 5 hours.
Still, @mrsaicharan1 may know more if I am wrong/missed something.

iamareebjamal:
OK. Thank You

PR #7012, opened 2020-05-15
test: Re-enable discount code creation test

### LINKED ISSUE #7011: Discount Code creation failed
After merging of #6985, discount code creation fails:

For some bizarre reason, user_id was passed in constructor of DiscountCode and then set to marketer_id :man_shrugging: 
Anyway it was still working because for another bizarre reason, marketer passed from the API was changed to user_id on server :man_shrugging: :shrug: 

So now, because the constructor is removed, there is no user_id kwarg in Discount Code and it throws. ~~The build didn't break because no dredd test was present for passing marketer ID.~~ I am not sure why is it even needed, but add the test for now to ensure it doesn't break in future

Edit: It was present in Dredd tests but disabled in #4854

iamareebjamal:
Fixes #7011

### DIFF
--- a/docs/api/blueprint/discount_codes.apib
+++ b/docs/api/blueprint/discount_codes.apib
@@ -167,18 +167,12 @@ Create a new Discount Code for event. (Only Admin)
         {
             "data": {
                 "relationships": {
-                    "events": {
+                    "event": {
                         "links": {
-                            "self": "/v1/discount-codes/4/relationships/events",
-                            "related": "/v1/discount-codes/4/events"
+                            "self": "/v1/discount-codes/1/relationships/event",
+                            "related": "/v1/discount-codes/1/event"
                         }
-                    },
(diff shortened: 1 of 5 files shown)


docs/api/blueprint/discount_codes.apib
```diff
@@ -167,18 +167,12 @@ Create a new Discount Code for event. (Only Admin)
...
-                    "tickets": {
+                    "event": {
                         "links": {
-                            "self": "/v1/discount-codes/1/relationships/tickets",
-                            "related": "/v1/discount-codes/1/tickets"
+                            "self": "/v1/discount-codes/1/relationships/event",
+                            "related": "/v1/discount-codes/1/event"
                         }
-                    }
+                     }
```
iamareebjamal, line 175:
```suggestion
                    }
```

PR #7098, opened 2020-06-30
fix: Add proper unique constraints and remove soft deletion from models

### LINKED ISSUE #7093: Change unique constraints of soft deleted models
All models were changed to soft deletion without much thought about which ones make sense to be soft deleted and which ones don't. This leads to many errors as some models have unique constraints like discount code and when they are deleted and same code is added again, it throws an error. Thus, we need to revisit which models make sense to add soft deletion to, which should never be deleted (order, ticket, discount code) and which should not have soft deletion (custom forms)

This issue will focus on at least adding correct unique constraint, like (discount code, deleted at) instead of just discount code so that deleted models are not considered for uniqueness. And as we fix those, we'll remove soft deletion from some models and also add validations to prevent deletion in permanent records

Haider8:
Fixes #7093 

Not all models requires soft deletion so we have to remove it from models like `CustomForms` and `EventSubTopic`. 

In some models (like `DiscountCode`) we need to add a unique constraint on `deleted_at` column.

#### Checklist


app/models/discount_code.py
```diff
@@ -13,7 +13,9 @@
 class DiscountCode(SoftDeletionModel):
     __tablename__ = "discount_codes"
     __table_args__ = (
-        UniqueConstraint('event_id', 'code', name='uq_event_discount_code'),
+        UniqueConstraint(
+            'event_id', 'code', 'deleted_at', name='uq_event_discount_code_deleted_at'
```
iamareebjamal, line 17: Don't add deleted_at in constraint name, it's an implementation detail

Haider8: Remove from all?

iamareebjamal: Yes

iamareebjamal:
Please fix the tests

iamareebjamal:
Also, for all the models where deleted_at was removed, you need to delete the items with non null deleted_at before removing the column

Haider8:
@iamareebjamal  There are 4 apib dredd fails like in `GET /v1/events/1/custom-form`. So, should I change that now?

iamareebjamal:
Yes

iamareebjamal:
> Also, for all the models where deleted_at was removed, you need to delete the items with non null deleted_at before removing the column

.

Haider8:
@iamareebjamal Done

migrations/versions/rev-2020-06-30-20:07:55-f177c14710f1_.py
```diff
@@ -18,6 +18,9 @@
 
 def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
+    op.execute('DELETE FROM custom_forms WHERE deleted_at IS NOT NULL')
+    op.execute('DELETE FROM event_sub_topics WHERE deleted_at IS NOT NULL')
```
iamareebjamal, line 22: Add equivalent statements in downgrade as well

iamareebjamal:
Will merge after manual testing

app/models/ticket_fee.py
```diff
@@ -1,12 +1,14 @@
...
 
+from app.models.base import SoftDeletionModel
+
 from app.models import db
 
 DEFAULT_FEE = 0.0
 
 
-class TicketFees(db.Model):
+class TicketFees(SoftDeletionModel):
```
iamareebjamal, line 11: Don't add more soft deletion

migrations/versions/rev-2020-06-30-20:07:55-f177c14710f1_.py
Haider8: What will be the equivalent?

app/models/ticket_fee.py
Haider8: Removed :+1:

app/models/permission.py
```diff
@@ -1,13 +1,17 @@
 from app.models import db
 
+from app.models.base import SoftDeletionModel
 
-class Permission(db.Model):
+
+class Permission(SoftDeletionModel):
```
iamareebjamal, line 6: Here as well. Don't add more soft deletion. It has already complicated matters a lot

Haider8: :+1:

PR #7254, opened 2020-09-10
fix: allow refund_policy to be none

snitin315:
for https://github.com/fossasia/open-event-frontend/issues/5036

### DIFF
--- a/app/models/event.py
+++ b/app/models/event.py
@@ -119,9 +119,7 @@ class Event(SoftDeletionModel):
     ical_url = db.Column(db.String)
     xcal_url = db.Column(db.String)
     is_sponsors_enabled = db.Column(db.Boolean, default=False)
-    refund_policy = db.Column(
-        db.String, default='All sales are final. No refunds shall be issued in any case.'
-    )
+    refund_policy = db.Column(db.String)
     is_stripe_linked = db.Column(db.Boolean, default=False)
     live_stream_url = db.Column(db.String)
     webinar_url = db.Column(db.String)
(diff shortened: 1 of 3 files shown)


app/models/event.py
```diff
@@ -119,9 +119,7 @@ class Event(SoftDeletionModel):
     ical_url = db.Column(db.String)
     xcal_url = db.Column(db.String)
     is_sponsors_enabled = db.Column(db.Boolean, default=False)
-    refund_policy = db.Column(
-        db.String, default='All sales are final. No refunds shall be issued in any case.'
-    )
+    refund_policy = db.Column(db.String)
```
snitin315, line 122: we need a migration for past events, right?

iamareebjamal:
Yes, we need to set refund policy to null for previous events.

iamareebjamal:
Thinking about it now, if some user changed their refund policy using the API, making it null will lead to data loss. So, wait for Mario to tell what should be done

snitin315:
> Thinking about it now, if some user changed their refund policy using the API, making it null will lead to data loss. So, wait for Mario to tell what should be done

I think we can apply migration only to events having the default `refund_policy` message.

mariobehling:
In regards to events we should delete the policy. There was not any organizer who edited the policy, we just introduced it a few hours ago.

Is that easy to do? I dont think we need migrations.

iamareebjamal:
We didn't add it a few hours ago. We used it as default refund policy for all events in the backend for a long time ago

PR #7256, opened 2020-09-10
feat: add is_custom in social_link model

snitin315:
for https://github.com/fossasia/open-event-frontend/pull/5038


iamareebjamal:
We don't want this. For the backend, there is no distinction between a custom and normal link. All that should be handled in the frontend. Simply check that the identifier is among the supported link and if it is not, then it is custom. A simple getter can implement that in the frontend model

snitin315:
I will update 👍🏻

iamareebjamal:
Besides this will create horrible inconsistencies. What if someone sends facebook link with `isCustom` true or boobalooba link with `isCustom` false. Then the backend would need to keep set of supported linkes to check if it is custom or not. And adding a new supported link will need change in both frontend and server. Even if it was needed in server, a column would not be needed at all because just a property in the model will be able to tell if the link is custom or not. We should never add columns for data which can be derived from other data. That's redundancy and causes integrity issues. In relational data model, there must be a single source of truth. Previous devs added such fields which cause problems till today.

PR #7320, opened 2020-10-04
chore(docs): Improve local installation guide

hongquan:
Fixes #

This PR is to improve installation guide for development.

After this PR, another improvement in settings system should be added to make our software adapt better with different Postgres connection methods for various circumstance. For example:

- For local development: Favor "ident" method (just need same username, no need password).
- For prod where PostgreSQL stays in the same machine as backend: Favor "ident" method.
- For container-based deployment: Follow database URL (because PostgreSQL run in other container), where password may not even needed, if we configure Postgres container to "trust" web app container.

- Short command to install dependency softwares.
- Favor `virtualenvwrapper` for creating virtual environment (explanation is included).
- Favor "ident" connection method for PostgreSQL in development. 

#### Checklist


docs/installation/local.md
```diff
@@ -2,169 +2,177 @@
...
-## Steps
+## Get a copy of source code
 
-Make sure you have the dependencies mentioned above installed before proceeding further.
+- Clone the Open Event Server repository (from the development branch) and `cd ` into the directory.
 
-* **Step 0** - Clone the Open Event Server repository (from the development branch) and ```cd ``` into the directory.
 ```sh
-git clone -b development https://github.com/fossasia/open-event-server.git
+git clone https://github.com/fossasia/open-event-server.git
```
iamareebjamal, line 14: It is important that only development branch gets cloned and not the gh-pages and other branches

docs/installation/local.md
```diff
@@ -2,169 +2,177 @@
...
 cd open-event-server
 ```
-**Note :** If you want to contribute, first fork the original repository and clone the forked repository into your local machine followed by ```cd``` into the directory
+
+- If you want to contribute, first fork the original repository to your GitHub profile and clone that fork into your local machine, followed by `cd` into the directory
+
 ```sh
-git clone https://github.com/USERNAME/open-event-server.git
+git clone git@github.com:USERNAME/open-event-server.git
```
iamareebjamal, line 21: It may not be necessary that beginners have used SSH method for GitHub authentication. Most beginners use HTTPS

docs/installation/local.md
```diff
@@ -2,169 +2,177 @@
...
-sudo -H pip3 install -r requirements.txt
-```
-hint: You may need to upgrade your pip version and install following packages if you encounter errors while installing the requirements.
+  + To let your personal fork up-to-date with the original FOSSASIA repo, add original repo to "remote" list and regularly fetch its new content.
 
-**Note:** For Mac OS Sierra users, if you get an error that 'openssl/aes.h' could not be found when installing requirements.txt using pip, try the steps shown here - [OSX openssl header error](https://tutorials.technology/solved_errors/1-OSX-openssl_opensslv_h-file-not-found.html)
+  ```sh
+  git remote add upstream https://github.com/fossasia/open-event-server.git
+  git fetch upstream
+  git merge upstream/development
```
iamareebjamal, line 34:
```suggestion
  git pull -r upstream development
```

docs/installation/local.md
```diff
@@ -2,169 +2,177 @@
...
 
-## Using pipenv
+Python virtual environment is a way to keep this project's collection of libraries not interfere with other projects or the system. There are many ways to create Python virtual environment (all have the same core), like:
 
-Using pipenv, you will not need to set up virtualenv. It will do it automatically for you
+- Use Python built-in tool:
 
-To setup a virtual environment and install the dependices, enter in a terminal
+  ```sh
+  python3 -m venv my-project-name
```
iamareebjamal, line 69:
```suggestion
  python3 -m venv .venv
```

.venv is a recognized and gitignored folder

scripts/heroku.sh
```diff
@@ -5,6 +5,8 @@ export INTEGRATE_SOCKETIO=false
 # socketio has problems with celery "blocking" tasks
 # also socketio is not used in a celery task so no problem to turn it off
 chmod -R 0777 ./static
+# FIXME: Need to revisit this, why "&" at the end? Because it causes logging hidden,
+# making us difficulty finding bug if something goes wrong.
```
iamareebjamal, line 9: This runs on heroku, so we need to run both commands using same script. The logs are not hidden  when using &

docs/installation/local.md
```diff
@@ -2,169 +2,177 @@
...
+### For Debian/Ubuntu
 
-Open a terminal and enter the following commands to setup a virtual environment
+Enter the project source folder and run:
 
 ```sh
-sudo apt-get install python3.7
-virtualenv -p python3.7 venv
-. venv/bin/activate
+xargs -a deb-packages.txt sudo apt install
```
iamareebjamal, line 57: I think we should keep the dependencies explicitly listed here to avoid indirection and it's easier to track updates as well

deb-packages.txt
```diff
@@ -0,0 +1 @@
+python3-dev python3-venv libjpeg-dev postgresql redis-server libpq-dev libffi-dev virtualenvwrapper
```
iamareebjamal, line 1: virtualenvwrapper is not a strict requirement and thus can be listed as additional install in the recommendation section

.env.example
```diff
@@ -1,14 +1,17 @@
...
 # TEST_DATABASE_URL=postgresql://test@localhost:5433/test
 
 APP_CONFIG=config.DevelopmentConfig
 
-POSTGRES_USER=open_event_user
-POSTGRES_PASSWORD=opev_pass
-POSTGRES_DB=open_event
+# POSTGRES_USER=open_event_user
+# POSTGRES_PASSWORD=opev_pass
+# POSTGRES_DB=open_event
```
iamareebjamal, line 15: If not needed, these lines can be removed

.env.example
```diff
@@ -1,14 +1,17 @@
-DATABASE_URL=postgresql://open_event_user:opev_pass@localhost:5432/oevent
+DATABASE_URL=postgresql:///oevent
+# DATABASE_URL=postgresql://open_event_user:opev_pass@localhost:5432/oevent
+TEST_DATABASE_URL=postgresql:///oevent
```
iamareebjamal, line 3: TEST_DATABASE_URL should be different from DATABASE_URL

docs/installation/local.md
hongquan: image

Without `-b` option, Git will choose the default branch, and this repo already set "development"as default branch.

docs/installation/local.md
hongquan: We should teach beginner the good practice. I already mention the benefit of SSH over HTTPS in README (no need to type password again and again).

docs/installation/local.md
hongquan: Are you sure you want `git pull -r`? Because `-r` means `rebase`, and `rebase` will rewrite Git history, which later on, will cause a lot of messes.

docs/installation/local.md
hongquan: I'm against placing virtual environment inside source folder, with explanation included in README.
In summary, if doing so, when you do some action involving scan current folder, you will accidentality scan files in the virtual environment as well, which slow down your task, or worse, can cause some unexpected side effect.

scripts/heroku.sh
hongquan: I remember wrong about behaviour of `&`. Yes. It doesn't hide the logs. Thanks for correcting me.

But running many commands from the same shell script then send to background like this will make the log from them mixed together, which cause trouble for debugging.

But if this is Heroku requirements, ok, let's pass on it.

deb-packages.txt
hongquan: Ok.

docs/installation/local.md
hongquan: But this `deb-packages.txt` will help deploy process, it allows deploy script to automatically install new dependency if it is added.

hongquan: I've just added a note to README, to let developer know that the list of dependencies are in that file.

docs/installation/local.md
prateekj117: @hongquan It will clone all branches. Try running `git branch -a` after simply running git clone.

docs/installation/local.md
prateekj117: @hongquan It would be one more barrier to cross for a beginner before starting to contribute here. Also, I think it's fine if they use `https` at the very beginning.

docs/installation/local.md
prateekj117: It would just rebase their branch with `upstream development` (Which should ideally be the case if they make a new branch to solve an issue, and if the developer is in the `development` branch itself, it would just update it). Why would it cause a mess?

docs/installation/local.md
prateekj117: I think it's more of personal preference. I personally prefer to have it inside the source directory so that if I ever want to see how things are implemented, I have easy access to the libraries it is using. Especially because we are using a fork of `flask-rest-jsonapi` which is maintained by FOSSASIA itself.

docs/installation/local.md
```diff
@@ -2,169 +2,177 @@
...
-```sh
-pipenv --python 3.7.3 install
-```
+- Use third-party tools:
+
+  ```sh
+  mkvirtualenv my-project-name -p /usr/bin/python3
+  ```
+
+I (Quân) recommend [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/) because:
```
prateekj117, line 78: It's better not to use names in the documentation. @iamareebjamal will be able to tell better, though I am pretty sure these points as well as personal opinions shouldn't be added to README.

prateekj117 (changes requested):

docs/installation/local.md
iamareebjamal: We can provide an additional section below saying if you are using SSH, use the command below.
Please realize the level of beginners approaching the repo is such that even when the README explicitly states that Python 3.7 is required to run the project, they run it on 3.6 or 3.8. Or in frontend repo, they miss a command before starting a server and regularly ask what's the issue. If we write a command in README which will not work right away for most users, they'll most likely not set up SSH with GitHub, they'll either complain or quit. Pragmatism beats perfectionism here, unfortunately.

docs/installation/local.md
iamareebjamal: Yes, this is important because merge will merge their local changes on development branch with upstream changes and that will mean their branches made from development branch will contain unnecessary changes from other PRs. We should explicitly state that you need to create a new branch for working, not committing on development branch, and this will ensure this doesn't happen. Besides, even if they run the command, the work will not be lost, rather replayed after the upstream commits

docs/installation/local.md
iamareebjamal: But your command here will create the environment variable in the project as well. Here, the only change I am suggesting is to use a standard name. You have mentioned your preferred method below which if contributors want, can use. The above command will be directly be copied by students and lo and behold, they'll have a `my-project-name` folder in their repo which they'll commit in their next PR with all the libraries they installed in it. Besides, having separate folder of virtual environments:

1. Requires 3rd party tools not available in python distribution and learn its syntax
2. Difficult to manage as instead of running a project local command `source .venv/bin/activate`, they have to run different commands for different projects depending on the tool of choice
3. Makes it difficult for us to write guides or help users as it is not as simple as to just say to run `source .venv/bin/activate`

But anyway, you already have written your recommendation of virtualenv below and should the student choose it, they can use those instructions. But this instruction is using the previous method so at least it should be causing least friction by using a standard .gitignored name

docs/installation/local.md
iamareebjamal: Since we use docker alpine to build the image, it doesn't help us in deployment particularly. Where else is it being used to install the dependencies?

docs/installation/local.md
hongquan: > if they make a new branch to solve an issue, and if the developer is in the development branch itself, it would just update it

@prabhupant `rebase` means "rewrite Git history" and that causes the mess.
"merge" also let the current branch have update from development branch, but via "merge", not causing "rewrite Git history" and it is safer.

This is from my years of experience, I don't know how to make you understand if you don't stumble upon this circumstance. But as someone who have experience, I need to tell beginner to avoid the bad practice, which just waste time to recover.

To summary:

- "Rebase / Rewrite Git history" is safe if your commits are still in local (not push to remote repo yet).
- "Rebase / Rewrite Git history" will cause mess if you are rewriting the commits which have been pushed before.

docs/installation/local.md
hongquan: @iamareebjamal If you write down:

```sh
sudo apt install libpq-dev
brew install python@3
```

as before this PR, does it help you build Alpine Docker container?

The old README doesn't help you build Alpine Docker container, why do you use Alpine as excuse to refuse this change?

docs/installation/local.md
hongquan: @prabhupant  I got lost in this discussion thread, so I replied in https://github.com/fossasia/open-event-server/pull/7320#discussion_r502944325

@iamareebjamal Is your comment to answer @prabhupant or me?

 > "contain unnecessary changes from other PRs."

The changes you referred as "unnecessary" are actually the ones which went into "upstream" FOSSASIA repo, it is not "unnecessary" (if it is unnecessary, it wouldn't be accepted to upstream).

If you don't incorporate those changes to your branch, later on, you will still do so, or your PR will be in "conflict" state. The later you merge (from upstream), the bigger that conflict grows, and later you have to make bigger effort to solve the conflict.

On the other hand, those "upstream" change is not "unnecessary" because you have to make sure your code is in harmony with code of other parts. Merging early will let you know early what need to change to make sure your code still work well with other parts.

docs/installation/local.md
hongquan: @prateekj117  @iamareebjamal 

> they'll have a my-project-name folder in their repo

Sorry, when I write down "python3 -m venv my-project-name", I think about that user should run it outside the repo. But everyone read here will think about running it inside the repo. I'm wrong here. I agree to change to `python3 -m venv .venv`.

docs/installation/local.md
hongquan: @prateekj117 I mention my name, just mean that I'm responsible for that guide, so that someone can look for me to ask for more (instead of asking you when you don't recommend that method).

> personal opinions shouldn't be added to README

This is the guide that I want to send to beginners who join this project. If I don't add to README, where should I write? I don't want to let the beginners waste time with not-good way of doing.

docs/installation/local.md
hongquan: @prateekj117 Agree. I will fix.

docs/installation/local.md
iamareebjamal: > But this deb-packages.txt will help deploy process, it allows deploy script to automatically install new dependency if it is added.

I was commenting about this. Since we use alpine builds, debian dependencies don't help us in deployment process. Never said that the current docs help us in deployment process. Since, a separate file does not help us in the deployment process, why do we need to add an extra file and an indirection. Any change should have a net positive impact in the maintenance and developer experience.

docs/installation/local.md
iamareebjamal: It's not a not good way of doing it. It's not a bad practice. Almost all tools are compatible with it and none discourage it. Can you show any python doc reference where they discourage to keep virtual environments inside the repo? Even our CI systems like Travis and CircleCI cache the dependencies because they are inside the repo and it is easier to find and make changes in it as well. Keeping it outside the repo is one of the ways of doing and a personal preference. Each way has pro and cons. Keeping it inside can lead to extraneous search results, keeping it outside can make it much more difficult to manage and cache, etx

docs/installation/local.md
hongquan: @iamareebjamal 

> debian dependencies don't help us in deployment process

This PR doesn't interfere the deployment process. It is addressing the "local installation guide". Before this PR, the documentation is guiding beginner to setup development environment on Ubuntu/Debian, and this PR still focus on it.
If you want to improve deployment process, let's do it in separate PR.

hongquan: > add an extra file and an indirection

That extra file to help developer install dependencies in one command, and also, if the dependencies list grows overtime, it just make that file row instead of this document. If you concern about list of packages stay in separate file, why don't you concern the _requirements.txt_, which is also a separate file to list dependencies?

docs/installation/local.md
hongquan: @iamareebjamal 

You keep make mistake between this "local installation guide" and the deployment script. Do the TravisCI, CircleCI tool read this documentation and follow this documentation? Are they human?

> Can you show any python doc reference where they discourage to keep virtual environments inside the repo

Those projects assume that the readers has enough Python knowledge to manage the virtual environment themselves, to resolve issues themselves. They don't tell users to keep virtual environment inside or outside and let it to be personal choice.

The guide I add here is a "tip" for some beginner who may run into problem. Experienced users can skip this, like me, when read a Python project REAME and the author tells using Pipevn, I just skip it.

I mentored some students before, and saw some cases when the students run testcases, the test runner failed because it accidentally scan the virtualenv folders, trying to discovering test cases there, running some code in it and got side effect. So I have to write this, to let other beginners not run into the same problem.

iamareebjamal: We need to keep the installation as simple as possible. We can have your recommendation as an alternative but not as the default command. There is no simple way to manage virtual environments outside the repo. 

> I mentored some students before, and saw some cases when the students ran testcases, the test runner failed because it accidentally scan the virtualenv folders, trying to discovering test cases there, running some code in it and got side effect.

Our testrunner is configured to not fail this. I have also mentored a lot of students for running this project and 80% of the problem boils down to them missing a step in the guide. Some don't start with the correct python version, and some don't initialize a virtual environment. Adding this step as a recommended one will only confuse users and they'll definitely be stuck in problems. We need to keep documentation exact and consistent. I have no problem with adding the recommendation, but I can't approve or merge the PR with it being the default method without it being easy to manage. When people tell me they're having an issue installing things, I ask them to `source .venv/bin/activate`. If they have more issues, I tell them to `rm -rf .venv` and repeated. Your recommendation will make it difficult for me to guide students on how to setup things because I will need to ask "Where is your virtual environment?" and then change my commands according to their environment every time. Many people come from JS background, they already have an idea of a per-project dependency environment like `node_modules`.

> The guide I add here is a "tip" for some beginner who may run into problem. Experienced users can skip this, like me, when reading a Python project README and the author tells using Pipevn, I just skip it.

And this is what I am saying, an expert will probably use your method. A beginner will most definitely be confused about what to do and fail in the next step. Pragmatism and practicality beats perfectionism. When we switch to poetry, we can add the section where the entire virtual environment is managed by poetry. Not now.

docs/installation/local.md
iamareebjamal: This is when they have created a different branch. Your assumption of people starting contribution is that they understand best practices of git. Most people beginning here make PRs from the development branch. If they run fetch merge (which is basically pull), then it'll be like dev-commit -> their-commit -> dev-latest-commit -> merge-commit

Which we don't want. Now their local development branch contains interleaved unnecessary commits. 

Also, we want people to rebase their local branch with upstream branch across works, not merge. Merges are disastrous with database migrations. We want their commits to look like:

dev1 - dev2 - dev3 - my1 - my2 - my3

With merge, it'll look like

dev1 - my1 - m2 - dev2 - my3 - dev3

Which we don't want and can also cause wrong migration sequences. It also makes it difficult to squash commits as their commits are not grouped together. And checking out a commit will give them wrong state of the project. Checking out on my3 will include changes from dev2 but not dev3 which is already merged in the backend

docs/installation/basic.md
```diff
@@ -3,10 +3,12 @@
...
 * Python 3.7
-* Postgres
+* PostgreSQL
+* Redis
+
+
 ```sh
-sudo apt-get update
-sudo apt-get install postgresql postgresql-contrib
+xargs -a deb-packages.txt sudo apt install
```
iamareebjamal, line 11: The repo is not cloned at this point. People will copy paste the command and be greeted with an error

docs/installation/local.md
```diff
@@ -2,169 +2,183 @@
...
-To setup a virtual environment and install the dependices, enter in a terminal
+  ```sh
+  python3 -m venv .venv
+  ```
 
-```sh
-pipenv --python 3.7.3 install
-```
+  Running this command will create a folder named *.venv* in your working copy. Please remember
+  not to commit it.
```
iamareebjamal, line 77: Not necessary. It is already .gitignored

docs/installation/local.md
```diff
@@ -2,169 +2,183 @@
...
+
+- When you do somethings that involve scanning source code (like running test cases, search for file content), you accidentally scan the Python packages in the virtual environment.
+- You don't know how many virtual environments you created.
+- You have to know the path of an virtual environment in order to enter (if you are not standing in project folder).
+
+After creating virtual environment, you should upgrade `pip`, `wheel` libraries inside it.
 
 ```sh
-pipenv shell
+pip install -U pip wheel setuptools
```
iamareebjamal, line 101: This is already installed in the virtualenv by default

docs/installation/local.md
```diff
@@ -2,169 +2,183 @@
...
 ```
 
+## Install Python packages
+
+- Activate the Python virtual environment you created for this project.
```
iamareebjamal, line 106: Please add the `source .venv/bin/activate` command. Beginners don't know this and will run the next command in root python shell

iamareebjamal:
Other that the pull rebase command, other changes LGTM

docs/installation/local.md
hongquan: Of course it is installed in the virtualenv by default. But please notice the swtich `-U`. It means "upgrade". I see many cases that the existing "pip", "setuptools" aren't new enough and make later `pip install` fail.

docs/installation/local.md
hongquan: I agree to make change as you suggest, to let communicate with beginner easier.

docs/installation/basic.md
hongquan: Then I move this command down to after "cloning".

hongquan: This part is already after "cloning".

docs/installation/local.md
hongquan: Ok. Because you guys are the ones who review the PR, I should follow the approach that you are familiar with.

docs/installation/local.md
```diff
@@ -2,169 +2,165 @@
...
 
-* **Step 5** - Create the tables. For that we will use `create_db.py`.
+Please run these inside Python virtual environment
 
 ```sh
 python3 create_db.py
 # enter email and password
 python3 manage.py db stamp head
 ```
-**Note 1:** In case you made your own username and password in Step 2 are now getting `FATAL:  password authentication failed for user "john"` , probable cause is non updation of `.env` file. To resolve it, open the `.env` file and update `DATABASE_URL=postgresql://USERNAME:PASSWORD@127.0.0.1:5432/oevent` and you are good to go.
```
iamareebjamal, line 153: This note is important. Since our .env file now has ident method of authentication, many people  on Mac will encounter this error because they will have created the database using username and password. So this should be kept

iamareebjamal (approved):

PR #7392, opened 2020-10-30
fix: prevent deletion of ticket that has sales

maze-runnar:
Fixes  https://github.com/fossasia/open-event-frontend/issues/5195

#### Checklist

### DIFF
--- a/app/api/tickets.py
+++ b/app/api/tickets.py
@@ -5,7 +5,7 @@ from sqlalchemy.orm.exc import NoResultFound
 
 from app.api.bootstrap import api
 from app.api.helpers.db import get_count, safe_query_kwargs
-from app.api.helpers.errors import ConflictError, UnprocessableEntityError, ForbiddenError
+from app.api.helpers.errors import ConflictError, ForbiddenError, UnprocessableEntityError
 from app.api.helpers.permission_manager import has_access, is_logged_in
 from app.api.helpers.query import event_query
 from app.api.helpers.utilities import require_relationship
@@ -240,12 +240,12 @@ class TicketDetail(ResourceDetail):
                     {'event_id': ticket.event.id},
                     "Event having paid ticket must have a payment method",
                 )
-        else:
-            if (data.get('deleted_at') and ticket.has_current_orders):
-                raise ForbiddenError(
-                    {'source': ''},
-                    "Can't delete a ticket that has sales",
-                )
+        
+        if (ticket.has_current_orders and data.get('deleted_at')):
+            raise ForbiddenError(
+                {'source': ''},
+                "Can't delete a ticket that has sales",
+            )
 
     decorators = (
         api.has_permission(
--- a/app/api/tickets.py
+++ b/app/api/tickets.py
@@ -243,7 +243,7 @@ class TicketDetail(ResourceDetail):
         
         if (data.get('deleted_at') and ticket.has_current_orders):
             raise ForbiddenError(
-                {'source': ''},
+                {'param': 'ticket_id'},
                 "Can't delete a ticket that has sales",
             )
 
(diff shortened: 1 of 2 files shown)


app/api/tickets.py
```diff
@@ -241,6 +241,12 @@ def before_update_object(self, ticket, data, view_kwargs):
                     {'event_id': ticket.event.id},
                     "Event having paid ticket must have a payment method",
                 )
+        else:
+            if (data.get('deleted_at') and ticket.has_order_tickets):
```
iamareebjamal, line 245: It should only check completed, placed, pending and initializing tickets

iamareebjamal: expired, cancelled orders may prevent deletion this way which will create issues like this https://github.com/fossasia/open-event-frontend/issues/5430

iamareebjamal: Create a new function as this will not just check if the ticket has orders. Name it `has_current orders` and use 
```python
Order.query.join(TicketHolder)
    .filter(TicketHolder.ticket_id == self.id, 
        or_(Order.status == 'completed',
            Order.status == 'placed',
            Order.status == 'pending',
            Order.status == 'initializing'))
```

Don't fetch all the orders like done above. Neither use `get_count`. Use `exists` query like here

https://github.com/fossasia/open-event-server/blob/d3582cbae6b65044baf9178c53f0885bfa11623b/app/models/video_stream.py#L41-L46

iamareebjamal:
@maze-runnar Status?

maze-runnar:
@iamareebjamal it is giving this error on deletion of ticket - 
```
 if (data.get('deleted_at') and ticket.has_current_orders):
    raise TypeError("Boolean value of this clause is not defined")
```

app/models/ticket.py
```diff
@@ -103,7 +102,15 @@ def has_completed_order_tickets(self):
...
         return bool(count > 0)
-
+    
+    @property
+    def has_current_orders(self):
+        return Order.query.join(TicketHolder).filter(TicketHolder.ticket_id == self.id, 
+                or_(Order.status == 'completed',
+                Order.status == 'placed',
+                Order.status == 'pending',
+                Order.status == 'initializing')).exists()
```
iamareebjamal, line 114: No, it has something missing from the query I wrote

iamareebjamal:
@maze-runnar That's because you didn't write what I commented. You can't just skip things randomly and expect everything to work

app/models/ticket.py
iamareebjamal: Where is `db.session.query(...).exists()`

app/api/tickets.py
```diff
@@ -240,12 +240,12 @@ def before_update_object(self, ticket, data, view_kwargs):
...
                     "Event having paid ticket must have a payment method",
                 )
-        else:
-            if (data.get('deleted_at') and ticket.has_current_orders):
-                raise ForbiddenError(
-                    {'source': ''},
-                    "Can't delete a ticket that has sales",
-                )
+        
+        if (ticket.has_current_orders and data.get('deleted_at')):
```
iamareebjamal, line 244: deleted_at should be checked first. There is no need to make an SQL query everytime there is a change in tickets. It only needs to be checked if ticket is being deleted

iamareebjamal (dismissed):

app/api/tickets.py
```diff
@@ -240,6 +240,12 @@ def before_update_object(self, ticket, data, view_kwargs):
                     {'event_id': ticket.event.id},
                     "Event having paid ticket must have a payment method",
                 )
+        
+        if (data.get('deleted_at') and ticket.has_current_orders):
+            raise ForbiddenError(
+                {'source': ''},
```
iamareebjamal, line 246:
```suggestion
                {'param': 'ticket_id'},
```

PR #7418, opened 2020-11-06
chore(ref): Refactor order complete logic

### LINKED ISSUE #7412: Fix order complete and email sending logic
The order complete logic is duplicated in 5 places and really really messed up. It is impossible to figure out what'll happen when with subtle differences. This is the reason why attachments to users were not working in the first place.

Fix and deduplicate the logic once and for all

iamareebjamal:
Fixes #7412 

Organizer will now receive emails about ticket purchase


mariobehling:
Nice!

PR #7443, opened 2020-11-13
fix: if ticket is not available for sale ,don't show event

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/5110

#### Checklist

### DIFF
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -842,6 +842,7 @@ class UpcomingEventList(EventList):
         :param view_kwargs:
         :return:
         """
+        from app.api.attendees import get_sold_and_reserved_tickets_count
         current_time = datetime.now(pytz.utc)
         query_ = (
             self.session.query(Event)
@@ -858,7 +859,7 @@ class UpcomingEventList(EventList):
                         Event.event_type_id != None,
                         Event.event_topic_id != None,
                         Event.event_sub_topic_id != None,
-                        Event.tickets.any(and_(Ticket.deleted_at == None, Ticket.is_hidden == False, Ticket.sales_ends_at > current_time, Ticket.is_available == True)),
+                        Event.tickets.any(and_(Ticket.deleted_at == None, Ticket.is_hidden == False, Ticket.sales_ends_at > current_time, Ticket.quantity > get_sold_and_reserved_tickets_count(Ticket.id))),
                     ),
                 ),
             )
--- a/app/api/events.py
+++ b/app/api/events.py
@@ -858,7 +858,7 @@ class UpcomingEventList(EventList):
                         Event.event_type_id != None,
                         Event.event_topic_id != None,
                         Event.event_sub_topic_id != None,
-                        Event.tickets.any(and_(Ticket.deleted_at == None, Ticket.is_hidden == False, Ticket.sales_ends_at > current_time)),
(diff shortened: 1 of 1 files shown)


app/api/events.py
```diff
@@ -858,7 +858,7 @@ def query(self, view_kwargs):
                         Event.event_type_id != None,
                         Event.event_topic_id != None,
                         Event.event_sub_topic_id != None,
-                        Event.tickets.any(and_(Ticket.deleted_at == None, Ticket.is_hidden == False, Ticket.sales_ends_at > current_time)),
+                        Event.tickets.any(and_(Ticket.deleted_at == None, Ticket.is_hidden == False, Ticket.sales_ends_at > current_time, Ticket.is_available == True)),
```
iamareebjamal, line 861: is_available is not a column. It's a computed property. It doesn't work like this

maze-runnar: the query that is written in `is_available` property is that right for the condition?can that query be written here again for this work?

iamareebjamal: Yes

iamareebjamal: But please run and confirm before committing. Just visit /v1/events/upcoming to verify

app/api/events.py
```diff
@@ -858,7 +859,7 @@ def query(self, view_kwargs):
                         Event.event_type_id != None,
                         Event.event_topic_id != None,
                         Event.event_sub_topic_id != None,
-                        Event.tickets.any(and_(Ticket.deleted_at == None, Ticket.is_hidden == False, Ticket.sales_ends_at > current_time, Ticket.is_available == True)),
+                        Event.tickets.any(and_(Ticket.deleted_at == None, Ticket.is_hidden == False, Ticket.sales_ends_at > current_time, Ticket.quantity > get_sold_and_reserved_tickets_count(Ticket.id))),
```
iamareebjamal, line 862: Won't work. This is not a python loop, only allowed operations here are sqlalchemy scalar and join type operations

iamareebjamal: Please verify by running before submitting

maze-runnar: it's working, I have checked it

maze-runnar: I have an event, it has two tickets and 1 ticket of each type is sold.then after applying above I edited event and changes ticket number to 1 of both events then the event is not on the front page anymore. Then i changed ticket numbers for one ticket to one and for another ticket to 100. Then event is showing on front page.

iamareebjamal: Hmm, yes, it'll work indeed. But I'll have to check if the query is efficient. Sorry for not noticing `Ticket.id` instead of `ticket.id`

iamareebjamal: Doesn't work, it statically returns all reserved and sold tickets and compares ticket quantity with it

maze-runnar: `get_sold_and_reserved_tickets_count(Ticket.id)` what is meaning of passing `ticked.id` here?

maze-runnar: `all reserved and sold tickets ` for a ticket ?

iamareebjamal: It won't work because it expects a ticket ID. Passing `Ticket.id` just returns all tickets sold count

PR #7456, opened 2020-11-18
wip: Cache control

iamareebjamal:
Microcache scenario:

```
If cache query string present or unauthenticated user:
  cache key - normal
else
  cache key - JWT + normal
```


maze-runnar:
I am closing it for now, if any one can complete it  feel free to reopen and work.

PR #7465, opened 2020-11-25
fix: updating condition to not show empty session in user section

maze-runnar:
…hout email

Fixes https://github.com/fossasia/open-event-frontend/issues/5342

#### Checklist

### DIFF
--- a/app/api/speakers.py
+++ b/app/api/speakers.py
@@ -190,7 +190,7 @@ class SpeakerDetail(ResourceDetail):
                 'Organizer access required to override email',
             )
         if (
-            data.get('is_email_overridden')
+            not data.get('is_email_overridden')
             and has_access('is_organizer', event_id=speaker.event_id)
             and not data.get('email')
         ):


app/api/speakers.py
```diff
@@ -190,7 +190,7 @@ def before_update_object(self, speaker, data, view_kwargs):
                 'Organizer access required to override email',
             )
         if (
-            data.get('is_email_overridden')
+            not data.get('is_email_overridden')
```
iamareebjamal, line 193: When did this happen?

maze-runnar: if email is overriden then there will be no email so email is null m but here are condition that is saying that if email is overriden and email is null then set it to user mail, while it should be, while email is not overriden and email is null set it to user mail

iamareebjamal (approved):

app/api/speakers.py
iamareebjamal: @maze-runnar You changed this but didn't make email optional in schema

This is causing https://github.com/fossasia/open-event-frontend/issues/5841

maze-runnar: > @maze-runnar You changed this but didn't make email optional in schema
> 
> This is causing [fossasia/open-event-frontend#5841](https://github.com/fossasia/open-event-frontend/issues/5841)

@iamareebjamal that is related to PR https://github.com/fossasia/open-event-server/pull/7469, and I have made email optional but still showing that error.

PR #7542, opened 2020-12-28
chore: Use Poetry to manage Python dependencies

### LINKED ISSUE #7319: Use Poetry to manage dependency

Managing packages with requirements.txt is too simple. For example when we want to upgrade packages, we want the packages are upgrade with newest version but still stay compatible with each other. Using requirements.txt means that we have to manually choose and try version for each packages, which is tiresome.

Let's use [Poetry](https://python-poetry.org/).

HarryHuy:
Resolve #7319 

Minor changes to the way of manipulating Python dependencies

For development, let's use [Poetry](https://python-poetry.org/docs) to reduce the overhead with Pip. The packages are specifying in `pyproject.toml` and `poetry.lock` files. Herewith some basic usages:

* Create project's virtual environment and install dependencies:

```
[Project folder]$ poetry install
```

* Active virtual environment

```
[Project folder]$ poetry shell
```

* Export dependencies to Pip's requirements file format

```
[Project folder]$ poetry export --without-hashes -o requirements.txt
```

For deployment and testing, Pip is still being used along with two requirements files exported by Poetry: 

    `./requirements/prod.txt` includes packages for deployment only

    `./requirements/dev.txt` includes all packages for deployment, development, and testing

These changes have been updated to the Dockerfile, scripts, settings files, and documentation as well. However, this project offers many ways for deployment and testing, so I don't know if something is missed. Please help me to review it. Thanks!

#### Checklist


iamareebjamal:
Hi, thanks for your contribution

Unfortunately, we cannot split the dependencies to be installed with 2 workflows, because then it may duplicate the work and result in incompatible environments locally, v/s in production. We want to use 1 solution across the board, i.e., in development and in production. 
In fact, other possible workflow is to use poetry for production but install dev and test dependencies using pip. Then we would at least have poetry across the board as the defacto dependency manager in production. But not the other way around. 
Using poetry for just installing dev dependencies does not move us forward in adopting poetry in any way.

So, possible workflows:
1. Manage all dependencies via poetry
2. Manage production dependencies in Docker and CI using poetry and installing dev dependencies using pip

HarryHuy:
@iamareebjamal 
I think my description wasn't clear enough and made you confused.

Poetry is used to manage **all dependencies**, not only dev dependencies. However, we only need to install and utilize it in the **development environment**. In the **production environment**, we use Pip to install the dependencies specified in the requirements files exported by Poetry. This helps to keep the container as well as the deployment script simple.

The environment incompatible wouldn't happened in this way. When the dependencies are resolved and locked by a specific version, both Poetry and Pip will get them from the same sources (Pypi, Github... or wherever you defined) to install.

Poetry brings a single CLI to manage both virtual environment and dependencies. It's beneficial for developers especially newcomers to Python. What needed in the production environment is to install the right version of dependencies specified by the developer. So the option 2 is meaningless.

Anyway, I agree that it's better if everything is synced, so I can make this PR further to utilize Poetry for the deployment if you are willing to update.

Involve the issue's owner for the other idea @hongquan

iamareebjamal:
Yes, we need to use single source for deployments as well.

> In the production environment, we use Pip to install the dependencies specified in the requirements files exported by Poetry

As far as I understand, that requirements file needs to be manually exported every time there is a dependency update. Thus, we cannot claim that the dependency are synced. They are inherently unsynced by the nature of no coupling between poetry dependency list and the requirements file. This is a recipe for confusion and probable bugs in future where any developer may forget to generate the requirements file after updating the poetry list. And it also breaks our automated dependency upgrade system as it will not sync the requirements file automatically

HarryHuy:
Added Poetry to the Docker image, updated Composer and CircleCI config.

Have tested with virtual machine deployment. Not yet tested with Heroku however, there are announcements regarding Poetry support on their repository. Let see how it works. (https://github.com/heroku/heroku-buildpack-python/issues/796)

I think you guys haven't used Kubernetes anymore so skipped updating its setup.

HarryHuy:
@iamareebjamal plese review, thanks!

Dockerfile
```diff
@@ -4,24 +4,36 @@ FROM python:3.8-alpine as base
...
+ADD poetry.lock ./
+
+RUN poetry install --no-root --no-dev
 
 ####
 
 FROM base
 
-COPY --from=builder /install /usr/local
+COPY --from=builder /opt/pysetup/.venv /opt/pysetup/.venv
```
iamareebjamal, line 33: Can we move it to /usr/local like before. Philosophy of the final container is to be structured like it is in built with the required packages, so it's better for them to be in /usr/local

iamareebjamal: Also, please add poetry install directory in the final container and in the PATH so we are able to install any ad hoc package on running container

iamareebjamal:
LGTM but circle CI is not running for some reason

HarryHuy:
Thanks for covering for me @iamareebjamal 

> LGTM but circle CI is not running for some reason

Is it because I missed the `app.json` file?

iamareebjamal:
No, maybe because of closing reopening the PR. Now it's fixed

PR #7554, opened 2020-12-31
fix order status change

### LINKED ISSUE #6134: Client can create wrong status of an order

Tlazypanda:
Fixes #6134 

Fix order status transition to only allow transition in the possible cases. Assuming certain things, have allowed :
- Super admin to surpass all checks
- Admin or coorganizer to change states in the correct manner (allowing them to mark order as completed too)
- User to change states in correct manner (but not allowing them to mark order as completed)

#### Checklist


iamareebjamal:
What is it preventing?

Tlazypanda:
@iamareebjamal invalid state change and for user preventing state change to completed. I guess reading through the comments this is what I understood and made the changes accordingly :sweat_smile:

iamareebjamal:
Could you make such changes before? Please verify

PR #7557, opened 2021-01-01
feat: return dummy for deleted users

### LINKED ISSUE #7424: Return dummy user for deleted ones except for admin user endpoint
Organizers/Admins can access user info such as for attendees. In cases a user deletes its account, they can still view their name and email and such. Please return a dummy user for such cases except on the admin page which shows deleted users.

Dummy user's name can be 'Deleted User' and email 'deleted.account@eventyay'

Tlazypanda:
Fixes #7424 

Return dummy for deleted users by modifying the dump according to the permission level (super admin can see email and name) but others will get "deleted user" in response.

#### Checklist

### DIFF
--- a/app/api/schema/users.py
+++ b/app/api/schema/users.py
@@ -1,11 +1,12 @@
 from flask_jwt_extended import current_user
-from marshmallow import post_dump
+from marshmallow import pre_dump
 from marshmallow_jsonapi import fields
 from marshmallow_jsonapi.flask import Relationship
 
 from app.api.helpers.permission_manager import is_logged_in
 from app.api.helpers.utilities import dasherize
 from app.api.schema.base import SoftDeletionSchema
+from app.models.user import User
 from utils.common import use_defaults
 
 
@@ -276,16 +277,22 @@ class UserSchema(UserSchemaPublic):
         many=True,
     )
 
-    @post_dump
+    @pre_dump
     def handle_deleted_users(self, data):
+        user = None
         if (
             not (
                 is_logged_in
-                and (current_user.is_staff or str(current_user.id) == data['id'])
+                and current_user
+                and data
+                and (current_user.is_staff or current_user.id == data.id)
             )
-            and data['deleted_at'] != None
+            and data.deleted_at != None
         ):
-            data['email'] = 'deleted@eventyay.com'
-            data['first_name'] = 'deleted'
-            data['last_name'] = 'user'
-        return data
+            user = User()
+            user.email = 'deleted@eventyay.com'
+            user.first_name = 'deleted'
+            user.last_name = 'user'
+            user.deleted_at = data.deleted_at
+            user.id = data.id
+        return user if user else data
--- a/app/api/schema/users.py
+++ b/app/api/schema/users.py
@@ -1,6 +1,9 @@
+from flask_jwt_extended import current_user
+from marshmallow import post_dump
 from marshmallow_jsonapi import fields
 from marshmallow_jsonapi.flask import Relationship
 
+from app.api.helpers.permission_manager import is_logged_in
 from app.api.helpers.utilities import dasherize
 from app.api.schema.base import SoftDeletionSchema
 from utils.common import use_defaults
@@ -272,3 +275,17 @@ class UserSchema(UserSchemaPublic):
         type_='event',
         many=True,
     )
+
+    @post_dump
+    def handle_deleted_users(self, data):
+        if (
+            not (
+                is_logged_in
+                and (current_user.is_staff or str(current_user.id) == data['id'])
+            )
+            and data['deleted_at'] != None
+        ):
+            data['email'] = 'deleted@eventyay.com'
+            data['first_name'] = 'deleted'
+            data['last_name'] = 'user'
+        return data
--- a/app/api/schema/users.py
+++ b/app/api/schema/users.py
@@ -279,20 +279,15 @@ class UserSchema(UserSchemaPublic):
 
     @pre_dump
     def handle_deleted_users(self, data):
-        user = None
-        if (
-            not (
-                is_logged_in
-                and current_user
-                and data
-                and (current_user.is_staff or current_user.id == data.id)
-            )
-            and data.deleted_at != None
+        if not data:
+            return data
+        if data.deleted_at != None and not (
+            is_logged_in
+            and current_user
+            and (current_user.is_staff or current_user.id == data.id)
         ):
-            user = User()
-            user.email = 'deleted@eventyay.com'
-            user.first_name = 'deleted'
-            user.last_name = 'user'
-            user.deleted_at = data.deleted_at
-            user.id = data.id
-        return user if user else data
+            user = User(
+                id=0, email='deleted@eventyay.com', first_name='deleted', last_name='user'
+            )
+            return user
+        return data
--- a/tests/all/integration/api/attendee/test_attendee_access_api.py
+++ b/tests/all/integration/api/attendee/test_attendee_access_api.py
@@ -1,12 +1,9 @@
 import json
 
-from flask_jwt_extended.utils import create_access_token
-
 from app.api.helpers.db import get_or_create
 from app.models.role import Role
 from app.models.users_events_role import UsersEventsRoles
 from tests.factories.attendee import AttendeeOrderTicketSubFactory
-from tests.factories.user import UserFactory
 
 
 def get_minimal_attendee(db, user=None, owner=False, event_status='published'):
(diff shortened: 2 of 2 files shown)


Tlazypanda:
@iamareebjamal can you kindly review this? :sweat_smile: Thanks :v:

app/api/schema/users.py
```diff
@@ -272,3 +274,11 @@ class Meta:
         type_='event',
         many=True,
     )
+
+    @post_dump
+    def handle_deleted_users(self, data):
+        if (not has_access('is_super_admin')) and data['deleted_at'] != None:
```
iamareebjamal, line 280: Use is_logged_in and current_user.is_staff

app/api/schema/users.py
```diff
@@ -272,3 +274,11 @@ class Meta:
...
+            data['email'] = 'deleted@eventyay.com'
+            data['first_name'] = 'deleted'
+            data['last_name'] = 'user'
+        return data
```
iamareebjamal, line 284: This method can still leak other data. Besides, if the user itself is accessing this data, it should be allowed

Tlazypanda:
Hey @iamareebjamal Have updated the mr with other suggestions but not sure how to deal with the data leakage issue? :sweat_smile: can you guide me on this? I tried searching for the serliaize method in marshmallow docs but instead found @post_dump and other methods used to manipulate data. Thanks :v:

iamareebjamal:
Is there any pre-dump? Just replacing name and email is not an option. It leaks ID and other details in the model

app/api/schema/users.py
```diff
@@ -272,3 +275,19 @@ class Meta:
...
+                is_logged_in
+                and current_user
+                and data
+                and (current_user.is_staff or current_user.id == data.id)
+            )
+            and data.deleted_at != None
+        ):
+            data.email = 'deleted@eventyay.com'
+            data.first_name = 'deleted'
+            data.last_name = 'user'
```
iamareebjamal, line 292: Please return a new user instance, not modify the data

Tlazypanda: @iamareebjamal so should I copy data into the new object and modify the fields then? something like
`user = data.copy()
            user.email = 'deleted@eventyay.com'
            user.first_name = 'deleted'
            user.last_name = 'user'`

iamareebjamal: Just create a new object of User model

Tlazypanda: Ohh I finally understood what you meant by data leak previously we just have to return these attributes and not copy the other fields :sweat_smile: :facepalm:

Tlazypanda: @iamareebjamal we aren't saving this instance to the db right? I am getting an error to set the id for this user since user model has function _is_role with line `uer = UER.query.filter_by(user=self, role=role).first()` which is trying to find user.id column

app/api/schema/users.py
```diff
@@ -272,3 +276,23 @@ class Meta:
...
+            )
+            and data.deleted_at != None
+        ):
+            user = User()
+            user.email = 'deleted@eventyay.com'
+            user.first_name = 'deleted'
+            user.last_name = 'user'
+            user.deleted_at = data.deleted_at
+            user.id = data.id
+        return user if user else data
```
iamareebjamal, line 298: return data

app/api/schema/users.py
```diff
@@ -272,3 +276,23 @@ class Meta:
...
+                and (current_user.is_staff or current_user.id == data.id)
+            )
+            and data.deleted_at != None
+        ):
+            user = User()
+            user.email = 'deleted@eventyay.com'
+            user.first_name = 'deleted'
+            user.last_name = 'user'
+            user.deleted_at = data.deleted_at
+            user.id = data.id
```
iamareebjamal, line 297:
```suggestion
            user.id = 0
            return user
```

app/api/schema/users.py
```diff
@@ -272,3 +276,23 @@ class Meta:
...
+                and current_user
+                and data
+                and (current_user.is_staff or current_user.id == data.id)
+            )
+            and data.deleted_at != None
+        ):
+            user = User()
+            user.email = 'deleted@eventyay.com'
+            user.first_name = 'deleted'
+            user.last_name = 'user'
```
iamareebjamal, line 295: All of this can be passed in constructor

app/api/schema/users.py
```diff
@@ -272,3 +276,23 @@ class Meta:
...
+            user.deleted_at = data.deleted_at
```
iamareebjamal, line 296: Why? We don't need to leak information

app/api/schema/users.py
```diff
@@ -272,3 +276,23 @@ class Meta:
         type_='event',
         many=True,
     )
+
+    @pre_dump
+    def handle_deleted_users(self, data):
+        user = None
```
iamareebjamal, line 282: Not needed

app/api/schema/users.py
```diff
@@ -272,3 +276,23 @@ class Meta:
...
+        if (
+            not (
+                is_logged_in
+                and current_user
+                and data
+                and (current_user.is_staff or current_user.id == data.id)
+            )
+            and data.deleted_at != None
```
iamareebjamal, line 290: This should be the first check

app/api/schema/users.py
iamareebjamal, line 282:
```suggestion
```

app/api/schema/users.py
iamareebjamal, line 298:
```suggestion
        return data
```

app/api/schema/users.py
iamareebjamal, line 296:
```suggestion
```

app/api/schema/users.py
iamareebjamal, line 297: Don't want to leak ID

app/api/schema/users.py
```diff
@@ -272,3 +276,23 @@ class Meta:
...
     )
+
+    @pre_dump
+    def handle_deleted_users(self, data):
+        user = None
+        if (
+            not (
+                is_logged_in
+                and current_user
+                and data
```
iamareebjamal, line 287: Add `if not data: return data` on first line

Tlazypanda:
@iamareebjamal updated mr! Thanks for pointing out all the suggestions :v: Will work on becoming better :raised_hands:

app/api/schema/users.py
```diff
@@ -272,3 +276,20 @@ class Meta:
...
+        if data.deleted_at != None and not (
+            is_logged_in
+            and current_user
+            and data
+            and (current_user.is_staff or current_user.id == data.id)
+        ):
+            user = User(
+                email='deleted@eventyay.com', first_name='deleted', last_name='user'
+            )
+            user.id = 0
```
iamareebjamal, line 293: Can id be passed in constructor as well

app/api/schema/users.py
```diff
@@ -272,3 +276,20 @@ class Meta:
...
     )
+
+    @pre_dump
+    def handle_deleted_users(self, data):
+        if not data:
+            return data
+        if data.deleted_at != None and not (
+            is_logged_in
+            and current_user
+            and data
```
iamareebjamal, line 287: Now this should not be needed

iamareebjamal:
Great work!

As now we are moving at a higher pace with server, it is important to verify changes with a test. Can you please write 2 tests, one with deleted user and one with non deleted user. And verify that appropriate data is returned in both cases?

You can check tests/integration/api files to get a reference

Tlazypanda:
Yeah sure! On it  :100:  @iamareebjamal

Tlazypanda:
@iamareebjamal for the integration tests in the main issue which endpoint are we hitting to get the deleted records? I tried the relationship between attendee and user but if a record is deleted it's not accessible ... same goes for the normal user detail ... I also tried the attendee detail one but in that first name and last name attributes are different from that of user :sweat_smile:

iamareebjamal:
Did you try `/event/ID/attendees?include=user`?

Tlazypanda:
The above url wasn't working so I tried with `/v1/attendees/{attendee.id}?include=user'`  but here too in case of deleted entries it doesn't fetch the included part. I tried with get_trashed param on /users/id route too but that only works for admin so I can't test for organizer @iamareebjamal :sweat_smile:

iamareebjamal:
OK, then please create a test for undeleted user so we have verification that changes don't crash the server

tests/all/integration/api/attendee/test_attendee_access_api.py
```diff
@@ -18,6 +21,68 @@ def get_minimal_attendee(db, user=None, owner=False, event_status='published'):
...
+    assert (
+        json.loads(response.data)['included'][0]['attributes']['first-name']
+        == user.first_name
+    )
+    assert (
+        json.loads(response.data)['included'][0]['attributes']['last-name']
+        == user.last_name
+    )
+
+    response = client.get(
```
iamareebjamal, line 68: I don't understand. Why are there three same calls and same assertions

tests/all/integration/api/attendee/test_attendee_access_api.py
```diff
@@ -1,9 +1,12 @@
 import json
 
+from flask_jwt_extended.utils import create_access_token
+
```
iamareebjamal, line 4:
```suggestion
```

tests/all/integration/api/attendee/test_attendee_access_api.py
```diff
@@ -1,9 +1,12 @@
...
 from app.api.helpers.db import get_or_create
 from app.models.role import Role
 from app.models.users_events_role import UsersEventsRoles
 from tests.factories.attendee import AttendeeOrderTicketSubFactory
+from tests.factories.user import UserFactory
```
iamareebjamal, line 9:
```suggestion
```

tests/all/integration/api/attendee/test_attendee_access_api.py
```diff
@@ -18,6 +21,68 @@ def get_minimal_attendee(db, user=None, owner=False, event_status='published'):
...
+    assert (
+        json.loads(response.data)['included'][0]['attributes']['first-name']
+        == user.first_name
+    )
+    assert (
+        json.loads(response.data)['included'][0]['attributes']['last-name']
+        == user.last_name
+    )
+
+
```
iamareebjamal, line 85:
```suggestion
```

PR #7563, opened 2021-01-03
feat: Video Room: Sanitization added

MeghalBisht:
Fixes fossasia/open-event-frontend#5909

Sanitization added for "Additional Information"  field in Video Room

#### Checklist

### DIFF
--- a/app/models/video_stream.py
+++ b/app/models/video_stream.py
@@ -4,6 +4,7 @@ from sqlalchemy.orm import backref
 
 from app.api.helpers.permission_manager import has_access
 from app.models import db
+from app.models.helpers.versioning import clean_html, clean_up_string
 from app.models.order import Order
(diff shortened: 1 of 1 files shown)


iamareebjamal:
LGTM but I was thinking to finally switch to markdown instead of this, because

1. Converting to HTML loses original text
2. You have to display HTML even though there is no rich text, which sucks
3. You have to change everywhere how the text is displayed, like in previews in table, it'll show paragraph tags otherwise

Thus leaving this open for some time

MeghalBisht:
@iamareebjamal For the user, makrdown will be difficult to work with, as they might not understand what are # * etc doing and maybe would prefer WYSIWYG instead?

iamareebjamal:
HTML is much more difficult than markdown. We have a WYSIWIG editor for that as well

PR #7578, opened 2021-01-06
feat: create date route for events

sachinchauhan2889:
create dates routes for getting distinct dates of an event.


sachinchauhan2889:
screenshot

this code is working fine.  according to my events i created 3 sessions. 2 on same date and third on next day. 
it is returning two dates but these are empty.

@iamareebjamal sir, i spended a huge time on this. Gone through many documentation.
I used cast and func.date both. both are not giving expected result.
while exploring i got your issue https://github.com/sqlalchemy/sqlalchemy/issues/4922. and use that but now it is not working.

sachinchauhan2889:
screenshoted 2

iamareebjamal:
```python
list(zip(*db.session.query(func.date(Session.starts_at)).distinct().filter(Session.event_id==1078, Session.starts_at != None).all()))[0]
```

@sachinchauhan2889

sachinchauhan2889:
new commit

iamareebjamal:
Use str to convert the dates to strings

sachinchauhan2889:
new commit today

iamareebjamal:
Cool, as you are exploring the server a bit, please also create an integration test. Ask for any event if you need it.

Create and event in the test,

Create 3 sessions 2 on 1 date and 1 on a date skipping 2 days from others

Thus, it will test that the API only returns distinct dates and skips the unscheduled dates

app/api/custom/events.py
```diff
@@ -1,24 +1,30 @@
...
+                    .filter(
+                        Session.event_id == event_id,
+                        Session.starts_at != None,
+                        or_(Session.state == 'accepted', Session.state == 'confirmed'),
+                    )
+                    .order_by(asc(func.date(Session.starts_at)))
+                    .all()
+                )
+            )[0],
+        )
```
iamareebjamal, line 28: Too much nesting, it was better before

```python
event=Event.query.filter_by(identifier=event_identifier).first()
dates = (
        list(map(str, list(zip(
            *db.session.query(func.date(Session.starts_at))
            .distinct()
            .filter(Session.event_id==event.id, Session.starts_at != None, or_(Session.state == 'accepted', Session.state == 'confirmed'))
            .order_by(asc(func.date(Session.starts_at)))
            .all()
        ))[0]))
)
```

iamareebjamal:
Good work with tests

app/api/custom/events.py
sachinchauhan2889: Pre commit changed this. whenever i try to commit pre-commit changes my code to this.

iamareebjamal: But you were able to commit this before. 

If you format your code like this and break any line longer than 120 chars, it will not change it

sachinchauhan2889: > But you were able to commit this before. 
> 
> If you format your code like this and break any line longer than 120 chars, it will not change it



Previously i commited on github directly. I will again do that.

sachinchauhan2889:
> Good work with tests

 Thank you sir.

PR #7588, opened 2021-01-09
feat: Admin Provide Option "Not to show on Frontpage"

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/4963

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] The unit tests pass locally with my changes 
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

- [ ] All the functions created/modified in this PR contain relevant docstrings.

PR #7589, opened 2021-01-09
fix: adding demoted, migration, API change for upcoming event

maze-runnar:
Fixes fossasia/open-event-frontend#4963
continuing https://github.com/fossasia/open-event-server/pull/7588

#### Checklist


migrations/versions/rev-2021-01-09-18:31:24-cd8488c19b16_.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'cd8488c19b16'
+down_revision = 'fb8dd6aed9a9'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('events', sa.Column('is_demoted', sa.Boolean(), nullable=False))
```
iamareebjamal, line 21: Where is server_default?

This migration will fail. Please test the changes carefully

migrations/versions/rev-2021-01-09-18:31:24-cd8488c19b16_.py
```diff
@@ -0,0 +1,32 @@
...
+    op.add_column('events_version', sa.Column('is_demoted', sa.Boolean(), autoincrement=False, nullable=True))
+    op.execute("UPDATE events SET is_demoted=False")
```
iamareebjamal, line 23: This won't work, the first command itself will fail

migrations/versions/rev-2021-01-09-18:31:24-cd8488c19b16_.py
```diff
@@ -0,0 +1,32 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('events_version', 'is_demoted')
+    op.drop_column('events', 'is_demoted')
+    op.execute("UPDATE events SET is_demoted=True")
```
iamareebjamal, line 31: If you drop the column is_demoted, how can you set it to True? And even if you could, would you make all events demoted? Now, no event is being shown on frontend, WOW

Please be careful and think about these things :/

maze-runnar: I thought about it. But you have said before once that downgrade should be opposite of upgrade 😕

migrations/versions/rev-2021-01-09-18:31:24-cd8488c19b16_.py
maze-runnar: > This migration will fail. Please test the changes carefully

But I have tested it with frontend, it's working. did server_default is not added itself in autogenerated migration?

iamareebjamal: No

migrations/versions/rev-2021-01-09-18:31:24-cd8488c19b16_.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'cd8488c19b16'
+down_revision = 'fb8dd6aed9a9'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('events', sa.Column('is_demoted', sa.Boolean(), nullable=False, server_default=False))
```
iamareebjamal, line 21:
```suggestion
    op.add_column('events', sa.Column('is_demoted', sa.Boolean(), nullable=False, server_default='False'))
```

server_default needs to be quoted

migrations/versions/rev-2021-01-09-18:31:24-cd8488c19b16_.py
```diff
@@ -0,0 +1,31 @@
...
+    op.add_column('events_version', sa.Column('is_demoted', sa.Boolean(), autoincrement=False, nullable=True, server_default=False))
+    op.execute("UPDATE events SET is_demoted=False")
```
iamareebjamal, line 23:
```suggestion
```

iamareebjamal:
> But you have said before once that downgrade should be opposite of upgrade

Then sorry, correct definition is downgrade should restore the DB structure and state to where it was before migration. 

So, you drop the column, it is already restored
Once you dropped the column, it is not there so how can you set demoted to true?

Secondly, let's say you set default demoted to false, it makes sense, would you restore demoted to true on downgrade? It doesn't make sense, it'll make all events demoted. All events were never demoted, so demoted was never = true, so it is not restoring to correct state

In case of making custom forms public

Original State:
All forms are private by default, maybe some custom fields are made public by users

Migration:
Make all forms public - WRONG

Loss of state

Correct Migration:
Make speaker fields x, y, z public
Make session fields x, y, z public

Downgrade:
Make all fields private - WRONG, orignal state may have some public fields which users made

Data Loss

Thus, only make those fields private which you made public to restore to the original state.

Make speaker fields x, y, z private
Make session fields x, y, z private

Development is not just taking an idea that you heard previously and applying it without thought.
The reason of the idea is important. Goal is to revert to the original state of database, not just write opposite commands of the migration

migrations/versions/rev-2021-01-09-18:31:24-cd8488c19b16_.py
```diff
@@ -18,9 +18,8 @@
 
 def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
-    op.add_column('events', sa.Column('is_demoted', sa.Boolean(), nullable=False, server_default=False))
-    op.add_column('events_version', sa.Column('is_demoted', sa.Boolean(), autoincrement=False, nullable=True, server_default=False))
-    op.execute("UPDATE events SET is_demoted=False")
+    op.add_column('events', sa.Column('is_demoted', sa.Boolean(), nullable=False, server_default='False'))
+    op.add_column('events_version', sa.Column('is_demoted', sa.Boolean(), autoincrement=False, nullable=True, server_default='False'))
```
iamareebjamal, line 22:
```suggestion
    op.add_column('events_version', sa.Column('is_demoted', sa.Boolean(), autoincrement=False, nullable=True))
```

PR #7596, opened 2021-01-13
fix: Implement Mark Accounts as Spam

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/issues/6325

Screenshot from 2021-01-13 14-03-39

#### Checklist


app/api/auth.py
```diff
@@ -58,6 +58,10 @@ def authenticate(allow_refresh_token=False, existing_identity=None):
     password = data.get('password')
     criterion = [username, password]
 
+    user = User.query.filter_by(_email=username).first()
+    if user.is_blocked:
+        return jsonify(error='Admin has marked this account as Spam'), 400
```
iamareebjamal, line 63: Wrong, this is leaking info about user being banned. I can now know if a user is banned or not by just knowing their email. Think more

app/api/schema/users.py
```diff
@@ -71,6 +71,7 @@ class Meta:
     is_user_moderator = fields.Boolean(dump_only=True)
     is_user_registrar = fields.Boolean(dump_only=True)
     is_verified = fields.Boolean()
+    is_blocked = fields.Boolean(dump_only=True)
```
iamareebjamal, line 74: This is dump_only, so how will it be editable by the admin?

migrations/versions/rev-2021-01-13-13:56:45-a3fb59fea6c2_.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'a3fb59fea6c2'
+down_revision = 'cd8488c19b16'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('users', sa.Column('is_blocked', sa.Boolean(), nullable=True, server_default='False'))
```
iamareebjamal, line 21: nullable False

app/api/auth.py
maze-runnar:
```
 user = User.query.filter_by(_email=username, password=password).first()
    if user and user.is_blocked:
        return jsonify(error='Admin has marked this account as Spam'), 400
```

is this ok? it will require username and password both.

iamareebjamal: See the entire function, what you've written above is already being done somewhee

app/api/helpers/jwt.py
```diff
@@ -18,6 +18,8 @@ def jwt_authenticate(email, password):
     if user is None:
         return None
     auth_ok = user.facebook_login_hash == password or user.is_correct_password(password)
+    if auth_ok and user.is_blocked:
+        return "spam"
```
iamareebjamal, line 22: Revert

app/api/schema/users.py
```diff
@@ -169,15 +169,15 @@ class Meta:
...
-    )
+    # owner_events = Relationship(
+    #     self_view='v1.user_owner_events',
+    #     self_view_kwargs={'id': '<id>'},
+    #     related_view='v1.event_list',
+    #     related_view_kwargs={'user_owner_id': '<id>'},
+    #     schema='EventSchema',
+    #     many=True,
+    #     type_='event',
+    # )
```
iamareebjamal: revert

app/api/schema/users.py
```diff
@@ -260,22 +260,22 @@ class Meta:
...
+    #     many=True,
+    # )
+    # sales_admin_events = Relationship(
+    #     self_view='v1.user_sales_admin_events',
+    #     self_view_kwargs={'id': '<id>'},
+    #     related_view='v1.event_list',
+    #     schema='EventSchema',
+    #     type_='event',
+    #     many=True,
+    # )
```
iamareebjamal: revert

app/models/user.py
```diff
@@ -97,22 +97,6 @@ class User(SoftDeletionModel):
...
-        secondaryjoin='Event.id == UserSystemRole.event_id',
-    )
-    sales_admin_events = db.relationship(
-        'Event',
-        viewonly=True,
-        secondary='join(UserSystemRole, CustomSysRole,'
-        ' and_(CustomSysRole.id == UserSystemRole.role_id, CustomSysRole.name == "Sales Admin"))',
-        primaryjoin='UserSystemRole.user_id == User.id',
-        secondaryjoin='Event.id == UserSystemRole.event_id',
-    )
```
iamareebjamal: revert

maze-runnar: reverted them previously.

app/api/helpers/jwt.py
maze-runnar: why?

maze-runnar: I am checking spam after login is successfull. is this logic wrong too?

iamareebjamal: Yes, the function returns User or None, not String. It breaks the contract of the function and will break the callers elsewhere. You never modify the signature of function to something else so easily. What if some caller calls this function and checks if return is not None, and expects the return type to be User like it previously was? Now, the server will crash because it is not user, it is String!

maze-runnar: ok 🙄

app/api/helpers/jwt.py
```diff
@@ -18,6 +18,8 @@ def jwt_authenticate(email, password):
     if user is None:
         return None
     auth_ok = user.facebook_login_hash == password or user.is_correct_password(password)
+    if auth_ok and user.is_blocked:
+        return None
```
iamareebjamal, line 22: Not needed

PR #7605, opened 2021-01-14
Revert "chore(deps): bump greenlet from 0.4.17 to 1.0.0"

iamareebjamal:
Reverts fossasia/open-event-server#7600


iamareebjamal:
Docker build was failing

Issue #7558, opened 2021-01-01
Add functionality to delete User Event Role


Tlazypanda:
**Is your feature request related to a problem? Please describe.**
Currently there is no functionality to delete User Event Role. Adding this functionality can allow event organizers to delete role assigned to any user.

Ref issue: #7235

Tlazypanda:
@iamareebjamal @mariobehling can we approve this issue? :sweat_smile:

Tlazypanda:
@iamareebjamal for creating this endpoint I was wondering if we can pass the id of user event role for deletion or not since I don't see it returned otherwise at any place. So should I instead pass role_invite id and from that query the user event role object and delete? Thanks :v:

iamareebjamal:
No, event user role, like all other endpoints

Tlazypanda:
@iamareebjamal so I should add this as a standalone misc endpoint in role_invites,py or create the api and schema file and set up the resource manager architecture since right now I don't see those files and creation for this model is handled in accept_invite function itself? :sweat_smile:

iamareebjamal:
create the api and schema file and set up the resource manager architecture

iaryangoyal:
can I work on this issue?

PR #7617, opened 2021-01-17
feat: remove owner role deletion

Tlazypanda:
Ref #7558 

Removes ability to delete user event role is the role is that of owner of the event.

#### Checklist


Tlazypanda:
@iamareebjamal Sir I tried this locally but the before_delete_object method is not being triggered and the object is getting deleted can you please check this out? Also while testing I also found out in the `roles` api (ex endpoint: `/v1/roles/7`) I was able to delete any of the 7 options (owner,organizer...) which should not allowed by the before_delete_object method (here the before_delete_object method was missing from the data layer but still adding that made no difference) :sweat_smile: kinda confused

app/api/users_events_roles.py
```diff
@@ -37,10 +39,27 @@ class UsersEventsRolesDetail(ResourceDetail):
...
 
-    decorators = (api.has_permission('is_coorganizer', methods="GET,PATCH,DELETE"),)
+    def before_delete_object(self, users_events_roles, view_kwargs):
+        """
+        method to check for proper permissions for deleting
+        :param users_events_roles:
+        :param view_kwargs:
+        :return:
+        """
+        if not current_user.is_staff and users_events_roles.role_id == 7:
```
iamareebjamal, line 49: What is role ID 7?

app/api/users_events_roles.py
```diff
@@ -37,10 +39,27 @@ class UsersEventsRolesDetail(ResourceDetail):
...
+            raise ForbiddenError(
+                {'source': 'Role'},
+                'You cannot remove the owner of the event unless you are the admin.',
```
iamareebjamal, line 52: Even admin shouldn't be able to remove the owner

app/api/users_events_roles.py
Tlazypanda: Since initial roles aren't allowed to be deleted so I set 7 which denotes role owner in the roles table.

app/api/users_events_roles.py
Tlazypanda: Ohh okay will update that 😅

app/api/users_events_roles.py
iamareebjamal: In my local machine, it is 5, on production, it is 12!
What guarantee is there that owner role ID is going to remain 7 on all machines, DB, forever?

Tlazypanda: ohh didn't think of that :sweat_smile: will update :v:

app/api/users_events_roles.py
```diff
@@ -46,7 +46,8 @@ def before_delete_object(self, users_events_roles, view_kwargs):
         :param view_kwargs:
         :return:
         """
-        if not current_user.is_staff and users_events_roles.role_id == 7:
+        role = Role.query.get_or_404(users_events_roles.role_id)
```
iamareebjamal, line 49: You already have users_events_roles.role

iamareebjamal:
As this is soft deleted, before_update is triggered instead

Tlazypanda:
@iamareebjamal so how should we implement this something like this in before_update method. I am not sure about this :sweat_smile: :- 
```
role = users_events_roles.role
        if role and data.get('deleted_at'):
            if role.name == "owner":
            raise ForbiddenError(
                {'source': 'Role'},
                'You cannot remove the owner of the event.',
            )
          users_events_roles.deleted_at = data['deleted_at']

```

iamareebjamal:
Yes

app/api/users_events_roles.py
```diff
@@ -38,27 +38,29 @@ class UsersEventsRolesDetail(ResourceDetail):
...
-                {'source': 'Role'},
-                'You cannot remove the owner of the event unless you are the admin.',
-            )
+        if role and data.get('deleted_at'):
+            if role.name == "owner":
+                raise ForbiddenError(
+                    {'source': 'Role'},
+                    'You cannot remove the owner of the event.',
+                )
+            users_events_roles.deleted_at = data['deleted_at']
```
iamareebjamal, line 55: I think this should be automatic, please check

Tlazypanda:
@iamareebjamal I think this is ready now :smile:

Tlazypanda:
@iamareebjamal is this alright? :sweat_smile:

app/api/users_events_roles.py
```diff
@@ -37,12 +38,31 @@ class UsersEventsRolesDetail(ResourceDetail):
...
+        :param view_kwargs:
+        :return:
+        """
+        role = users_events_roles.role
+        if role and data.get('deleted_at'):
+            if role.name == "owner":
+                raise ForbiddenError(
+                    {'source': 'Role'},
+                    'You cannot remove the owner of the event.',
+                )
```
iamareebjamal, line 55:
```suggestion
                )

```

app/api/users_events_roles.py
```diff
@@ -37,12 +38,31 @@ class UsersEventsRolesDetail(ResourceDetail):
     users_events_roles detail by id
     """
 
+
```
iamareebjamal, line 41:
```suggestion
```

Tlazypanda:
@iamareebjamal Thanks for the review 😄 ! Should I next look into other places of soft delete where delete method is used instead of update or the all query param for user fav sessions?

iamareebjamal:
> or the all query param for user fav sessions

What?

And no, other places are using it correctly, and if not, it means soft deletion is not really needed for them

Tlazypanda:
The ?all for user favourite sessions which was to be done. 

And the deletion for roles (owner, organizer ...) Was working for me locally which shouldn't be allowed I guess 😅

iamareebjamal:
Yes :+1:

PR #7656, opened 2021-01-23
Removed the limitations in considering mail ids and reduced the time complexity

sudipg4112001:
Fixes #129

`if not re.match(r'[^@]+@[^@]+\.[^@]+', email):` this line is limited only to some mail ids. So, I have tried to resolve this issue by replacing this line by `if not re.match(r'^[a-z0-9]+[\._]?[a-z0-9]+[@]\w+[.]\w+$', email):` I guess this will resolve the issue and all types of mail ids will be considered.
Secondly, 
` continue`
`ask_email = False`
 I feel these lines can be omitted. After, omission it will reduce the time complexity without affecting the main process.

### DIFF
--- a/create_db.py
+++ b/create_db.py
@@ -18,8 +18,6 @@ def create_default_user(email, password):
             email = input("Enter email for super_admin    : ")
             if not re.match(r'^[a-z0-9]+[\._]?[a-z0-9]+[@]\w+[.]\w+$', email):
                 print('\nInvalid email address\n')
-                continue
-            ask_email = False
     if not password:
         ask_password = True
         while ask_password:
--- a/create_db.py
+++ b/create_db.py
@@ -16,7 +16,7 @@ def create_default_user(email, password):
         ask_email = True
         while ask_email:
             email = input("Enter email for super_admin    : ")
-            if not re.match(r'[^@]+@[^@]+\.[^@]+', email):
+            if not re.match(r'^[a-z0-9]+[\._]?[a-z0-9]+[@]\w+[.]\w+$', email):
                 print('\nInvalid email address\n')
                 continue
             ask_email = False


iamareebjamal:
Making it O(∞) reduces complexity?

sudipg4112001 (commented):
Other than checking only "@" and "." in a mail, I feel we should also check alphabets, numbers, alphanumeric characters.

create_db.py
```diff
@@ -16,10 +16,8 @@ def create_default_user(email, password):
         ask_email = True
         while ask_email:
             email = input("Enter email for super_admin    : ")
-            if not re.match(r'[^@]+@[^@]+\.[^@]+', email):
+            if not re.match(r'^[a-z0-9]+[\._]?[a-z0-9]+[@]\w+[.]\w+$', email):
                 print('\nInvalid email address\n')
-                continue
```
sudipg4112001, line 21: For what purpose is the `continue` statement used ?

create_db.py
```diff
@@ -16,10 +16,8 @@ def create_default_user(email, password):
         ask_email = True
         while ask_email:
             email = input("Enter email for super_admin    : ")
-            if not re.match(r'[^@]+@[^@]+\.[^@]+', email):
+            if not re.match(r'^[a-z0-9]+[\._]?[a-z0-9]+[@]\w+[.]\w+$', email):
```
sudipg4112001, line 19: It is checking only "@" and "." rather it should check alphanumeric characters also, to bring more accuracy in checking the email ids.

create_db.py
```diff
@@ -16,10 +16,8 @@ def create_default_user(email, password):
...
                 print('\nInvalid email address\n')
-                continue
-            ask_email = False
```
sudipg4112001, line 22: What is the use of this statement ?

PR #7657, opened 2021-01-23
Improved the checking process for the mails.

sudipg4112001:
Fixes #4586

Other than checking "@" and "." we should also check alphanumeric characters also.

### DIFF
--- a/create_db.py
+++ b/create_db.py
@@ -16,7 +16,7 @@ def create_default_user(email, password):
         ask_email = True
         while ask_email:
             email = input("Enter email for super_admin    : ")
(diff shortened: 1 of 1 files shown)


iamareebjamal:
Why?

sudipg4112001:
> Why?

That statement is considering only "@" and "." in a mail. Basically, it is counting the number of these characters. Nothing other than this it is checking.

iamareebjamal:
So?

sudipg4112001:
> So?

It is not the correct way to check mails, so I am trying to give perfection to this line.

iamareebjamal:
This is a rarely used script and I don't know if your regex is better, it is not a good improvement. Let's make improvement in actual server

PR #7661, opened 2021-01-24
Password made more stronger

sudipg4112001:
Fixes #8765
The password should contain numerical, lowercase and uppercase letters, special characters. This will make the pass word stronger


iamareebjamal:
Not needed

sudipg4112001:
> Not needed

Why ?

PR #7681, opened 2021-01-28
feat: Add ICS attachment to order emails

### LINKED ISSUE #7632: Order Email: Add ics as attachment and standardize for calendar
When an email arrives from other event platforms Gmail and other email clients automatically provide an option to add it to the calendar. 

1. Please research and implement the structure of the email and possibly tags, so these emails provide the same functionality.
2. Add an .ics file as an attachment to order emails for the event as well.

Example how an event shows up on Gmail.

Screenshot from 2021-01-19 10-53-13

ShauryaAg:
Fixes #7632

- Add an .ics file as an attachment to order emails for the event as well.

- Added `.ics` attachment when a user creates an order.

#### Checklist

### DIFF
--- a/app/api/helpers/mail.py
+++ b/app/api/helpers/mail.py
@@ -9,6 +9,7 @@ from sqlalchemy.orm import joinedload
 
 from app.api.helpers.db import save_to_db
 from app.api.helpers.files import make_frontend_url
+from app.api.helpers.ICalExporter import ICalExporter
 from app.api.helpers.log import record_activity
 from app.api.helpers.system_mails import MAILS
 from app.api.helpers.utilities import get_serializer, str_generator, string_empty
@@ -333,14 +334,17 @@ def send_email_to_attendees(order):
     event_id = event.id
     filedir = os.path.join(
         current_app.config.get('BASE_DIR'),
-        f'static/uploads/{event_id}/',
+        f'static/uploads/temp/{event_id}/',
     )
 
     if not os.path.isdir(filedir):
         os.makedirs(filedir)
     filename = "ical.ics"
     ical_file_path = os.path.join(filedir, filename)
-    if ical_file_path:
+    with open(ical_file_path, "w") as temp_file:
+        temp_file.write(str(ICalExporter.export(event_id), 'utf-8'))
+
+    if os.path.exists(ical_file_path):
         if attachments == None:
             attachments = [ical_file_path]
         else:
--- a/app/api/helpers/mail.py
+++ b/app/api/helpers/mail.py
@@ -8,8 +8,7 @@ from flask import current_app, render_template
 from sqlalchemy.orm import joinedload
 
 from app.api.helpers.db import save_to_db
-from app.api.helpers.files import make_frontend_url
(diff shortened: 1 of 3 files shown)


app/api/helpers/mail.py
```diff
@@ -329,6 +329,23 @@ def send_email_to_attendees(order):
...
+
+    if not os.path.isdir(filedir):
+        os.makedirs(filedir)
+    filename = "ical.ics"
+    ical_file_path = os.path.join(filedir, filename)
+    if ical_file_path:
+        if attachments == None:
+            attachments = [ical_file_path]
+        else:
+            attachments.append(ical_file_path)
```
iamareebjamal, line 347: Just generate on the go

ShauryaAg: We can surely do that, won't it be a bad practice since, there'd be multiple copies of the same file. As it already generates a `.ics` file when an event is created.

iamareebjamal:
No need of generating or reading the file, just generate ICS on the go

ShauryaAg:
I am not sure I follow exactly. We need to attach a file to the email, how will just generating the ICS help. I get it that we can just generate the content while using Sendgrid for sending emails, but SMTP mails still need the path to a file in order to attach the file.

iamareebjamal:
https://stackoverflow.com/questions/44672524/how-to-create-in-memory-file-object

ShauryaAg:
`io.StringIO` or `io.BytesIO` will create a file-like object (they don't create an actual file), but there's no way to get a filepath from those, which is what we need in order to attach it to an email. So, the only option is to use an already present file's path.

This is the error we get when we use `io.StringIO` object to attach to an email `kombu.exceptions.EncodeError: Object of type StringIO is not JSON serializable`

iamareebjamal:
Alright, then don't duplicate the code, use a common function to generate ICS

app/api/helpers/mail.py
```diff
@@ -329,6 +329,16 @@ def send_email_to_attendees(order):
     if current_app.config['ATTACH_ORDER_PDF']:
         attachments = [order.ticket_pdf_path, order.invoice_pdf_path]
 
+    event = order.event
+    event_id = event.id
+    ical_file_path = generate_ics_file(event_id)
```
iamareebjamal, line 334:
```suggestion
    ical_file_path = generate_ics_file(event.id)
```

app/api/helpers/tasks.py
```diff
@@ -116,16 +116,28 @@ def send_email_task_sendgrid(payload):
...
+                with open(attachment, 'rb') as f:
+                    file_data = f.read()
+                    f.close()
+                encoded = base64.b64encode(file_data).decode()
+                attachment = Attachment()
+                attachment.file_content = FileContent(encoded)
+                attachment.file_type = FileType('text/calendar')
+                attachment.file_name = FileName('ical.ics')
+                attachment.disposition = Disposition('attachment')
+                message.add_attachment(attachment)
```
iamareebjamal, line 140: Only few lines are different, why duplicating this much?

PR #7682, opened 2021-01-28
fix: allow deletion of user when it's order are being past

maze-runnar:
Fixes https://github.com/fossasia/open-event-frontend/pull/6501

#### Checklist

### DIFF
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -1,5 +1,6 @@
 import base64
 import logging
+from datetime import datetime
 
 from flask import Blueprint, abort, jsonify, make_response, render_template, request
 from flask_jwt_extended import current_user, verify_fresh_jwt_in_request
@@ -309,8 +310,15 @@ class UserDetail(ResourceDetail):
                         )
                         .exists()
                     ).scalar()
+
+                    present_event_orders = db.session.query(
+                        TicketHolder.query.filter_by(user=user)
+                        .join(Order.event)
+                        .filter(Event.ends_at > datetime.now())
+                        .exists()
+                    ).scalar()
                     # If any pending or completed order exists, we cannot delete the user
-                    if order_exists:
+                    if order_exists and present_event_orders:
                         logger.warning(
                             'User %s has pending or completed orders, hence cannot be deleted',
                             user,
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -303,16 +303,15 @@ class UserDetail(ResourceDetail):
                         .join(Order.event)
                         .filter(Event.ends_at > datetime.now())
                         .filter(
-                              or_(
-                                  Order.status == 'completed',
-                                  Order.status == 'placed',
-                                  Order.status == 'initializing',
-                                  Order.status == 'pending',
+                            or_(
+                                Order.status == 'completed',
+                                Order.status == 'placed',
+                                Order.status == 'initializing',
+                                Order.status == 'pending',
                             )
                         )
                         .exists()
                     ).scalar()
-
                     # If any pending or completed order exists, we cannot delete the user
                     if order_exists:
                         logger.warning(
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from flask import Blueprint, abort, jsonify, make_response, render_template, request
 from flask_jwt_extended import current_user, verify_fresh_jwt_in_request
 from flask_rest_jsonapi import ResourceDetail, ResourceList, ResourceRelationship
-from sqlalchemy import and_, or_
+from sqlalchemy import or_
 from sqlalchemy.orm.exc import NoResultFound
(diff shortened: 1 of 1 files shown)


app/api/users.py
```diff
@@ -309,8 +310,15 @@ def before_update_object(self, user, data, view_kwargs):
...
+
+                    present_event_orders = db.session.query(
+                        TicketHolder.query.filter_by(user=user)
+                        .join(Order.event)
+                        .filter(Event.ends_at > datetime.now())
+                        .exists()
+                    ).scalar()
                     # If any pending or completed order exists, we cannot delete the user
-                    if order_exists:
+                    if order_exists and present_event_orders:
```
iamareebjamal, line 321: Doesn't fix anything

iamareebjamal:
How does this fix the issue. Did you test it?

maze-runnar:
> How does this fix the issue. Did you test it?

yes I have tested it. It is working.

iamareebjamal:
What does it fix? I don't understand

The issue is about not preventing account deletion if the tickets are in past, but here, instead you add a new check without modifying the old check. How can it fix the issue?

maze-runnar:
> What does it fix? I don't understand
> 
> The issue is about not preventing account deletion if the tickets are in past, but here, instead you add a new check without modifying the old check. How can it fix the issue?

So i have added a check with  `and`. if order -> event -> endsAt is greater than present time is true and status of ticket is not cancelled is true(if and only if both are true) it will throw forbidden error. otherwise deletion will happen as usual.

iamareebjamal:
Why not remove the old check and only check for new check. The old check is redundant

iamareebjamal:
Change this filter in old query itself. No need for 2 queries. More chances of bugs

maze-runnar:
old check is checking status of order, and this one is event end time of that order. how any of it is redundant?

maze-runnar:
tested" 
Screenshot from 2021-01-29 00-16-32

app/api/users.py
```diff
@@ -299,16 +300,21 @@ def before_update_object(self, user, data, view_kwargs):
                     order_exists = db.session.query(
                         TicketHolder.query.filter_by(user=user)
                         .join(Order)
+                        .join(Order.event)
```
iamareebjamal, line 303:
```suggestion
                        .join(Order.event)
                        .filter(Event.ends_at > datetime.now())
```

app/api/users.py
```diff
@@ -299,16 +300,21 @@ def before_update_object(self, user, data, view_kwargs):
...
-                                Order.status == 'initializing',
-                                Order.status == 'pending',
+                            and_(
+                                or_(
+                                    Order.status == 'completed',
+                                    Order.status == 'placed',
+                                    Order.status == 'initializing',
+                                    Order.status == 'pending',
+                                ),
+                                Event.ends_at > datetime.now(),
```
iamareebjamal, line 312:
```suggestion
                                or_(
                                    Order.status == 'completed',
                                    Order.status == 'placed',
                                    Order.status == 'initializing',
                                    Order.status == 'pending',
```

app/api/users.py
```diff
@@ -299,16 +300,21 @@ def before_update_object(self, user, data, view_kwargs):
...
                             )
                         )
                         .exists()
                     ).scalar()
+
```
iamareebjamal, line 317:
```suggestion
```

app/api/users.py
```diff
@@ -1,10 +1,11 @@
 import base64
 import logging
+from datetime import datetime
 
 from flask import Blueprint, abort, jsonify, make_response, render_template, request
 from flask_jwt_extended import current_user, verify_fresh_jwt_in_request
 from flask_rest_jsonapi import ResourceDetail, ResourceList, ResourceRelationship
-from sqlalchemy import or_
+from sqlalchemy import and_, or_
```
iamareebjamal, line 8:
```suggestion
from sqlalchemy import or_
```

app/api/users.py
```diff
@@ -299,16 +300,19 @@ def before_update_object(self, user, data, view_kwargs):
...
-                            or_(
-                                Order.status == 'completed',
-                                Order.status == 'placed',
-                                Order.status == 'initializing',
-                                Order.status == 'pending',
+                              or_(
+                                  Order.status == 'completed',
+                                  Order.status == 'placed',
+                                  Order.status == 'initializing',
+                                  Order.status == 'pending',
```
iamareebjamal, line 310:
```suggestion
                            or_(
                                Order.status == 'completed',
                                Order.status == 'placed',
                                Order.status == 'initializing',
                                Order.status == 'pending',
```

iamareebjamal:
Check the solution now

iamareebjamal:
Is it working?

maze-runnar:
> Is it working?

yes

PR #7735, opened 2021-02-11
feat: add organizers in bcc of session notify mail

sachinchauhan2889:
implement point 3 in https://github.com/fossasia/open-event-frontend/pull/6620

### DIFF
--- a/app/api/helpers/mail.py
+++ b/app/api/helpers/mail.py
@@ -197,10 +197,7 @@ def send_email_session_state_change(email, session, mail_override: Dict[str, str
         logger.error('No mail found for session state change: ' + session.state)
         return
 
-    if mail.get('bcc'):
-        bcc = list(set(organizers_email + mail.get('bcc')))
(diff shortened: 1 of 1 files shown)


app/api/helpers/mail.py
```diff
@@ -197,12 +197,17 @@ def send_email_session_state_change(email, session, mail_override: Dict[str, str
         logger.error('No mail found for session state change: ' + session.state)
         return
 
+    if mail.get('bcc'):
+        bcc = list(set(organizers_email + mail.get('bcc')))
+    else:
+        bcc = organizers_email
```
iamareebjamal, line 203:
```suggestion
    bcc = list(set(organizers_email + mail.get('bcc', [])))
```

sachinchauhan2889: sir this is giving error.
error when mail.bcc is empty---> cannot add list with nonetype.

PR #7744, opened 2021-02-14
fix: Check email in bcc before remove

iamareebjamal:
Fixes https://github.com/fossasia/open-event-frontend/issues/6686


mariobehling:
Thanks it is working now, but the frontend still shows all organizers as reply to in the frontend.

PR #7757, opened 2021-02-18
feat: Correct Video Stream Moderator API

iamareebjamal:
For https://github.com/fossasia/open-event-frontend/issues/6208


iamareebjamal:
@maze-runnar When I was changing the API to be email based, I noticed that almost nothing worked. Neither creating the moderators, not querying it.

PR #7759, opened 2021-02-19
feat: Use user_id query for exporting my schedule ICS

Tlazypanda:
For https://github.com/fossasia/open-event-frontend/issues/6694

### DIFF
--- a/app/api/custom/calendars.py
+++ b/app/api/custom/calendars.py
@@ -1,6 +1,7 @@
 from flask import request
 from flask.blueprints import Blueprint
 from flask.helpers import make_response
+from flask_jwt_extended.view_decorators import jwt_optional
 
 from app.api.helpers.calendar.ical import to_ical
 from app.api.helpers.permissions import to_event_id
@@ -13,6 +14,7 @@ calendar_routes = Blueprint('calendars', __name__, url_prefix='/v1/events')
     '/<string:event_identifier>.ics',
 )
 @to_event_id
+@jwt_optional
 def export_event(event_id):
     event = Event.query.get_or_404(event_id)
     include_sessions = 'include_sessions' in request.args
@@ -20,7 +22,7 @@ def export_event(event_id):
     user_id = request.args.get('user_id')
 
     response = to_ical(
-        event, user_id, include_sessions=include_sessions, my_schedule=my_schedule
+        event, include_sessions=include_sessions, my_schedule=my_schedule, user_id=user_id
     )
 
     response = make_response(response)
--- a/app/api/custom/calendars.py
+++ b/app/api/custom/calendars.py
@@ -1,7 +1,6 @@
 from flask import request
 from flask.blueprints import Blueprint
 from flask.helpers import make_response
-from flask_jwt_extended.view_decorators import jwt_optional
 
 from app.api.helpers.calendar.ical import to_ical
 from app.api.helpers.permissions import to_event_id
@@ -14,13 +13,15 @@ calendar_routes = Blueprint('calendars', __name__, url_prefix='/v1/events')
     '/<string:event_identifier>.ics',
 )
 @to_event_id
-@jwt_optional
 def export_event(event_id):
     event = Event.query.get_or_404(event_id)
     include_sessions = 'include_sessions' in request.args
     my_schedule = 'my_schedule' in request.args
+    user_id = request.args.get('user_id')
 
-    response = to_ical(event, include_sessions=include_sessions, my_schedule=my_schedule)
+    response = to_ical(
+        event, user_id, include_sessions=include_sessions, my_schedule=my_schedule
+    )
 
     response = make_response(response)
     response.headers['Content-Type'] = 'text/calendar'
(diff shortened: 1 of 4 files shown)


Tlazypanda:
@iamareebjamal there are some other places dependent on export function so should I keep user_id as default value there? :sweat_smile:

app/api/helpers/calendar/ical.py
```diff
@@ -1,14 +1,12 @@
...
-from flask_jwt_extended import current_user
 from icalendar import Calendar, Event
 from sqlalchemy import or_
 from sqlalchemy.orm import joinedload
 
 from app.models.session import Session
 
 
-def to_ical(event, include_sessions=False, my_schedule=False):
+def to_ical(event, user_id, include_sessions=False, my_schedule=False):
```
iamareebjamal, line 9: Should be optional, like other params. This is not a required param to export my schedule

app/api/helpers/calendar/ical.py
```diff
@@ -39,10 +37,8 @@ def to_ical(event, include_sessions=False, my_schedule=False):
             .order_by(Session.starts_at.asc())
         )
         if my_schedule:
-            if not current_user:
```
iamareebjamal, line 42: It should also work with current user

app/api/helpers/ICalExporter.py
```diff
@@ -4,9 +4,9 @@
 
 class ICalExporter:
     @staticmethod
-    def export(event_id):
+    def export(event_id, user_id):
```
iamareebjamal, line 7: No change needed in this file

app/api/custom/calendars.py
```diff
@@ -14,13 +13,15 @@
     '/<string:event_identifier>.ics',
 )
 @to_event_id
-@jwt_optional
```
iamareebjamal, line 17: No need to remove this

tests/all/integration/api/helpers/test_icalexporter.py
```diff
@@ -5,15 +5,15 @@
 from tests.factories.session import SessionSubFactory
 
 
-def test_export_basic(db):
+def test_export_basic(db, user):
```
iamareebjamal, line 8: No change needed in this file

app/api/helpers/calendar/ical.py
Tlazypanda: So we are also keeping the jwt_optional header then too right? :sweat_smile:

Tlazypanda:
@iamareebjamal ohh okay got it I thought we are removing header completely :sweat_smile:

app/api/helpers/calendar/ical.py
iamareebjamal: Yes

app/api/helpers/calendar/ical.py
```diff
@@ -8,7 +7,7 @@
 from app.models.session import Session
 
 
-def to_ical(event, include_sessions=False, my_schedule=False):
+def to_ical(event, include_sessions=False, my_schedule=False, user_id=current_user.id):
```
iamareebjamal, line 10: Default should be none. Python default arguments are evaluated once on load time, so this will never work

Tlazypanda: Yes just pushing that change only :sweat_smile: realised that after pushing :p

app/api/helpers/calendar/ical.py
```diff
@@ -40,8 +40,8 @@ def to_ical(event, include_sessions=False, my_schedule=False, user_id=current_us
         if my_schedule:
             if not (current_user or user_id):
                 return jsonify(error='Forbidden Access'), 401
-            if not user_id:
-                user_id = current_user.id
+
+            user_id = current_user.id if not user_id else user_id
```
iamareebjamal, line 45: user_id overrides current_user

Tlazypanda: Sir I have set it to current_user.id if user_id is not set :sweat_smile: so user_id will override current_user

app/api/helpers/calendar/ical.py
```diff
@@ -39,10 +39,12 @@ def to_ical(event, include_sessions=False, my_schedule=False):
             .order_by(Session.starts_at.asc())
         )
         if my_schedule:
-            if not current_user:
-                return jsonify(error='Login Required'), 401
+            if not (current_user or user_id):
+                return jsonify(error='Forbidden Access'), 401
+
+            user_id = current_user.id if not user_id else user_id
```
iamareebjamal, line 45:
```suggestion
            user_id = user_id or current_user.id
```

app/api/helpers/calendar/ical.py
```diff
@@ -39,10 +39,12 @@ def to_ical(event, include_sessions=False, my_schedule=False):
             .order_by(Session.starts_at.asc())
         )
         if my_schedule:
-            if not current_user:
-                return jsonify(error='Login Required'), 401
+            if not (current_user or user_id):
+                return jsonify(error='Forbidden Access'), 401
```
iamareebjamal, line 43:
```suggestion
                return jsonify(error='Login or User ID required'), 401
```

app/api/helpers/calendar/ical.py
Tlazypanda: Ohh I need to start writing better code :joy:

PR #7767, opened 2021-02-21
for add video room modal

akash2408:
reqd : fossasia/open-event-frontend#6416

-
-
-

#### Checklist


app/models/event.py
```diff
@@ -105,6 +105,7 @@ class Privacy:
     is_ticketing_enabled = db.Column(db.Boolean, default=False)
     is_donation_enabled = db.Column(db.Boolean, default=False)
     is_ticket_form_enabled = db.Column(db.Boolean, default=True, nullable=False)
+    is_video_room_enabled = db.Column(db.Boolean, default=True, nullable=False)
```
MeghalBisht, line 108: nullable=True

MeghalBisht:
Migration is missing

app/models/event.py
hongquan: @MeghalBisht For boolean fields, most of time we expect it to has two values: `true` or `false`, `yes` or `no`. We will be confused if it is `null`, because we will wonder: what `null` should be considered, Yes or No? So the combination of `default=True` and `nullable=False` is to prevent that confusing.

Of course there is case where we want the field to has `null` value. That is when we want the field to have *three* possible states. But in this context, based on the name, `is_video_room_enabled`, I guess that we expect it to have two possible values only.

MeghalBisht:
stale!

PR #7823, opened 2021-03-11
fix: Remove autoescape from after order message in email

sachinchauhan2889:
Fixes https://github.com/fossasia/open-event-frontend/issues/6983


mariobehling:
It looks like the test link is not linkified. What about other HTML. The organizer can put in some HTML formatting, e.g. bold, aligned, bullet points etc. Could you ensure please that this works? Thanks.

iamareebjamal:
No need for strip tags or nl2br now

sachinchauhan2889:
@mariobehling  sir, Screenshot updated

PR #7824, opened 2021-03-11
fix: change session-ical to event-ical in Ticket Order Email

sachinchauhan2889:
Fixes https://github.com/fossasia/open-event-frontend/issues/6984


iamareebjamal:
I don't get it, why so many changes

PR #7853, opened 2021-03-27
feat: Create private chat room per event

MeghalBisht:
1. Separated Login and Registration functions for Rocket chat.
2. Private room are now created per event instead of system wide single room
3. Bot user with email openeventbot@openevent.com is owner of all rooms and would add users to a room. (as  users cannot add themselves inside private rooms)

#### Checklist


MeghalBisht:
Screenshot from 2021-03-31 04-40-56

app/api/schema/settings.py
```diff
@@ -270,6 +270,9 @@ class Meta:
     smtp_encryption = fields.Str(allow_none=True)  # Can be tls, ssl, none
 
     rocket_chat_registration_secret = fields.Str(allow_none=True)
+    rocket_bot_email = fields.Str(allow_none=True)
+    rocket_bot_password = fields.Str(allow_none=True)
+    rocket_bot_name = fields.Str(allow_none=True)
```
iamareebjamal, line 275: Why in schema? Shouldn't these be privaate internal fields?

MeghalBisht:
ready for review! @iamareebjamal

app/api/schema/events.py
```diff
@@ -83,6 +83,7 @@ def validate_timezone(self, data, original_data):
     is_billing_info_mandatory = fields.Bool(default=False)
     is_donation_enabled = fields.Bool(default=False)
     is_chat_enabled = fields.Bool(default=False)
+    chat_room_id = fields.Str(allow_none=True)
```
iamareebjamal, line 86: Shouldn't be needed, this is internal and not editable or viewable

MeghalBisht: room id is needed to add a user into a room. From where will it be accessed?

iamareebjamal: Is it needed in frontend?

iamareebjamal: If so, this should be dumps_only

MeghalBisht: not in frontend, but for add user api call

iamareebjamal:
@MeghalBisht I still can't see where in frontend do we use chat_room_id? We just use event.name + event.identifier?

Which is still not a good solution IMO

app/api/chat/rocket_chat.py
```diff
@@ -24,76 +27,165 @@ def __init__(self, message, code=CODES.UNHANDLED, response=None) -> None:
...
+        db.session.commit()
+        add_in_room(event, user_id, api_url, login_url)
+
+
+def add_in_room(event: Event, rocket_user_id, api_url='', login_url=''):
+    bot_info = check_or_create_bot(event, api_url, login_url)
+    data = get_rocket_chat_token(bot_info, event, False, True)
+    bot_token = data['token']
+    bot_id = data['res']['data']['userId']
+    room_info = {'roomId': event.chat_room_id, 'userId': rocket_user_id}
```
MeghalBisht, line 146: chat room id is used here

iamareebjamal: That's not frontend, no need to add something in schema which is not used in frontend. It gets exposed in the API

MeghalBisht: Ok, but how will room id be given here then? as room id is needed to add a user into a room.

iamareebjamal: Schema and Models are different things. Schema is for API, and model is for code. No change would be needed here if chat_room_id is not in the schema. Schema exposes stuff or takes a field as an input publicly through an API

MeghalBisht:
currently special characters are not included in name.
What can be a better way to rename a room in such a case?

app/models/event.py
```diff
@@ -61,6 +61,7 @@ class Privacy:
     is_promoted = db.Column(db.Boolean, default=False, nullable=False)
     is_demoted = db.Column(db.Boolean, default=False, nullable=False)
     is_chat_enabled = db.Column(db.Boolean, default=False, nullable=False)
+    chat_room_name = db.Column(db.String)
```
iamareebjamal, line 64: No need for a field

app/models/event.py
```diff
@@ -385,6 +386,11 @@ def calc_revenue(self, start=None, end=None):
             or 0
         )
 
+    @property
+    def rocket_room_name(self):
+        self.chat_room_name = self.name.replace(' ', '-') + '-' + self.identifier
+        return self.chat_room_name
```
iamareebjamal, line 392: Why assigning to a field and returning. Doesn't make sense

iamareebjamal:
> What can be a better way to rename a room in such a case?

Use a regex

app/models/event.py
MeghalBisht: what about frontend parameter?

app/models/event.py
```diff
@@ -388,8 +388,7 @@ def calc_revenue(self, start=None, end=None):
 
     @property
     def rocket_room_name(self):
```
iamareebjamal, line 390: Rename to chat_room_name and remove the field

app/models/event.py
iamareebjamal: That's backed by the property

app/api/schema/events.py
```diff
@@ -83,7 +83,6 @@ def validate_timezone(self, data, original_data):
     is_billing_info_mandatory = fields.Bool(default=False)
     is_donation_enabled = fields.Bool(default=False)
     is_chat_enabled = fields.Bool(default=False)
-    chat_room_name = fields.Str(allow_none=True)
```
iamareebjamal: If you remove it from schema, how will frontend access it?

iamareebjamal: This should be dumps_only field

MeghalBisht: yes I saw, will push the changes with regex commit

app/api/schema/events.py
```diff
@@ -83,6 +83,7 @@ def validate_timezone(self, data, original_data):
     is_billing_info_mandatory = fields.Bool(default=False)
     is_donation_enabled = fields.Bool(default=False)
     is_chat_enabled = fields.Bool(default=False)
+    chat_room_name = fields.Str(dump_only=True, allow_none=True)
```
iamareebjamal, line 86:
```suggestion
    chat_room_name = fields.Str(dump_only=True)
```

PR #7864, opened 2021-04-02
fix: convert email message to html file and remove unwanted emails

sachinchauhan2889:
this PR convert Email message to HTML and remove unwanted emails.


sachinchauhan2889:
testing in remaining

PR #7890, opened 2021-04-13
feat: Added check of group role in is_organizer

codedsun:
Fixes #

-
-
-

#### Checklist


app/models/user.py
```diff
@@ -451,6 +453,11 @@ def rocket_chat_username(self):
     def rocket_chat_password(self):
         return get_serializer().dumps(f'rocket_chat_user_{self.id}')
 
+    @property
+    def has_group_role(self):
+        group = UsersGroupsRoles.query.filter_by(user_id=self.id)
```
iamareebjamal, line 458: This shouldn't be added in user. Secondly this gives every user in group team the access level of organizer

iamareebjamal (changes requested):

app/models/user.py
```diff
@@ -451,6 +453,11 @@ def rocket_chat_username(self):
...
+        return group is not None
```
iamareebjamal, line 459: In fact, this gives a user organizer access to all events because it doesn't even check for a particular event.

In fact it will always return True because the query doesn't even execute, it is always not None

app/models/user.py
```diff
@@ -196,11 +197,16 @@ def _is_role(self, role_name, event_id=None):
...
         """
+        from app.models.users_groups_role import UsersGroupsRoles
+
         role = Role.query.filter_by(name=role_name).first()
         uer = UER.query.filter_by(user=self, role=role)
+        ugr = UsersGroupsRoles.query.filter_by(user=self, role=role, accepted=True)
         if event_id:
             uer = uer.filter_by(event_id=event_id)
-        return bool(uer.first())
+            event = Event.query.filter_by(id=event_id).first()
```
iamareebjamal, line 207:
```suggestion
            event = Event.query.get(id)
```

codedsun:
image

@iamareebjamal Review now!

app/models/user.py
```diff
@@ -196,11 +197,16 @@ def _is_role(self, role_name, event_id=None):
...
+            ugr = ugr.filter_by(group=event.group)
```
iamareebjamal, line 208: event.group may be null

Test what happens in that case

codedsun: Then user event role will be tested.

iamareebjamal: Will it crash or not or apply the permission regardless of group

codedsun: Then we should have check like this
if event.group: 
     event = Event.query.filter_by(id=event_id).first()
    ugr = ugr.filter_by(group=event.group)

codedsun:
@iamareebjamal Test are passing now!

PR #7903, opened 2021-04-17
fix: change schema in video_channel api

sachinchauhan2889:
required for https://github.com/fossasia/open-event-frontend/pull/7140

### DIFF
--- a/app/api/video_channel.py
+++ b/app/api/video_channel.py
@@ -9,6 +9,17 @@ from app.models.video_channel import VideoChannel
 from app.models.video_stream import VideoStream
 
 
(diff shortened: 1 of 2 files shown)


app/api/video_channel.py
```diff
@@ -18,7 +18,7 @@ def before_get(self, args, kwargs):
 
     methods = ['GET', 'POST']
     decorators = (api.has_permission('is_admin', methods="POST"),)
-    schema = VideoChannelSchemaPublic
+    schema = VideoChannelSchema
```
iamareebjamal, line 21: Wrong. This will show API key to everyone. We already set proper schema according to admin in before_get

PR #7904, opened 2021-04-19
fix: rename chat group name if event renamed

### LINKED ISSUE #7884: Event Room Chat: Renaming Event name shows no room exists
**Describe the bug**

**To Reproduce**
1. Go to 'event page' after renaming the event
2. Click on 'side panel - chat'

Screenshot from 2021-04-12 22-34-19

**Additional Context**
https://developer.rocket.chat/api/rest-api/methods/groups/rename

MeghalBisht:
Fixes #7884

Screenshot from 2021-04-20 01-28-42

### DIFF
--- a/app/api/chat/rocket_chat.py
+++ b/app/api/chat/rocket_chat.py
@@ -135,11 +135,11 @@ class RocketChat:
             return self.register(user, event)
 
     def check_or_create_bot(self):
-        bot_email = 'openeventbot35@open-event.invalid'
+        bot_email = 'open-event-bot@open-event.invalid'
         bot_user, _ = get_or_create(
             User,
             _email=bot_email,
-            defaults=dict(password=generate_pass(), first_name='openevent-bot35'),
+            defaults=dict(password=generate_pass(), first_name='open-event-bot'),
         )
 
         return bot_user
--- a/app/api/chat/rocket_chat.py
+++ b/app/api/chat/rocket_chat.py
@@ -135,11 +135,11 @@ class RocketChat:
             return self.register(user, event)
 
     def check_or_create_bot(self):
-        bot_email = 'open-event-bot@open-event.invalid'
+        bot_email = 'openeventbot35@open-event.invalid'
         bot_user, _ = get_or_create(
             User,
             _email=bot_email,
-            defaults=dict(password=generate_pass(), first_name='open-event-bot'),
+            defaults=dict(password=generate_pass(), first_name='openevent-bot35'),
         )
 
         return bot_user
--- a/app/api/chat/rocket_chat.py
+++ b/app/api/chat/rocket_chat.py
@@ -209,6 +209,8 @@ def get_rocket_chat_token(user: User, event: Event = None):
 
 
 def rename_rocketchat_room(event: Event):
+    if not event.chat_room_id:
+        return
     settings = get_settings()
     if not (api_url := settings['rocket_chat_url']):
         raise RocketChatException(
--- a/app/api/chat/rocket_chat.py
+++ b/app/api/chat/rocket_chat.py
@@ -210,7 +210,7 @@ def get_rocket_chat_token(user: User, event: Event = None):
 
 def rename_rocketchat_room(event: Event):
     settings = get_settings()
-    if not event.chat_room_id and not (api_url := settings['rocket_chat_url']):
+    if not event.chat_room_id or not (api_url := settings['rocket_chat_url']):
         return
 
     rocket_chat = RocketChat(api_url)
--- a/app/api/chat/rocket_chat.py
+++ b/app/api/chat/rocket_chat.py
@@ -209,13 +209,9 @@ def get_rocket_chat_token(user: User, event: Event = None):
 
 
 def rename_rocketchat_room(event: Event):
-    if not event.chat_room_id:
-        return
     settings = get_settings()
-    if not (api_url := settings['rocket_chat_url']):
-        raise RocketChatException(
-            'Rocket Chat Integration is not enabled', RocketChatException.CODES.DISABLED
-        )
+    if not event.chat_room_id and not (api_url := settings['rocket_chat_url']):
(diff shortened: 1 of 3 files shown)


app/api/events.py
```diff
@@ -525,6 +531,7 @@ def before_update_object(self, event, data, view_kwargs):
         :param view_kwargs:
         :return:
         """
+        EventDetail.event_name = event.name
```
iamareebjamal, line 534: You don't need to do this, and this is wrong, You have data and obj in after update object, use that

app/api/events.py
```diff
@@ -552,6 +559,8 @@ def before_update_object(self, event, data, view_kwargs):
             start_image_resizing_tasks(event, data['original_image_url'])
 
     def after_update_object(self, event, data, view_kwargs):
+        if EventDetail.event_name != event.name:
+            rename_rocketchat_room(event=event)
```
iamareebjamal, line 563: This should be a celery task so that it doesn't slow down the event PATCHH request

app/api/chat/rocket_chat.py
```diff
@@ -206,3 +206,34 @@ def get_rocket_chat_token(user: User, event: Event = None):
 
     rocket_chat = RocketChat(api_url)
     return rocket_chat.get_token(user, event)
+
+
+def rename_rocketchat_room(event: Event):
+    settings = get_settings()
+    if not (api_url := settings['rocket_chat_url']):
+        raise RocketChatException(
```
iamareebjamal, line 214: This method call is called for every event whose name is changed, meaning it'll fail for every event which doesn't have chat enabled, which is most events because chat is disabled by default

app/api/events.py
MeghalBisht: How will i get the previous event name for comparison? Also, what is obj in after update object?

iamareebjamal: Use before update object and flask.g

app/api/helpers/tasks.py
```diff
@@ -663,3 +662,12 @@ def delete_translations(self, zip_file_path):
...
     except:
         logger.exception('Error while deleting translations zip file')
+
+
+@celery.task(name='rename.chat.room')
+def rename_chat_room(event_id):
+    event = db.session.query(Event).filter_by(id=int(event_id)).first()
+    try:
+        rename_rocketchat_room(event=event)
+    except:
```
iamareebjamal, line 672: Never catch bare exception. Better let it throw so we atleast know in stacktrace why it failed

app/api/chat/rocket_chat.py
iamareebjamal: Still not resolved

app/api/events.py
```diff
@@ -552,6 +554,12 @@ def before_update_object(self, event, data, view_kwargs):
             start_image_resizing_tasks(event, data['original_image_url'])
 
     def after_update_object(self, event, data, view_kwargs):
+        event_id = str(event.id)
+        if data.get('name') != g.event_name and event.chat_room_id:
```
iamareebjamal, line 558: Use event.name instead of data

app/api/helpers/tasks.py
```diff
@@ -663,3 +662,15 @@ def delete_translations(self, zip_file_path):
...
         logger.exception('Error while deleting translations zip file')
+
+
+@celery.task(name='rename.chat.room')
+def rename_chat_room(event_id):
+    event = db.session.query(Event).filter_by(id=int(event_id)).first()
+    try:
+        rename_rocketchat_room(event=event)
+        logging.info("Rocket chat room renamed successfully")
+    except Exception as e:
```
iamareebjamal, line 673: Don't catch bare exception

app/api/chat/rocket_chat.py
MeghalBisht: this is resolved now

app/api/helpers/tasks.py
MeghalBisht: I dont understand, what is supposed to be done here?

iamareebjamal: Nothing, remove try catch

app/api/chat/rocket_chat.py
```diff
@@ -209,6 +209,8 @@ def get_rocket_chat_token(user: User, event: Event = None):
 
 
 def rename_rocketchat_room(event: Event):
+    if not event.chat_room_id:
+        return
     settings = get_settings()
     if not (api_url := settings['rocket_chat_url']):
         raise RocketChatException(
```
iamareebjamal, line 216: No exception please, return here as well, this function is not a caandidate for throwing exceptions

app/api/helpers/tasks.py
```diff
@@ -666,6 +666,6 @@ def delete_translations(self, zip_file_path):
 
 @celery.task(name='rename.chat.room')
 def rename_chat_room(event_id):
-    event = db.session.query(Event).filter_by(id=int(event_id)).first()
+    event = Event.query.get(event_id)
     rename_rocketchat_room(event=event)
```
iamareebjamal, line 670:
```suggestion
    rename_rocketchat_room(event)
```

app/api/chat/rocket_chat.py
```diff
@@ -209,13 +209,9 @@ def get_rocket_chat_token(user: User, event: Event = None):
...
 
 def rename_rocketchat_room(event: Event):
-    if not event.chat_room_id:
-        return
     settings = get_settings()
-    if not (api_url := settings['rocket_chat_url']):
-        raise RocketChatException(
-            'Rocket Chat Integration is not enabled', RocketChatException.CODES.DISABLED
-        )
+    if not event.chat_room_id and not (api_url := settings['rocket_chat_url']):
```
iamareebjamal:
```suggestion
    if not event.chat_room_id or not (api_url := settings['rocket_chat_url']):
```

PR #8021, opened 2021-06-09
fix: allow attendee checkin for several days

sachinchauhan2889:
allow attendee checkin for several days

required for https://github.com/fossasia/open-event-frontend/pull/7391


MeghalBisht (approved):
Works.

PR #7705, opened 2021-02-03
Migration from flask-script to flask inbuilt cli

ashutoshc8101:
Fixes #7692 

#### Short description of what this resolves:
flask_script is replaced with flask's inbuilt cli.

#### Changes proposed in this pull request:
- `python manage.py runserver` is migrated to `flask run`
Screenshot from 2021-02-03 14-29-16

- `python manage.py` is migrated to `flask`
Before:
Screenshot from 2021-02-03 14-04-09
After:
op_flask

- `python manage.py db` is migrated to `flask db`
Before:
Screenshot from 2021-02-03 14-04-39

After:
Screenshot from 2021-02-03 14-12-00
- `python manage.py list_routes` is migrated to `flask list_routes`
flask list_routes

- `python manage.py add_event_identifier` is migrated to `flask add_event_identifier`
Screenshot from 2021-02-03 14-23-28

- `python create_db.py` is migrated to `flask create_db`
Screenshot from 2021-02-03 14-23-52

- `python drop_db.py` is migrated to `flask drop_db`
Screenshot from 2021-02-03 14-24-13

- `python manage.py fix_digit_identifier` is migrated to `flask fix_digit_identifier`
Screenshot from 2021-02-03 14-25-22

- `python manage.py fix_event_and_speaker_images` is migrated to `flask fix_event_and_speaker_images`
Screenshot from 2021-02-03 14-25-53

- `python manage.py fix_speaker_images` is migrated to 'flask fix_speaker_images`
Screenshot from 2021-02-03 14-26-10

- `python manage.py initialize_db` is migrated to `flask initialize_db`
Screenshot from 2021-02-03 14-26-33

- `python manage.py prepare_db` is migrated to `flask prepare_db`
Screenshot from 2021-02-03 14-27-22

#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia) and my PR follows them.
- [ ] My branch is up-to-date with the Upstream `development` branch.
- [x] The unit tests pass locally with my changes 
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

- [ ] All the functions created/modified in this PR contain relevant docstrings.

I will squash the commits, once, all the changes requested by maintainers will be made.

PR #8077, opened 2021-07-05
[WIP] Migration from flask-script to flask inbuilt cli

### LINKED ISSUE #7692: Migrate from flask-script to flask inbuilt cli
**Feature Request**
It would be better if functionality to create, populate and drop the database is available using the `manage.py` file instead of using three additional files `create_db.py`, `drop_db.py`, and `populate_db.py`.

The developer should be able to create a database using `python manage.py create_database`, drop a database using `python manage.py drop_db`, and populate using `python manage.py populate_db`

**Do you want to work on this?**
Yes, I want to work on this and make the developer experience better.

hongquan:
My attempt to bring up the (closed) PR #7705 to life. I don't want @ashutoshc8101's effort to be wasted.
This is to fix #7692.


ashutoshc8101:
@hongquan Thanks!

hongquan:
I'm a bit stuck with this, because the way the code is organized is weird. One consequence is that, we have to use a global variable `app_created` to indicate if the app is created. The proper way is that we don't need that indicator.

https://github.com/fossasia/open-event-server/blob/90617d69b23d7be4505bb8396822211801c2d9b7/app/instance.py#L74-L79

To solve this, we need a big change to restructure the code base.

hongquan:
This code structure cause some operations running twice, like adding the same view twice, and cause exception (crash).

PR #8258, opened 2021-10-02
fix: distributed tax calculations causing bugs

### LINKED ISSUE #8250: PDF Invoice: Show Discounts in the invoice
The PDF invoice for ticket buyers does not show discount codes clearly. Please implement the following:
* Add the discount code into the description with "Discount code: samplecodehere, Discount: 10% or USD 2.00 etc."
* Show the discount in the item price list with the original price strike through

### LINKED ISSUE #8254: Unify tax calculation at one place
Right now tax calculation of the system is distributed in different places in different files. There is also repetition of code and some of this calculation is in frontend also. This creates confusion while editing tax info in one place. Also there in `/calculate-amount` endpoint in API for calculating `order_amount`. The idea is to unify tax calculation at one place, preferably `/calculate-amount` endpoint in API

pc-beast:
Fixes #8254 
Partially fixes #8250 (invoice changes remaining)
Partially fixes https://github.com/fossasia/open-event-frontend/issues/7871 (frontend changes remaining)

#### Checklist


app/api/helpers/order.py
```diff
@@ -277,6 +286,8 @@ def calculate_order_amount(tickets, discount_code=None):
                 'discount': discount_data,
                 'ticket_fee': round(ticket_fee, 2),
                 'sub_total': round(sub_total, 2),
+                'ticket_tax': round(ticket_tax, 2),
+                'discounted_tax': round(discounted_tax, 2)
```
maze-runnar, line 290: changes are ok. But I wanted to ask, what if we only keep one 'ticket_tax' and only show it there. As when there will be discount code, we will update it, so we have to only keep track of one tax variable. are there any problems using this approach?

pc-beast: As the idea is to move all the tax calculations to one place, we may lose track of the original tax value on applying discount code if we use only one variable as we need both in frontend at once.

maze-runnar: cool. So this PR is completed or anything else need to be done here?

maze-runnar (approved):

app/api/helpers/order.py
pc-beast: Yes. This is complete. But please do not merge it until #7871 's pr gets completed.

PR #8298, opened 2021-10-19
fix: make emails & invoices translatable

### LINKED ISSUE #8293: Make invoices and emails fully translatable
Emails and invoices are not fully translated. This results in email with mixed languages. Some strings are not labeled ready for translations. Therefore Weblate will not recognize them as translatable. For example in Weblate German shows up as 100% translated even though text is not translated. If Weblate shows up 100% translated, but areas are not really translated something is wrong. Compare https://hosted.weblate.org/projects/open-event/server/de/

For example check the ticket order email. Only the first sentence is translatable.

progmatic-99:
Fixes #8293


mariobehling:
Please post some screenshots.

PR #8339, opened 2021-10-31
chore: add app_name to after event template

### LINKED ISSUE #8321: Translations issues second part
Incomplete sentences
- has just scheduled
- on %(app_name)s.
- of your account
- of your account.
- Late payments can be subject to a
- finance fee.
- This is an automatic email from %(app_name)s for organizers of the
- Your
- has received a new session proposal with the following details
- You can view all speaker details on the speaker profile page and the session proposal on the
- To change the status of sessions please go to the
- and click on the status badge of the session.
- Or paste this token in your

No full stop
- You can unsubscribe from emails from this group by unfollowing the group
- You have been invited as a speaker for the session submission "%(session_title)s" for the event

Wrong sentence - what is the real meaning?
- Click on the following link to manage your event

Change "eventyay" with "app name"
- The invoice for your event %(event_name)s on eventyay.com for %(date)s is now available.

progmatic-99:
Fixes #8321

### DIFF
--- a/app/translations/de/LC_MESSAGES/messages.po
+++ b/app/translations/de/LC_MESSAGES/messages.po
@@ -6,19 +6,18 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: v1\n"
+"Project-Id-Version:  v1\n"
 "Report-Msgid-Bugs-To: open-event@googlegroups.com\n"
-"POT-Creation-Date: 2021-10-31 12:09+0530\n"
+"POT-Creation-Date: 2021-11-01 12:32+0530\n"
 "PO-Revision-Date: 2021-10-31 08:05+0000\n"
 "Last-Translator: Mario Behling <mb@mariobehling.de>\n"
-"Language-Team: German <https://hosted.weblate.org/projects/open-event/server/"
-"de/>\n"
 "Language: de\n"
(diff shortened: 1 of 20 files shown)


app/translations/de/LC_MESSAGES/messages.po
```diff
@@ -284,8 +284,8 @@ msgid ""
 "This is an automatic email from %(app_name)s for organizers of the "
 "%(session_event_name)s."
 msgstr ""
-"Dies ist eine automatische E-Mail von %(app_name)s für die Organisatoren der "
-"Sitzungen %(session_event_name)s."
+"Dies ist eine automatische E-Mail von %(app_name)s für die Organisatoren "
+"der Sitzungen %(session_event_name)s."
```
mariobehling, line 288: I saw that before and got confused already. Shouldn't that variable be called "session_name" instead of "session_event_name"?

progmatic-99: Earlier it was this only. Should I change this?? It will not matter what the name is, the correct value would be put there regardless.

mariobehling: If it is easy to change would be great if you could change it to bring more logic to this.

mariobehling (approved):

Issue #8333, opened 2021-10-31
Make subjects of all mails translatable


progmatic-99:
Currently session state mails and subjects of all mails are not translatable. Please make it possible to translate them.

For example, where are the strings for this email? Where can this be translated?

```
        },
        'canceled': {
            'subject': 'Canceled! Your submission for {event_name} titled {session_name} has been Canceled',
            'message': "Hello,<br/><br/>"
            "This is an automatic message from {app_name}.<br/><br/>"
            "Your session status for the submission {session_name} for {event_name} was changed to \"Canceled\".<br/><br/>"
            "The status change was done by event organizers. If there are questions about this change please contact the organizers.<br/><br/>"
            "You can also check the status and details of your submission on the session page {session_cfs_link}. You need to be logged in to view it.<br/><br/>"
            "More details about the event are on the event page at {event_link}.<br/><br/>"
            "Thank you.<br/>"
            "{frontend_link}",
        },
```

_Originally posted by @mariobehling in https://github.com/fossasia/open-event-server/issues/8321#issuecomment-954752178_

mariobehling:
After the last PR in the mails "New order for" there is a quote sign in front of eventyay (appName).

![Screenshot from 2021-11-10 15-35-12](https://user-images.githubusercontent.com/1583873/141132851-3af91908-c4e6-4235-95e2-fbebf78477b8.png)

progmatic-99:
will correct this in this PR

mariobehling:
The subject of mails is translatable in the translation system, but it does not show up in the subject of recipients if an email is in German for example. Therefore reopening this.

![Screenshot from 2021-11-20 07-02-12](https://user-images.githubusercontent.com/1583873/142716343-1b03a722-2a28-4197-87ca-f11e29835d31.png)

anjali1102:
Kindly asign me this issue

progmatic-99:
@anjali1102 We don't assign issues to external contributors. Please start working on the issue and make a draft PR. We'll help you along the way.

mariobehling:
Subjects still are not translatable after the last PR got merged. There are no additional translation options in Weblate. Mails of orders in other languages have the content in another language than English, but the subject is still not translated.

progmatic-99:
https://stackoverflow.com/questions/17380165/python-gettext-different-languages-at-same-time

PR #8371, opened 2021-11-10
feat: translate date and time according to user language preference in mails & invoice

### LINKED ISSUE #8337: Change language of date and time based on user choice
The language for date and time is not changed in messages even if the message is not in English.

progmatic-99:
Fixes #8337


mariobehling:
Thanks. Could you please look into the DeepSource test that fails?

progmatic-99:
@codedsun @pc-beast please check on your local machines

app/api/helpers/mail.py
```diff
@@ -794,11 +822,23 @@ def send_email_after_event_speaker(email, event_name):
     )
 
 
-def convert_to_event_timezone(date, timezone=None, format='%B %d, %Y %H:%M (%Z%z)'):
-    if not date:
+def convert_to_event_timezone(date, email, timezone=None, fmt='full'):
+    if not date or not email:
```
maze-runnar, line 826: what if `if date and not email`?

maze-runnar (changes requested):

app/api/helpers/mail.py
```diff
@@ -794,11 +822,26 @@ def send_email_after_event_speaker(email, event_name):
...
+
+def convert_to_user_locale(email, date, fmt):
+    """Convert datetime to user selected language"""
+    if date and email:
+        user = User.query.filter(User.email == email).first()
+        user_locale = 'en'
+
+        if user and user.language_prefrence:
+            user_locale = user.language_prefrence
+            return format_datetime(date, format=fmt, locale=user_locale)
```
pc-beast, line 843: This line can be removed.

app/api/helpers/mail.py
```diff
@@ -554,13 +561,6 @@ def send_email_to_attendees(order):
         order_view_url=order.site_view_link,
     )
 
-    event_date = convert_to_event_timezone(
-        order.event.starts_at, order.event.timezone, format='%d %B %Y'
-    )
-    event_time = convert_to_event_timezone(
-        order.event.starts_at, order.event.timezone, format='%H:%M (%Z%z)'
-    )
-
```
pc-beast, line 563: Why remove these? Isn't it better to just change the parameters as this is being used repetitively. :thinking:

pc-beast (changes requested):

progmatic-99:
**ticket_purchased_organizer email:**
ss_1637348080

**ticket_purchased** email
ss_1637348113
ss_1637348128

**ticket_attendee** pdf
ss_1637348140

mariobehling:
Thanks. I am confused about the example emails. I see the text is in English and the date/time is in German here. As we already translated nearly everything to German I would expect everything to be in German in this example.

progmatic-99:
@mariobehling My browser language is set to English, that's why this happened.

**ticket_attendee** mail
ss_1637385702

**ticket_purchased_organizer** mail
ss_1637385734

mariobehling (approved):
Looks good for me. Would be good if @maze-runnar could review the changes.

maze-runnar:
Code changes are ok, but repeated a lot. If we have to convert startsat and endsat so many places, then directly convert them in schema at only one place.

progmatic-99:
@maze-runnar In some mails, we need the date part of datetime object, in others we want only time part and in some places we want datetime + timezone info, so for which of the 3 cases I will convert in the schema??

mariobehling:
@progmatic-99 Could you please change the PR title/commit message to something a bit more meaningful as well? Thank you.

maze-runnar:
> @maze-runnar In some mails, we need the date part of datetime object, in others we want only time part and in some places we want datetime + timezone info, so for which of the 3 cases I will convert in the schema??

Okay then then this method is ok. We can merge this, and then think of a way to improve it

mariobehling:
Regarding this
> My browser language is set to English, that's why this happened.

No matter what language your browser is set it will still show in the original language. Only if you activated auto-translate, e.g. in Chrome, then it will show in English or to whatever you have set it. The subject language has nothing to do with the browser. The email is sent out by the server and needs to be solved in the server.

progmatic-99:
There is a different issue #8333 for subjects of mails. This can be merged @mariobehling @pc-beast @codedsun

pc-beast (approved):

pc-beast:
@mariobehling can be merged now

codedsun (approved):

PR #8415, opened 2021-11-22
fix: errors due to wrong logic in donation tickets

### LINKED ISSUE #8414: Error messages due to confusing donation ticket logic
**Describe the bug**

There is confusion in donation tickets logic which has been improved later in ```calculate_order_amount()```` but the older logic is still being used on top of new logic. Moreover, price of discount ticket is also ambiguous at places.

pc-beast:
Fixes #8414


mariobehling:
Ok, thanks. I think we need to test this on the test system in detail. Therefore merging it.

mariobehling (approved):

codedsun:
This should not have been merged without discussion.

mariobehling:
As discussed in the meeting this can be solved without adding column. So, I am reverting this.

PR #8428, opened 2021-12-01
feat: Added payment intent for stripe

### LINKED ISSUE #8366: Integrate Klarna, Sepa in Stripe
Integrate Klarna, Sepa payment methods in Stripe

codedsun:
Fixes #8366 

-
-
-

#### Checklist


migrations/versions/rev-2021-12-09-16:49:44-b0985248e3d9_.py
```diff
@@ -0,0 +1,30 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'b0985248e3d9'
+down_revision = 'c4fc87788cd5'
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('orders', sa.Column('stripe_payment_intent_id', sa.String(), nullable=True))
+    op.drop_column('orders', 'stripe_session_id')
```
codedsun, line 22: wrong, if we drop then the data will be lost for previous users

app/models/order.py
```diff
@@ -82,7 +82,7 @@ class Status:
     exp_year = db.Column(db.Integer)
     last4 = db.Column(db.String)
     stripe_token = db.Column(db.String)
-    stripe_session_id = db.Column(db.String)
+    stripe_payment_intent_id = db.Column(db.String)
```
codedsun, line 85: we should use in the same column

codedsun (commented):
Added comment

app/models/order.py
pc-beast: we do not need this data specifically for any other purpose other than confirming the order status. Once order status is ` completed` it is of no further use. That's why dropping won't cause any problem to past orders.

codedsun: @pc-beast Still payment data should be available for every order irrespective of time.

progmatic-99 (approved):
The payment flow is working fine. But the details are not appearing correctly.
Screenshot 2021-12-21 at 16-06-58 Completed Order - dfb0cd22-3bf7-4ed5-adb6-da93229d1fa1 Orders Open Event

PR #8447, opened 2021-12-09
fix: change number format & currency according to user language prefe…

### LINKED ISSUE #8338: Number formats and currencies do not adhere to language setting
Currencies and formats in different languages are defined differently, e.g. 2 Euro in French are 2,00 Euro, but in English it is 2.00 Euro. Please implement this in the invoices, emails and UI of the system for all languages.

progmatic-99:
Fixes #8338 

Format of Currencies is changed according to user language preference, but the currency is **not converted to user locale currency**

### DIFF
--- a/app/api/helpers/mail.py
+++ b/app/api/helpers/mail.py
@@ -26,6 +26,7 @@ from babel.dates import (
     format_time,
     format_datetime
 )
+from babel.numbers import format_currency
 
 logger = logging.getLogger(__name__)
 # pytype: disable=attribute-error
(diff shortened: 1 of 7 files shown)


progmatic-99:
@pc-beast @codedsun Please test this on your local systems, as I am not able to create an order on my local machine

pc-beast (commented):
image
works partially. a few other areas need changes. Currency is also being displayed here :+1:

progmatic-99:
@pc-beast What else needs to be done in this??

pc-beast:
> @pc-beast What else needs to be done in this??

In the above SS, format is not converted under `Price` and `Sub-Total` section

mariobehling (approved):
Ok, thanks. Let's try this on the live system.

